PR #7, opened 2022-09-13
[Frontend][Bug] Data Entry Form: Disaggregations Bug For Multi-System Agencies

mxosman:
## Description of the change

Fixes disaggregations for multi-system agencies -- currently, it displays one system's disaggregations for all systems.

Example (agency reporting for the Supervision & Prison systems):

This should show the disaggregations for Parole and not Prisons:

Corrected:
Screen Shot 2022-09-13 at 2 31 15 PM

## Related issues

Closes #15207

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/DataEntryForm.tsx
+++ b/publisher/src/components/Reports/DataEntryForm.tsx
@@ -311,7 +311,7 @@ const DataEntryForm: React.FC<{
                             reportID={reportID}
                             currentIndex={index}
                             reportMetrics={reportMetrics}
-                            metric={reportMetrics[index]}
+                            metric={metric}
                             updateFieldDescription={updateFieldDescription}
                             disabled={isPublished || hasVersionConflict}
                           />


terryttsai (approved):
Great!

publisher/src/components/Reports/DataEntryForm.tsx
```diff
@@ -311,7 +311,7 @@ const DataEntryForm: React.FC<{
                             reportID={reportID}
                             currentIndex={index}
```
lilidworkin, line 312: Is it okay to use `index` here? (given that `index` is now within a system, not across them). I imagine so, just double checking!

mxosman: Ooo - great question! Yes - I believe this is fine to use here because `index` is the index of the currently mapped `metrics` which is being pulled from `metricsBySystem` - but the `TabbedDisaggregations` component in general feels a bit confusing. I don't think we need to find a metric via `index` at all - it could just use the metric that's being passed in.

This brought up another thing that I just noticed was off - I'll put up another fix - and remove `index` all together.

PR #16, opened 2022-09-14
[Frontend] Clear reports and datapoints state when switching agencies

### LINKED ISSUE #15: [Frontend] Clear all report and datapoints state when switching agencies, regardless of what page you're on.
@terryttsai commented on [Fri Aug 12 2022](https://github.com/Recidiviz/recidiviz-data/issues/14618)

This avoids the issue of switching agencies and seeing reports for more than one agency in the Reports list.

terryttsai:
## Description of the change

https://www.loom.com/share/8ec11e17e8004218859c43a77141a461

Set up reactions to clear reports and datapoints state when user switches agencies. 

For the Reports page, we only fetch the reports list when the user navigates to that page, OR the user agency id has changed.

For the Metrics page, we only fetch the datapoints when the user navigates to that page, OR the user agency id has changed.

This means if the user switches agencies while on a different page, we clear, but don't fetch the new data until the user visits a page that needs to load that data, minimizing needless fetches.

## Type of change

## Related issues

Closes #15 

### Development

### Code review

### DIFF
--- a/publisher/src/stores/DatapointsStore.ts
+++ b/publisher/src/stores/DatapointsStore.ts
@@ -15,7 +15,12 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import { makeAutoObservable, runInAction } from "mobx";
+import {
+  IReactionDisposer,
+  makeAutoObservable,
+  reaction,
+  runInAction,
+} from "mobx";
 
 import {
   DatapointsByMetric,
@@ -38,6 +43,8 @@ class DatapointsStore {
 
   loading: boolean;
 
+  disposers: IReactionDisposer[] = [];
+
   constructor(userStore: UserStore, api: API) {
     makeAutoObservable(this);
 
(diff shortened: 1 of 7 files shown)


publisher/src/stores/DatapointsStore.ts
```diff
@@ -46,6 +53,21 @@ class DatapointsStore {
...
+        (currentAgencyId, previousAgencyId) => {
+          if (previousAgencyId !== undefined) {
+            this.resetState();
+          }
+        }
+      )
+    );
+  }
+
+  deconstructor() {
```
terryttsai, line 69: Following the advice here https://stackoverflow.com/questions/59345330/how-do-you-dispose-a-reaction-defined-in-a-class-constructor to handle disposing of a reaction defined in a mobx class, this might be overkill since at least right now we never need to dispose of this since we never need to de-instantiate any of our mobx stores

lilidworkin (commented):
@terryttsai this is very cool!

I tried it out locally and one thing I noticed is that if you switch to a different agency while on a particular report, you get an error:
Screen Shot 2022-09-15 at 10 10 35 AM

This is probably fine (I'm not sure what the right behavior should be in this case?) but just wanted to call it out in case there's an easy fix.

terryttsai:
Thanks for calling that out, I put up a fix for this case by navigating you back to the reports list when you try to switch agencies while on a report: https://www.loom.com/share/f99f609f88e44633868835bc80363ef9

I also replace the browser navigation stack so you can't press the back button and go back to that report page since you'd see an error there.

lilidworkin (approved):
yay!

PR #55, opened 2022-10-04
Enforce Recidiviz copyright header

terryttsai:
## Description of the change

Enforce Recidiviz copyright header using Eslint plugin

Screen Shot 2022-10-04 at 12 07 51 PM

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/.eslintrc.js
+++ b/publisher/.eslintrc.js
@@ -1,0 +1,7 @@
+module.exports = {
+  extends: ["@recidiviz"],
+  plugins: ["header"],
+  rules: {
+    "header/header": [2, "license-header.js"],
+  },
+};
--- a/publisher/license-header.js
+++ b/publisher/license-header.js
@@ -14,5 +14,3 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
-
-/// <reference types="react-scripts" />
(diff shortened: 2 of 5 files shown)


publisher/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  extends: ["@recidiviz"],
+  plugins: ["header"],
+  rules: {
+    "header/header": [2, "license-header.js"],
```
mxosman, line 5: Do you now what the `2` or that position in the array represents?

terryttsai:
So in 2023, if we want to update every header to say 2023, we should be able to update the template and run autofix for it to do that for us.

If we only want new files moving forward to say 2023, and allow old files to say 2022, then we could configure it using a regular expression to allow the year to be 2022, 2023, 2024 etc. See https://github.com/Stuk/eslint-plugin-header#regular-expressions

mxosman:
> So in 2023, if we want to update every header to say 2023, we should be able to update the template and run autofix for it to do that for us.
> 
> If we only want new files moving forward to say 2023, and allow old files to say 2022, then we could configure it using a regular expression to allow the year to be 2022, 2023, 2024 etc. See https://github.com/Stuk/eslint-plugin-header#regular-expressions

What is standard convention for these copyright header years? Every year all files are updated to the current year - or each file has their own "creation" year header?

If we use the regex, would we still have to manually update the file year?

These are just all curiosities! Not sure if it's worth it to go down any route if it feels easy enough for us to remember and update the template annually.

terryttsai:
> What is standard convention for these copyright header years? All files are updated to the current year header - or each file has their own "creation" year header?

Not sure, I heard from Ian that it's not critical but it's not a bad idea to update all files to match the current year

> If we use the regex, would we still have to manually update the file year?

Yeah we'd still have to at least change the template once a year

PR #64, opened 2022-10-06
Settings: Metric Configuration (1/n)

### LINKED ISSUE #63: [Frontend] New Settings Page: Metric Configuration (1/n)

mxosman:
## Description of the change

Removes the `/metrics` route (and from the Menu) and refactors the MetricsView component to match the updated design (up until the detailed metrics view which is currently styled the same as before]). This will help set up the next phase of work within the Metric Configuration which will be updating the detailed metrics view per the new design.

Demo of current change:

https://user-images.githubusercontent.com/59492998/194413204-9c17a55d-be9f-40a4-869c-8fd44cef3bb6.mov

## Related issues

Closes #63 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricsView/MetricsView.tsx
+++ b/publisher/src/components/MetricsView/MetricsView.tsx
@@ -69,6 +69,7 @@ import {
   PanelContainerRight,
   RadioButtonGroupWrapper,
   Slider,
+  StickyHeader,
   Subheader,
   ToggleSwitch,
   ToggleSwitchInput,
@@ -115,11 +116,6 @@ type MetricBoxProps = {
   setActiveMetricKey: React.Dispatch<React.SetStateAction<string>>;
 };
 
-const reportFrequencyBadgeColors: BadgeColorMapping = {
-  ANNUAL: "ORANGE",
-  MONTHLY: "GREEN",
-};
-
 const MetricBox: React.FC<MetricBoxProps> = ({
   metricKey,
   displayName,
@@ -133,23 +129,14 @@ const MetricBox: React.FC<MetricBoxProps> = ({
     <MetricBoxContainer
       onClick={() => setActiveMetricKey(metricKey)}
       enabled={enabled}
-      selected={metricKey === activeMetricKey}
     >
-      <MetricNameBadgeToggleWrapper>
-        <MetricNameBadgeWrapper>
-          <MetricName>{displayName}</MetricName>
-          <Badge
-            color={reportFrequencyBadgeColors[frequency]}
-            disabled={!enabled}
-          >
-            {frequency}
-          </Badge>
-        </MetricNameBadgeWrapper>
-
-        {!enabled && <NotReportedIcon noTooltip />}
-      </MetricNameBadgeToggleWrapper>
-
+      <MetricName>{displayName}</MetricName>
       <MetricDescription>{description}</MetricDescription>
+      <MetricNameBadgeWrapper>
+        <Badge color={!enabled ? "GREY" : "GREEN"} disabled={!enabled} noMargin>
+          {frequency}
+        </Badge>
+      </MetricNameBadgeWrapper>
     </MetricBoxContainer>
   );
 };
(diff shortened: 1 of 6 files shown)


mxosman:
Splitting this up in this way so it's not a huge diff.

publisher/src/components/MetricsView/MetricsView.tsx
```diff
@@ -827,129 +797,78 @@ export const MetricsView: React.FC = observer(() => {
...
+                activeMetricKey={activeMetricKey}
+                setActiveMetricKey={setActiveMetricKey}
+              />
+            ))}
+
+          {/* Metric Configuration */}
+          {activeMetricKey && (
+            <MetricConfigurationDisplay>
+              <BackToMetrics onClick={() => setActiveMetricKey("")}>
+                ← Back to Metrics
```
terryttsai, line 851: This back navigation arrow sort of implies to me that you can press the browser back button to navigate back to metrics, like for example in the data entry page, but it doesn't do that. 

It's probably fine, I can't think of a quick way to make this more consistent anyways.

terryttsai (commented):
Are we removing the ability for users to see data visualizations? Or are we adding it back in a later PR?

mxosman:
> Are we removing the ability for users to see data visualizations? Or are we adding it back in a later PR?

Great question! I'm not sure. We should ask Humphrey & Juan. In the mocks, there's a link to a `Data` page in the menu, which is where I assume it will live? Not 100% sure.

publisher/src/components/MetricsView/MetricsView.tsx
mxosman: Oh interesting! I don't see it that way - if I'm coming from the list of metrics, it looks contained within & relevant to the metric config display that I interpret it as back to the list of metrics completely unrelated to the browser back button. Feel free to bring this up on the Figma design!

mxosman: Maybe saying something like "Back to List of Metrics" would be helpful? Not sure.

The mocks say "Back to All Metrics" but we made a decision a while back on that original sprint to remove the `All Metrics` filter as it didn't feel necessary. Not sure.

terryttsai:
Guess it's not a blocker since you're working for this on a feature branch right? And we won't be releasing that feature branch to production until it's ready?

terryttsai (approved):

publisher/src/components/MetricsView/MetricsView.tsx
terryttsai: Good point, I'll comment on the figma! Meanwhile I don't think this should be a blocker and the current design is good with me!

mxosman:
> Guess it's not a blocker since you're working for this on a feature branch right? And we won't be releasing that feature branch to production until it's ready?

Yeah - exactly! And, if it does end up going in its own page - I would imagine we'd get a mock up of that page too (or we should request one).

I'm also still trying to figure out the best way to split these PRs up - open to any suggestions - frontend is always hard for me to split because of how intertwined everything is.

lilidworkin:
@mxosman I actually LOVE how you split this up! Thank you. So much easier to reason about and review!

re: data page -- great catch @terryttsai ! Can't believe I didn't think of this during design review! Can you or @mxosman post in @justice-counts and tag Humphrey/Juan about this? If we can get a design for how to move the existing data viz into a new page, that might be a perfect project for a frontend contractor.

> Guess it's not a blocker since you're working for this on a feature branch right? And we won't be releasing that feature branch to production until it's ready?

Yes @terryttsai  exactly -- I think this is a really nice feature of the feature branch system -- it should hopefully make it a lot easier to merge in incomplete PRs like this, which we wouldn't want to get into production yet, but we've vetted all the logic as a solid building block.

lilidworkin:
@mxosman just checking, do the uploaded files now show up on their own in the uploaded files tab?

terryttsai:
Loving the feature branches system already!

PR #79, opened 2022-10-12
Render entire datapoints view in agency dashboard

terryttsai:
## Description of the change
Screen Shot 2022-10-12 at 12 17 28 PM

This renders some example datapoints data in the agency dashboard.

I disconnected the DatapointsView component from the mobx store and moved it to the `common` directory. Use the `ConnectedDatapointsView` component to pipe in data from the publisher app's Datapoints store. This is important since the agency dashboard will have a different store for storing datapoints data.

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review


agency-dashboard/src/DashboardView.tsx
```diff
@@ -15,8 +15,88 @@
...
 // =============================================================================
 
+import { DatapointsView } from "@justice-counts/common/components/DataViz/DatapointsView";
+import { DatapointsGroupedByAggregateAndDisaggregations } from "@justice-counts/common/types";
 import React from "react";
 
 import { Container } from "./DashboardView.styles";
 
-export const DashboardView = () => <Container>Hello</Container>;
+const exampleDatapointsGroupedByAggregateAndDisaggregations = {
```
terryttsai, line 24: example data will be replaced with live data once the new endpoint for agency dashboard is created

common/components/DataViz/DatapointsView.styles.tsx
```diff
@@ -22,12 +22,60 @@ import {
...
   Dropdown,
   DropdownMenu,
+  DropdownMenuItem,
   DropdownToggle,
 } from "@recidiviz/design-system";
 import React from "react";
 import styled from "styled-components/macro";
 
-import { ExtendedDropdownMenuItem } from "../Menu/Menu.styles";
+export const ExtendedDropdownMenuItem = styled(DropdownMenuItem)<{
```
terryttsai, line 31: Copied over from Menu.styles.tsx for now but in a follow up we can clean up these styles

terryttsai:
Note the use of `ConnectedDatapointsView` and `DatapointsView`. `DatapointsView` has no knowledge of the store so it can be reused between apps + easily tested. `ConnectedDatapointsView` connects the view to the store with `useStore`.

lilidworkin:
@terryttsai this is awesome! So just to make sure I'm understanding -- DatapointsView is the shared, common component, not connected to any store. ConnectedDatapointsView is an extension of that component used by Publisher, and connected to the Publisher Store, and DashboardView is an extension of that component used by Agency Dashboards, currently using fake data? I really love the re-use of the base view!

lilidworkin (approved):
This looks great to me -- thanks for the small, easily reviewed PR!

common/components/GlobalStyles/index.ts
```diff
@@ -15,6 +15,35 @@
...
+    font-weight: 500;
+    font-size: 16px;
+    background-color: ${palette.solid.white};
+    color: ${palette.solid.darkgrey};
+  }
+
+  input, textarea {
+    font-family: "Inter", sans-serif;
+  }
+`;
```
mxosman, line 45: Personal nit - it feels a bit strange to include this in an `index.ts` with other exports and not its own file and having an export here in the index file. Makes sense that you're organizing the global styles in one place with other global style related components - but I don't know that I would know to come here to update the `styled-components` `GlobalStyle`.

No strong feelings either way - just sharing a personal nit!

mxosman:
> I disconnected the DatapointsView component from the mobx store and moved it to the common directory. Use the ConnectedDatapointsView component to pipe in data from the publisher app's Datapoints store. 

Sorry Terry - a bit lost here. Is the DatapointsView the "generalized" datapoints rendering component? The move to commons then allows publisher and agency dashboards to hydrate it with data and render charts?

terryttsai:
@lilidworkin yes that's exactly right!

@mxosman yes, that's right, `DatapointsView` is just a react component, no mobx store involved. Publisher's `ConnectedDatapointsView` is connected to its mobx store and hydrates the `DatapointsView`. 

Right now Agency Dashboard just uses the DatapointsView with fake data, but in the future it will have its own `ConnectedDatapointsView` since its store is going to look different than Publisher's store.

BTW I realized I used the `Connected` terminology because redux has a `connect` function to hydrate a component with redux state, and the lingo is to call it that a "connected component". In mobx it's called `observer`. I'm open to suggestions for names here!

common/components/GlobalStyles/index.ts
terryttsai: Good question, since we've been only really using `index.ts` to export other things. 

I've seen this practice before of putting definitions / components in index.ts, it made sense to me to put it there since the folder is already called `GlobalStyles`, else I'm not sure where else to put it. In constants.ts?

mxosman: > @lilidworkin yes that's exactly right!
> 
> @mxosman yes, that's right, `DatapointsView` is just a react component, no mobx store involved. Publisher's `ConnectedDatapointsView` is connected to its mobx store and hydrates the `DatapointsView`.
> 
> Right now Agency Dashboard just uses the DatapointsView with fake data, but in the future it will have its own `ConnectedDatapointsView` since its store is going to look different than Publisher's store.
> 
> BTW I realized I used the `Connected` terminology because redux has a `connect` function to hydrate a component with redux state, and the lingo is to call that a "connected component". In mobx we don't really have something like this. I'm open to suggestions for names here!

That makes so much sense! Thank you for clarifying!

Ah - hmm - I'm good with `Connected` for now since we don't use Redux until we think of something better that captures its essense!

mxosman:
> @lilidworkin yes that's exactly right!
> 
> @mxosman yes, that's right, `DatapointsView` is just a react component, no mobx store involved. Publisher's `ConnectedDatapointsView` is connected to its mobx store and hydrates the `DatapointsView`.
> 
> Right now Agency Dashboard just uses the DatapointsView with fake data, but in the future it will have its own `ConnectedDatapointsView` since its store is going to look different than Publisher's store.
> 
> BTW I realized I used the `Connected` terminology because redux has a `connect` function to hydrate a component with redux state, and the lingo is to call that a "connected component". In mobx we don't really have something like this. I'm open to suggestions for names here!

That makes so much sense! Thank you for clarifying!

Ah - hmm - I'm good with `Connected` for now since we don't use Redux until we think of something better that captures its essense!

common/components/GlobalStyles/index.ts
mxosman: I was originally thinking in its own file `GlobalStyle.ts` but that's only because I haven't seen this practice before. I think it's perfectly fine staying in this file - we haven't really had any need to change any of those settings, it just stuck out to me!

PR #88, opened 2022-10-15
Add auth to agency-dashboard to allow easy api development for now

terryttsai:
## Description of the change

So the Agency Dashboard needs a way to fetch some published datapoints to hydrate the charts. 

To do this, the agency dashboard needs to hook into some endpoint with datapoints. One way we can do this is to use the existing service that the publisher dashboard hooks into, however, we'd need to give the agency dashboard Auth0 access to do this. So this PR gives the agency dashboard Auth0 authorization.

Other ideas I considered:
- Making the new endpoint public, so it doesn't require auth. 
  - This opens up a huge can of worms regards to security and privacy, which may not be easy to resolve
- Create a dummy endpoint with dummy data
  - We could do this, however we'd have to manually update the mock data, and it makes it that I can't develop the api in actual backend code
- Create a brand new backend service that has some justice counts test data
  - This goes ahead to create the Aggregator datastore that mimics the model we are going for in the future, but it's a lot of work to do this right now and creating the Aggregator datastore isn't our priority atm.

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review


agency-dashboard/src/stores/API.ts
```diff
@@ -0,0 +1,150 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import { makeAutoObservable, runInAction, when } from "mobx";
+
+// import { trackLoadTime, trackNetworkError } from "../analytics";
```
lilidworkin, line 20: Same as above -- just remove the imports

agency-dashboard/src/stores/API.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.isSessionInitialized = false;
+    this.csrfToken = "";
+
+    when(
+      () => authStore.isAuthorized,
+      () => this.initSession()
+    );
+  }
+
+  async initSession(): Promise<void | string> {
```
lilidworkin, line 51: Do we need `initSession`? I was kind of hoping the frontend wouldn't need to POST anything to the backend, and that the API could be read-only, in which case we don't need a CSRF token

agency-dashboard/src/stores/AuthStore.ts
```diff
@@ -0,0 +1,141 @@
...
+      } else {
+        // strip away all query params just to be safe
+        replacementUrl = `${window.location.origin}${window.location.pathname}`;
+      }
+
+      window.history.replaceState({}, document.title, replacementUrl);
+    }
+
+    if (await auth0.isAuthenticated()) {
+      const user = await auth0.getUser();
```
lilidworkin, line 99: Do we even need the get and store the user? We should avoid doing anything with the user since this is all only temporary, so we don't want to rely on having a user object anywhere

lilidworkin (commented):
Looks good! A few small comments. Love how you were able to move so much stuff to the common directory! @mxosman should for sure take a look as well

agency-dashboard/src/Auth/AuthWall.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  if (!authStore.emailVerified) {
+    return <VerificationPage logoutUser={authStore.logoutUser} />;
+  }
+
+  return authStore.isAuthorized ? <>{children}</> : null;
+};
+
+export default observer(AuthWall);
```
mxosman, line 57: Wondering if it's worth exploring the shared Auth component - unless it's too restrictive/won't play well with what you're setting up. (https://github.com/Recidiviz/web-libraries/tree/main/packages/auth) Probably not worth it if it is more cumbersome.

agency-dashboard/src/stores/API.ts
```diff
@@ -0,0 +1,150 @@
...
+          //     "Your session has expired. Redirecting you to the login page...",
+          //     false,
+          //     "red"
+          //   );
+          // Wait before reloading so user has a chance to see the toast
+          setTimeout(() => {
+            window.location.reload();
+          }, 1000);
+        }
+        throw error;
```
mxosman, line 142: Could get rid of this since it'll throw an error regardless?

mxosman:
This is awesome, Terry! No other feedback outside of a couple small non-blocking comments & Lili's comments. No issues playing around with it & it's awesome to see more components in the commons directory!

agency-dashboard/src/setupProxy.js
```diff
@@ -0,0 +1,29 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+const { createProxyMiddleware } = require("http-proxy-middleware");
+
+module.exports = function (app) {
+  app.use(
+    ["/auth", "/api", "/app_public_config.js"],
+    createProxyMiddleware({
+      //   target: process.env.REACT_APP_PROXY_HOST,
```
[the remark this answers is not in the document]
terryttsai: Thanks for catching that, I'll clear out the code!

agency-dashboard/src/Auth/AuthWall.tsx
terryttsai: Good idea, perhaps for the AuthWall that could be a shared component! Let me try it out

terryttsai:
After [discussing](https://github.com/Recidiviz/recidiviz-data/pull/16024#issuecomment-1282906118) this with Lili, decided we don't need to implement Auth in the agency dashboard, so I'm going to close this! This will greatly simplify the frontend app, since the stores we'll use will be just for fetching and storing public published datapoints

PR #101, opened 2022-10-19
Recreate MetricsView Component for Data Visualization

### LINKED ISSUE #99: [Frontend] Recreate Metrics View for Data Viz Only

mxosman:
## Description of the change

Recreates the old MetricsView component to render data visualizations that were previously a part of the metric configuration. I only kept what was needed to render the metrics list and the data visualization. The metric boxes look different because I reused the metric box component from the new metric config (which are styled slightly differently).

Demo:

https://user-images.githubusercontent.com/59492998/196814023-e48ad674-a392-424d-a994-55a527124c73.mov

[Update] Added spacing between the right side header and the data viz:

Screen Shot 2022-10-19 at 5 17 49 PM

## Related issues

Closes #99 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/MetricsView.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricsView.tsx
@@ -36,7 +36,7 @@ import {
   MetricNameBadgeWrapper,
   MetricSettingsObj,
   MetricsViewContainer,
-  MetricsViewControlPanel,
+  MetricsViewControlPanelOverflowHidden,
   MetricViewBoxContainer,
   PanelContainerLeft,
   PanelContainerRight,
@@ -215,7 +215,7 @@ export const MetricsView: React.FC = observer(() => {
           </TabbedOptions>
         </TabbedBar>
 
-        <MetricsViewControlPanel>
+        <MetricsViewControlPanelOverflowHidden>
           {/* List Of Metrics */}
           <PanelContainerLeft>
             {filteredMetricSettings &&
@@ -258,7 +258,7 @@ export const MetricsView: React.FC = observer(() => {
 
             <ConnectedDatapointsView metric={activeMetricKey} />
           </PanelContainerRight>
-        </MetricsViewControlPanel>
+        </MetricsViewControlPanelOverflowHidden>
       </MetricsViewContainer>
     </>
   );
--- a/publisher/src/components/MetricConfiguration/MetricsView.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricsView.tsx
@@ -1,0 +1,265 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2022 Recidiviz, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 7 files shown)


terryttsai:
design nit: should we add a horizontal line on the top of the dropdowns? I see it in the agency dashboard as well as the old metrics view.
Screen Shot 2022-10-19 at 3 58 37 PM

publisher/src/components/MetricConfiguration/MetricsView.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [loadingError, setLoadingError] = useState<string | undefined>(
+    undefined
+  );
+  const [activeMetricKey, setActiveMetricKey] = useState<string>("");
+  const [metricSettings, setMetricSettings] = useState<{
+    [key: string]: Metric;
+  }>({});
+
+  const filteredMetricSettings: MetricSettingsObj = Object.values(
```
terryttsai, line 107: Could we share the logic between here and MetricsConfiguration?

publisher/src/components/MetricConfiguration/MetricsView.tsx
```diff
@@ -0,0 +1,265 @@
...
+    .filter(
+      (metric) =>
+        metric.system.toLowerCase() === activeMetricFilter?.toLowerCase()
+    )
+    ?.reduce((res: MetricSettingsObj, metric) => {
+      res[metric.key] = metric;
+      return res;
+    }, {});
+
+  const fetchAndSetReportSettings = async () => {
```
terryttsai, line 119: Same with `fetchAndSetReportSettings`?  Would it be sufficient to share these functions, to avoid going the full route of creating a mobx store for metric settings?

terryttsai: Though w a mobx store or some sort of global state, we could avoid having to re-fetch metric settings every time the user enters / leaves one of these pages

lilidworkin: I think I agree with @terryttsai here actually ... it seems worthwhile creating some sort of global state for this metric information, since we're now using it in two different views (do we also use it on Data Entry or is it not needed there?) and will likely use it in additional views later on. Can definitely be done in a followup PR, it's not urgent -- we can taskify for now and come back to it at the end of the sprint if there's time. In the meantime, I like the suggestion of factoring out the shared code!

Also, unrelated, but can we rename it? `reportSettings` confused me -- they're metric settings, not report settings, right? And really it's more like metric metadata?

mxosman: Very good points! I'll taskify and work on it in a follow up PR.

> since we're now using it in two different views (do we also use it on Data Entry or is it not needed there?) 

Unless I'm mistaken, we don't use this function or get any settings in the data entry.

> Also, unrelated, but can we rename it? `reportSettings` confused me -- they're metric settings, not report settings, right? And really it's more like metric metadata?

Haha - yeah, I guess they are technically metric settings for a report? Maybe that's why we went with that in the beginning. I'm good with `fetchAndSetMetricSettings`.

mxosman:
> design nit: should we add a horizontal line on the top of the dropdowns? I see it in the agency dashboard as well as the old metrics view. Screen Shot 2022-10-19 at 3 58 37 PM

Oh strange! Great catch - I didn't mess with the data viz component at all. Perhaps the prior ones had a bottom-border on the heading? I'll take a look.

mxosman:
> > design nit: should we add a horizontal line on the top of the dropdowns? I see it in the agency dashboard as well as the old metrics view. Screen Shot 2022-10-19 at 3 58 37 PM
> 
> Oh strange! Great catch - I didn't mess with the data viz component at all. Perhaps the prior ones had a bottom-border on the heading? I'll take a look.

Yeah - it looks like the top border was never added to the data viz component - it was just using the Tabs bottom border. I'll add a top border.

Screen Shot 2022-10-20 at 11 17 04 AM

terryttsai (approved):
Thanks for adding the top border and taskifying global state for metric settings sounds good to me! Data view looks awesome!

PR #108, opened 2022-10-24
Settings: Metric Configuration: Save Definitions (and Replace Mocks) (6/n)

### LINKED ISSUE #82: [Frontend] New Settings Page: Metric Configuration - Save Definitions - Replace Mocks (6/n)

mxosman:
## Description of the change

Implements functionality to save includes/excludes definition responses and update UI with the changes.

https://user-images.githubusercontent.com/59492998/197611887-00f8d42e-3c62-457c-b6b8-a94bd546eea5.mov

## Related issues

Closes #82 

### Development

### Code review


publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -139,6 +160,34 @@ export const MetricConfiguration: React.FC<{
...
+        }
+
+        return {
+          ...prev,
+          [updatedSetting.key]: {
+            ...prev[metricKey],
+            settings: updatedSettingsArray,
+          },
+        };
+      }
```
mxosman, line 189: Apologies for how this and the `DIMENSION_SETTINGS` looks - it was the simplest way to retrofit into this large gnarly updater function. If we have time to work on a store at the end, each one of these conditionals could be their own functions. But, for the sake of time, I am conforming to this structure as opposed to rewriting everything.

lilidworkin (commented):
I am going to leave this one to @terryttsai  to review since it's heavier on logic! I will say that I definitely think we can/should invest in a store + refactor after this. We should definitely have the time this week!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -139,6 +160,34 @@ export const MetricConfiguration: React.FC<{
...
+      if (typeOfUpdate === "METRIC_SETTING") {
+        let updatedSettingsArray;
+
+        if (
+          prev[metricKey].settings?.length === updatedSetting.settings?.length
+        ) {
+          updatedSettingsArray = updatedSetting.settings;
+        } else {
+          updatedSettingsArray = prev[metricKey].settings?.map((setting) => {
+            if (setting.key === updatedSetting.settings?.[0].key) {
```
lilidworkin, line 172: Would it make sense to store settings mapped by key? e.g. instead of `settings?: MetricConfigurationSettings[]` it would be `settings: Dict[str, MetricConfigurationSetting]`. Not only would this be faster, I feel like it would make the code simpler?

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -259,6 +308,71 @@ export const MetricConfiguration: React.FC<{
         };
       }
 
+      if (typeOfUpdate === "DIMENSION_SETTING") {
+        const updatedDisaggregations = prev[metricKey].disaggregations.map(
+          (disaggregation) => {
+            if (
```
lilidworkin, line 314: To reduce indentation and improve readability, can we reverse some of these `if` statements, so the shorter one comes first, and then we can put the longer one on the main code path?

mxosman: Ah, I normally follow this convention when updating in a map function:

```
thingToMap.map(item => {
   if (special condition) {
       // update just this item in the list
   }
   return item;    // item that doesn't need to be updated just gets returned
}
```

Is your suggestion the following?

```
thingToMap.map(item => {
   if (!special condition) {
       return item;
   }
   
   // update item that needs to be updated
}
```

I can do that - but, it does feel a little less explicit that we're updating an item that matches the exact key.

lilidworkin: Yep that was my suggestion -- generally I do think it improves readability to do the "fast exit" case first, though I don't have strong feelings if you think it's not as clear!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -259,6 +308,71 @@ export const MetricConfiguration: React.FC<{
...
+              disaggregation.key === updatedSetting.disaggregations?.[0].key
+            ) {
+              return {
+                ...disaggregation,
+                dimensions: disaggregation.dimensions.map((dimension) => {
```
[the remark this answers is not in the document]
mxosman: Definitely worth exploring in a bigger refactor!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
mxosman: Oh, I should've used a better word - when I said "quicker" I definitely was not referring to time complexity at all! I should've said something like "fewer transformational (from array > object > array) steps in my mind".

Thank you for clarifying how you were seeing this - very helpful to see the example! This makes sense and absolutely could work in JS world. I can try that and let you know how it works - definitely agree that it is cleaner!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
mxosman: Oh, nothing about this old code is clear to me anymore... 😭 😢 

Thank you for confirming! No strong feelings either - happy to make that adjustment!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
mxosman: Although, I don't think I can actually just try this within this current structure - but, will keep this in mind for when we can refactor to mapping objects such as this to a key: value structure and storing them to get the benefit of a quick lookup as you explained!

lilidworkin: I was imagining we'd change the type in the store (or wherever it gets sent back to) as well! But yea, this is just an idea to think about moving forward, it's certainly not blocking.

terryttsai: So although rendering from an array is a bit nicer than having to convert an object into an array first, if we're updating values in the array by key, like in this metric settings array, I would store the array as a dict instead, so our updaters look like what Lili suggested above, this makes it clearer what we're intending to update and it's more efficient than looping through the entire array each time we update.

When we render, at that point it can be converted back into an array for rendering.

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -139,6 +160,34 @@ export const MetricConfiguration: React.FC<{
         };
       }
 
+      if (typeOfUpdate === "METRIC_SETTING") {
```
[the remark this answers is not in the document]
mxosman: Would love to sync on this - haven't formulated any plans yet! But yes exactly, the way I'm thinking about this is inline with what you described - a bit similar to the data entry store and how it updates.

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
mxosman: This all makes sense - thank you both for sharing your thoughts on a cleaner way to implement this with a dictionary!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -117,6 +126,18 @@ export const MetricConfiguration: React.FC<{
       });
 
       setListOfMetrics(listOfMetricsForMetricNavigation);
+
+      /** Update activeDimension when settings are updated */
```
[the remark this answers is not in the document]
mxosman: This kicked my butt on Friday - I couldn't figure out why certain things were not updating and it's because `activeDimension` needs to also update when there's a change to the metricSettings. The MetricDefinitions component currently relies on an up to date `activeDimension`.

terryttsai (approved):
Code in its current paradigm looks good to me for playtesting!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -259,6 +311,71 @@ export const MetricConfiguration: React.FC<{
         };
       }
 
+      if (typeOfUpdate === "DIMENSION_SETTING") {
+        const updatedDisaggregations = prev[metricKey].disaggregations.map(
+          (disaggregation) => {
+            if (
+              disaggregation.key !== updatedSetting.disaggregations?.[0].key
+            ) {
+              return disaggregation;
```
lilidworkin, line 320: Thanks for reversing the order of these! I do actually think this is a lot better! The way I think about it is, before when when reading it, my brain goes "Ok lots of stuff is happening here if the key matches, but there's a nagging curiosity in my head about what happens when it doesn't match" vs now it's like "Ok, I know what happens when the key is NOT equal -- nothing happens -- now I can focus all my energy on what happens when the key DOES match!"

mxosman: The way you explain it makes **_so_** much sense!

PR #114, opened 2022-10-24
[Frontend] Implement BaseDatapointsStore class that is shared between our frontend apps

terryttsai:
## Description of the change

Create a new BaseDatapointsStore class that is implemented by both the Publisher and Agency Dashboard. Each app overrides the BaseDatapointsStore with their own fetch method.

This allows us to share code between the two apps. 

Note that although mobx allows subclassing, they recommend to use composition over inheritance. Perhaps composition could be achieved by passing in the fetch method and the userStore and setting up the disposers somehow?

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/common/stores/BaseDatapointsStore.ts
+++ b/common/stores/BaseDatapointsStore.ts
@@ -15,7 +15,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import BaseDatapointsStore from "@justice-counts/common/stores/BaseDatapointsStore";
 import {
   DatapointsByMetric,
   DataVizAggregateName,
@@ -23,11 +22,15 @@ import {
   RawDatapoint,
 } from "@justice-counts/common/types";
 import { isPositiveNumber } from "@justice-counts/common/utils";
-import { makeObservable, override, runInAction } from "mobx";
+import {
+  action,
+  computed,
+  makeObservable,
+  observable,
+  runInAction,
+} from "mobx";
 
-import { request } from "../utils/networking";
-
-class DatapointsStore extends BaseDatapointsStore {
+abstract class DatapointsStore {
   rawDatapoints: RawDatapoint[];
 
   dimensionNamesByMetricAndDisaggregation: DimensionNamesByMetricAndDisaggregation;
@@ -35,9 +38,14 @@ class DatapointsStore extends BaseDatapointsStore {
   loading: boolean;
 
   constructor() {
-    super();
     makeObservable(this, {
-      getDatapoints: override,
+      rawDatapoints: observable,
+      dimensionNamesByMetricAndDisaggregation: observable,
+      loading: observable,
+      metricKeyToDisplayName: computed,
+      datapointsByMetric: computed,
+      getDatapoints: action,
+      resetState: action,
     });
     this.rawDatapoints = [];
     this.dimensionNamesByMetricAndDisaggregation = {};
@@ -114,31 +122,7 @@ class DatapointsStore extends BaseDatapointsStore {
   }
 
   async getDatapoints(agencyId: number): Promise<void | Error> {
-    try {
-      const response = (await request({
-        path: `/api/agencies/${agencyId}/published_datapoints`,
-        method: "GET",
-      })) as Response;
-      if (response.status === 200) {
-        const result = await response.json();
-        runInAction(() => {
-          this.rawDatapoints = result.datapoints;
-          this.dimensionNamesByMetricAndDisaggregation =
-            result.dimension_names_by_metric_and_disaggregation;
-        });
-      } else {
-        const error = await response.json();
-        throw new Error(error.description);
-      }
-      runInAction(() => {
-        this.loading = false;
-      });
-    } catch (error) {
-      runInAction(() => {
-        this.loading = false;
-      });
-      throw error;
-    }
+    return Promise.reject("getDatapoints not implemented");
   }
 
   resetState() {
--- a/common/stores/BaseDatapointsStore.ts
+++ b/common/stores/BaseDatapointsStore.ts
@@ -121,9 +121,7 @@ abstract class DatapointsStore {
     }, {});
   }
 
-  async getDatapoints(agencyId: number): Promise<void | Error> {
-    return Promise.reject("getDatapoints not implemented");
-  }
+  abstract getDatapoints(agencyId: number): Promise<void | Error>;
 
   resetState() {
     // reset the state
--- a/common/stores/BaseDatapointsStore.ts
+++ b/common/stores/BaseDatapointsStore.ts
@@ -124,7 +124,6 @@ abstract class DatapointsStore {
   abstract getDatapoints(agencyId: number): Promise<void | Error>;
 
   resetState() {
-    // reset the state
     runInAction(() => {
       this.rawDatapoints = [];
       this.loading = true;
--- a/agency-dashboard/src/stores/RootStore.ts
+++ b/agency-dashboard/src/stores/RootStore.ts
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
+
 import DatapointsStore from "./DatapointsStore";
 
 class RootStore {
--- a/agency-dashboard/src/stores/index.ts
+++ b/agency-dashboard/src/stores/index.ts
@@ -15,6 +15,5 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-export * from "./DatapointsStore";
 export { default as rootStore } from "./RootStore";
 export * from "./StoreProvider";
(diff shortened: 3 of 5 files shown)


lilidworkin (commented):
@terryttsai this looks great! Thanks for the idea / initiative to do this.

Re: inheritance vs. composition -- is [this](https://mobx.js.org/subclassing.html) the page in the mobX docs you were looking at? That's interesting. I think what you have to totally fine, but just so I understand / for my own education, can you clarify what you mean by "Perhaps composition could be achieved by passing in the fetch method and the userStore and setting up the disposers somehow?" I'm (naively) imagining we could achieve composition in a simpler way, e.g.

class DatapointsStore {
   baseStore: BaseDatapointsStore;

  get metricKeyToDisplayName(): {
     return baseStore.getMetricKeyToDisplayName()
  }

  get datapointsByMetric(): {
    return baseStore.datapointsByMetric()
  }
}

etc. Definitely more boilerplate involved there, though.

terryttsai:
I was thinking of something like this:
```
class BaseDatapointsStore {
  rawDatapoints: RawDatapoint[];
  dimensionNamesByMetricAndDisaggregation: DimensionNamesByMetricAndDisaggregation;
  loading: boolean;

  constructor(fetchDatapoints) {
     ...
     this.fetchDatapoints = fetchDatapoints;
  }
  
  ...
  
  async getDatapoints(): Promise<void | Error> {
    try {
      const { currentAgency } = this.userStore;
      if (currentAgency === undefined) {
        // If user is not attached to an agency,
        // no need to bother trying to load this data.
        runInAction(() => {
          this.loading = false;
        });
        return;
      }
      const response = (await fetchDatapoints()) as Response;
      ...
  }
```

And both agency and publisher would supply the fetchDatapoints method. Not sure what to do to set up the disposers in the constructor though.

terryttsai:
For your suggestion, would that mean both the BaseDatapointsStore and the DatapointsStore would have their own copy of data?

agency-dashboard/src/stores/DatapointsStore.ts
```diff
@@ -15,26 +15,30 @@
...
 
   dimensionNamesByMetricAndDisaggregation: DimensionNamesByMetricAndDisaggregation;
 
   loading: boolean;
 
   constructor() {
-    makeAutoObservable(this);
+    super();
+    makeObservable(this, {
+      getDatapoints: override,
```
mxosman, line 40: I'm trying to understand the `override` annotation and still can't quite figure it out - but I feel like I'm almost there! Is it saying the annotation will be overridden in the subclass or is this how you let it know that this is a method that will be overridden by the subclass's method?

common/stores/BaseDatapointsStore.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  constructor() {
+    makeObservable(this, {
+      rawDatapoints: observable,
+      dimensionNamesByMetricAndDisaggregation: observable,
+      loading: observable,
+      metricKeyToDisplayName: computed,
+      datapointsByMetric: computed,
+      getDatapoints: action,
+      resetState: action,
```
mxosman, line 48: I personally love this way of listing out the observable members of the class - it feels super clear. How do you decide between using `makeObservable` and listing them out like this vs. a blanket `makeAutoObservable(this)`?

common/stores/BaseDatapointsStore.ts
```diff
@@ -0,0 +1,135 @@
...
+          [dp.dimension_display_name]: sanitizedValue,
+          frequency: dp.frequency,
+          dataVizMissingData: 0,
+        };
+      }
+      return res;
+    }, {});
+  }
+
+  abstract getDatapoints(agencyId: number): Promise<void | Error>;
```
mxosman, line 124: I understand why this class would be `abstract` and that it can now only be extended. Would you mind explaining the `abstract` method? Is it because whichever class extends this needs to use their own `getDatapoints` method - or is there another reason?

common/stores/BaseDatapointsStore.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  resetState() {
+    // reset the state
```
mxosman, line 127: super tiny nit: I don't think you need this comment - the function name is super clear.

mxosman (approved):
Code looks good to me, Terry! I learned so much from this PR - love the way you are structuring this and utilizing different aspects of mobX to build a shared store. 

I only had questions for my learnings! Love seeing how you're thinking through this. The proposed structure makes sense to me - each app supplying its own fetch method.

mxosman:
> Not sure what to do to set up the disposers in the constructor though. In Publisher we want to reset the datapoints state when we switch agencies.
> 
> Maybe that logic to reset the datapoints state when we switch agencies can live outside that datapoints store though

Ah - that's a good point. Apologies if this is a silly question - but since the `DatapointStore` in publisher extends `BaseDatapointsStore` - wouldn't it be able to add its own methods like the disposers and the deconstructor?

agency-dashboard/src/stores/DatapointsStore.ts
terryttsai: Sorry, I meant to link this guide here: https://mobx.js.org/subclassing.html

> Use `override` annotation for methods/getters overridden in subclass

So I think it's used for anything overridden in the subclass!

common/stores/BaseDatapointsStore.ts
terryttsai: Turns out mobx gives you a warning if you try to use `makeAutoObservable` in a subclass!

terryttsai: I think if you use `makeAutoObservable` you're not allowed to use the `super` call in the constructor

common/stores/BaseDatapointsStore.ts
terryttsai: Yes, BaseDatapointsStore is an abstract class and `getDatapoints` is an abstract method since it's up to the class that uses this to implement it

lilidworkin:
> For your suggestion, would that mean both the BaseDatapointsStore and the DatapointsStore would have their own copy of data?

No, I guess I was imagining BaseDatapointsStore wouldn't actually hold a copy of the data (so maybe the name is a misnomer -- maybe it's DatapointsUtilities) but rather just hold the utility functions for acting on the data. It seems like the main shared logic between the two is metricKeyToDisplayName and datapointsByMetric, which are really utilities in any case, and don't really rely on the fact that it's a store? Unless I'm missing something?

Anyway, this is entirely academic -- what you have is great! I'm not worried about the inheritance vs. composition thing, I just think it's interesting to talk about :D

lilidworkin (approved):

common/stores/BaseDatapointsStore.ts
mxosman: Hahaha - how about an `alert()`?

common/stores/BaseDatapointsStore.ts
mxosman: Awesome! Thank you for explaining that!

common/stores/BaseDatapointsStore.ts
mxosman: Ohhh! OK - good to know!

agency-dashboard/src/stores/DatapointsStore.ts
mxosman: OK - got it! Then I wonder how you would annotate this as an `action`, `computed` or `observable` - wondering if that happens under the hood or if the subclass can re-annotate?

terryttsai: Not sure what you mean, I think the parent class annotates it as an `action`, `computed`, or `observable`, and the child class that overrides it, it's still an `action`, `computed`, or `observable`.

terryttsai:
> Ah - that's a good point. Apologies if this is a silly question - but since the DatapointStore in publisher extends BaseDatapointsStore - wouldn't it be able to add its own methods like the disposers and the deconstructor?

I was thinking of some way of only creating one DatapointsStore and have both apps instantiate it with parameters for fetching and disposers, but it feels weird to try to implement the disposers when it's only used in the publisher datapointsstore. 

> No, I guess I was imagining BaseDatapointsStore wouldn't actually hold a copy of the data (so maybe the name is a misnomer -- maybe it's DatapointsUtilities) but rather just hold the utility functions for acting on the data. It seems like the main shared logic between the two is metricKeyToDisplayName and datapointsByMetric, which are really utilities in any case, and don't really rely on the fact that it's a store? Unless I'm missing something?

Yeah we could just share those utility functions, but both DatapointsStore also use the same mobx observables `rawDatapoints`, `dimensionNamesByMetricAndDisaggregation`, and `loading`, so I wanted to share that too.

agency-dashboard/src/stores/DatapointsStore.ts
mxosman: Oh! That actually clears it up - I forgot this is the subclass... Parent annotates, subclass marks it as `override` but it is still whatever it was annotated at the parent level. Got it! Thanks for explaining that.

PR #115, opened 2022-10-25
Settings: Metric Configuration: Playtesting Followups - Enable Dimension Selection When Disaggregation is Off (4/n) 

mxosman:
## Description of the change

Allows user to re-enable a dimension field (by clicking the unchecked area) when the entire disaggregation is off. The result will be enabling the clicked dimension and enabling the disaggregation.

Demo:

https://user-images.githubusercontent.com/59492998/197656591-eca39beb-e69a-4ff4-b257-b509741ff20c.mov

## Related issues

Contributes to #109 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
@@ -40,6 +40,7 @@ import {
   Configuration,
   Metric,
   MetricBox,
+  MetricBoxBottomPaddingContainer,
   MetricConfigurationDisplay,
   MetricConfigurationWrapper,
   MetricDefinitions,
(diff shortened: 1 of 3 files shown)


publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -279,6 +279,7 @@ export const MetricConfiguration: React.FC<{
 
               return {
                 ...disaggregation,
+                enabled: true,
```
lilidworkin, line 282: Do we always want to do this? Or just if the update has been turning a dimension from off to on? What if the update was turning the last dimension from on to off? Then shouldn't the whole disaggregation turn off?

mxosman: Great question! Definitely had to think this one through - but the case you're referring to is taken care of in L255-278. Otherwise, I think it's safe to say that if you're clicking a dimension and it's not the last dimension, that the disaggregation should also be on (if it is off). 

I did at one point have a conditional that explicitly checked if disaggregation was disabled and the dimension was being enabled and it wasn't the last dimension - but that ended up not being necessary. What do you think?

lilidworkin: Oh my bad I missed that case above! This is all good then!

lilidworkin (approved):
Thank you!

terryttsai (approved):

PR #128, opened 2022-10-31
Metric Configuration Store (1/2)

### LINKED ISSUE #104: [Frontend][Publisher] Create Metric Settings Store

mxosman:
## Description of the change

Introduces mobX store (`MetricConfigStore`) to house and organize metric configuration logic.

## Related issues

Closes #104

### Development

### Code review

### DIFF
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -129,7 +129,9 @@ class MetricConfigStore {
       throw new Error("There was an issue retrieving the metric settings.");
     }
 
-    return response;
+    const metrics: Metric[] = await response.json();
+
+    return metrics;
   }
 
   async saveMetricSettings(updatedMetricSettings: MetricSettings[]) {
@@ -155,8 +157,7 @@ class MetricConfigStore {
   }
 
   async initializeMetricConfigStoreValues() {
-    const response = await this.getMetricSettings();
-    const metrics: Metric[] = await response.json();
+    const metrics = await this.getMetricSettings();
 
     runInAction(() => {
       metrics.forEach((metric) => {
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -21,7 +21,7 @@ import {
   MetricConfigurationSettingsOptions,
   MetricContext,
 } from "@justice-counts/common/types";
-import { makeAutoObservable } from "mobx";
+import { makeAutoObservable, runInAction } from "mobx";
 
 import { MetricSettings } from "../components/MetricConfiguration";
 import { isPositiveNumber, removeCommaSpaceAndTrim } from "../utils";
@@ -32,6 +32,8 @@ class MetricConfigStore {
   userStore: UserStore;
 
   api: API;
+
+  activeMetricKey: string | undefined;
 
   metrics: {
     [systemMetricKey: string]: {
@@ -91,6 +93,7 @@ class MetricConfigStore {
 
     this.api = api;
     this.userStore = userStore;
+    this.activeMetricKey = undefined;
     this.metrics = {};
     this.metricDefinitionSettings = {};
     this.disaggregations = {};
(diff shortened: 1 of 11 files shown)


publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    /** Update value */
+    this.metrics[systemMetricKey].enabled = enabledStatus;
+
+    /** Return an object in the desired backend data structure for saving purposes */
+    return {
+      key: metricKey,
+      enabled: enabledStatus,
+    };
```
mxosman, line 249: I was wondering - what if these updater functions return an object in the desired backend data structure? That way, in the UI, when a user makes an adjustment, we update the state, receive back an object and use that to fire the save function/debounced save function. What do you think?

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+
+  api: API;
+
+  activeMetricKey: string | undefined;
+
+  metrics: {
+    [systemMetricKey: string]: {
+      enabled?: boolean;
+    };
+  };
```
mxosman, line 42: Will need to also include the necessary metadata for rendering (e.g. display_name, label, etc.)

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+class MetricConfigStore {
+  userStore: UserStore;
+
+  api: API;
+
+  activeMetricKey: string | undefined;
+
+  metrics: {
+    [systemMetricKey: string]: {
+      enabled?: boolean;
```
terryttsai, line 40: Do we also want to store the other stuff from the backend response like category, description, definitions, frequency, label, reporting_note, system, unit, etc?

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+    this.dimensionDefinitionSettings = {};
+    this.contexts = {};
+  }
+
+  static getSystemMetricKey(metricKey: string, system: string) {
+    return `${system}-${metricKey}`;
+  }
+
+  static splitSystemMetricKey(systemMetricKey: string) {
+    const [system, metricKey] = systemMetricKey.split("-");
```
terryttsai, line 110: Wonder if a metric or system would have a hyphen in it in the future

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+    return { system, metricKey };
+  }
+
+  async getMetricSettings() {
```
terryttsai, line 114: So this method returns a fetch response, not the actual metric settings object?

publisher/src/stores/MetricConfigStore.tsx
terryttsai: That sounds good to me, so like call updateX to both update the mobx store and return the backend-formatted object, and then saveMetricSettings to send it to the backend?

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+import { isPositiveNumber, removeCommaSpaceAndTrim } from "../utils";
+import API from "./API";
+import UserStore from "./UserStore";
+
+class MetricConfigStore {
+  userStore: UserStore;
+
+  api: API;
+
+  activeMetricKey: string | undefined;
```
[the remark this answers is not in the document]
mxosman: Great question! I don't think so. As I was playing around with connecting the UI to the store, I think we might need an activeSystem string somewhere too (here and eventually in the url route?) as its used by the SettingsMenu component.

publisher/src/stores/MetricConfigStore.tsx
mxosman: Exactly!

publisher/src/stores/MetricConfigStore.tsx
mxosman: Yeah absolutely! I was working on a follow up PR to this trying to see what it would take to use what we made here to render the UI. I added an optional `metadata?` param to the updater function that allows an object of whatever metadata we need `{ frequency, description, label, ...}` so when we initialize it, it will add the metadata as its going through the backend response.

I'll work out what we need to render the UI in that follow up PR and update the types according!

publisher/src/stores/MetricConfigStore.tsx
mxosman: Yeah - I thought about this too. It feels like the convention on the backend is _ snakecase. Since we're creating a new hyphenation - at the very least this should capture the first one (the one we set) - but, you're right, if the system has a hyphen it won't work.

publisher/src/stores/MetricConfigStore.tsx
mxosman: It could return the settings object json - I can do that!

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -0,0 +1,474 @@
...
+  }
+
+  static getSystemMetricKey(metricKey: string, system: string) {
+    return `${system}-${metricKey}`;
+  }
+
+  static splitSystemMetricKey(systemMetricKey: string) {
+    const [system, metricKey] = systemMetricKey.split("-");
+    return { system, metricKey };
+  }
```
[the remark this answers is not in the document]
terryttsai: Yeah and the static methods make for a clear distinction in how they're more like utility functions!

terryttsai (approved):
This is looking awesome!

PR #129, opened 2022-10-31
[Agency Dashboard] Download metric data using feed data

terryttsai:
## Description of the change

So accessing the feed endpoint for an agency like `http://localhost:5001/feed/147` gives us a result like this:

```
system,metric_name,metric_url
LAW_ENFORCEMENT,arrests,/feed/147?system=LAW_ENFORCEMENT&metric=arrests
LAW_ENFORCEMENT,arrests_by_gender,/feed/147?system=LAW_ENFORCEMENT&metric=arrests_by_gender
LAW_ENFORCEMENT,arrests_by_race,/feed/147?system=LAW_ENFORCEMENT&metric=arrests_by_race
LAW_ENFORCEMENT,arrests_by_type,/feed/147?system=LAW_ENFORCEMENT&metric=arrests_by_type
LAW_ENFORCEMENT,calls_for_service,/feed/147?system=LAW_ENFORCEMENT&metric=calls_for_service
LAW_ENFORCEMENT,calls_for_service_by_type,/feed/147?system=LAW_ENFORCEMENT&metric=calls_for_service_by_type
LAW_ENFORCEMENT,reported_crime,/feed/147?system=LAW_ENFORCEMENT&metric=reported_crime
LAW_ENFORCEMENT,reported_crime_by_type,/feed/147?system=LAW_ENFORCEMENT&metric=reported_crime_by_type
```

I get an array of metric_urls, and I can call the url at each line to download the data feed for that url.

In the dashboard, I implemented a button that first downloads the overall feed information, then uses regex to find the lines that are relevant to that dashboard (for example, for the `LAW_ENFORCEMENT_ARRESTS` metric, look for lines that contain the words "LAW", "ENFORCEMENT", and "ARRESTS" and for the lines that match, download the data feed for each line. In this case, it would download 4 files as "arrests.txt", "arrests_by_gender.txt", "arrests_by_race.txt", and "arrests_by_type.txt".

This is a quick sketch and prone to breaking, but it gets us a working data download button quick.

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/DashboardView.tsx
+++ b/agency-dashboard/src/DashboardView.tsx
@@ -65,13 +65,8 @@ const DashboardView = () => {
   };
 
   const downloadFeedData = async (queryStr: string) => {
-    const result = (await request({
-      path: `/feed/${agencyId}${queryStr}`,
-      method: "GET",
-    })) as Response;
-    const blob = await result.blob();
-    const a = document.createElement("a");
-    a.href = URL.createObjectURL(blob);
+    var a = document.createElement("a");
+    a.href = `/feed/${agencyId}${queryStr}`;
     a.setAttribute("download", "data.txt");
     a.click();
   };
--- a/agency-dashboard/src/DashboardView.tsx
+++ b/agency-dashboard/src/DashboardView.tsx
@@ -64,10 +64,10 @@ const DashboardView = () => {
     return text;
   };
 
-  const downloadFeedData = async (queryStr: string) => {
+  const downloadFeedData = async (queryStr: string, filename: string) => {
     var a = document.createElement("a");
     a.href = `/feed/${agencyId}${queryStr}`;
-    a.setAttribute("download", "data.txt");
+    a.setAttribute("download", `${filename}.txt`);
     a.click();
   };
 
@@ -78,7 +78,7 @@ const DashboardView = () => {
     const lines = feedOverview.match(regex);
 
     lines?.forEach((line) => {
-      downloadFeedData(`?${line.split("?")[1]}`);
+      downloadFeedData(`?${line.split("?")[1]}`, line.split(",")[1]);
     });
   };
 
--- a/agency-dashboard/src/DashboardView.tsx
+++ b/agency-dashboard/src/DashboardView.tsx
@@ -65,7 +65,7 @@ const DashboardView = () => {
   };
 
   const downloadFeedData = async (queryStr: string, filename: string) => {
-    var a = document.createElement("a");
+    const a = document.createElement("a");
     a.href = `/feed/${agencyId}${queryStr}`;
     a.setAttribute("download", `${filename}.txt`);
     a.click();
--- a/agency-dashboard/src/DashboardView.tsx
+++ b/agency-dashboard/src/DashboardView.tsx
@@ -20,8 +20,14 @@ import { observer } from "mobx-react-lite";
 import { useEffect } from "react";
 import { useLocation, useNavigate, useParams } from "react-router-dom";
 
-import { Container, MetricTitle } from "./DashboardView.styles";
+import {
+  Container,
+  DownloadFeedButton,
+  DownloadFeedButtonContainer,
+  MetricTitle,
+} from "./DashboardView.styles";
 import { useStore } from "./stores";
+import { request } from "./utils/networking";
 
 const DashboardView = () => {
   const navigate = useNavigate();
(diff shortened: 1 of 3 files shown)


agency-dashboard/src/DashboardView.tsx
```diff
@@ -49,6 +55,38 @@ const DashboardView = () => {
...
+
+  const downloadFeedData = async (queryStr: string) => {
+    const result = (await request({
+      path: `/feed/${agencyId}${queryStr}`,
+      method: "GET",
+    })) as Response;
+    const blob = await result.blob();
+    const a = document.createElement("a");
+    a.href = URL.createObjectURL(blob);
+    a.setAttribute("download", "data.txt");
```
terryttsai, line 75: Could change the filename to equal the name of the csv file (like "arrests", "arrests_by_type", etc)

terryttsai: Edit: implemented downloading by file name, right now it saves as txt but maybe should save as csv?

lilidworkin (commented):
@terryttsai this looks good! I wonder if we actually have to get the feed overview first though? Can we just directly construct the URL with the system and metric name? e.g. `/feed/{agency_name}?system={system}&metric={metric_key}`? That would save us a request to get the overview, and would also mean we didn't have to do any fragile RegEx string matching.

I think we probably do want the download format to be csv. If you just set the filename to `.csv`, does that work? Or else we might need to change the headers on the backend to `feed_response.headers["Content-type"] = "text/csv"` instead of "text/plain" which it currently is.

terryttsai:
I thought about constructing the URL with the system and metric name, I think I would have to define a set of mappings like given the key "LAW_ENFORCEMENT_ARRESTS", look up the feed metrics "arrests", "arrests_by_gender", "arrests_by_race", "arrests_by_type".

Since I don't think the backend gives me that information right now, maybe the backend could give me that information in the published_datapoints backend? I could define it the mapping in the frontend as well?

lilidworkin:
@terryttsai ah right, good point! I wasn't thinking it through fully. Yea, I don't see why you couldn't change the backend to give you that information! We should definitely feel free to adjust the backend for this -- what I have for the /feed endpoint right now is really just a prototype.

I think the way you'd do it is using our SYSTEM_TO_METRICFILES dictionary (see metricfile_registry.py) and then doing the following:

from collections import defaultdict
system_and_metric_key_to_metricfiles = defaultdict(list)
for system, metricfiles in SYSTEM_TO_METRICFILES.items():
   for metricfile in metricfiles:
        system_and_metric_key_to_metricfiles[(system, metricfile.definition.key)].append(metricfile.canonical_filename)


That *should* give you a mapping that looks like {(LAW_ENFORCEMENT, ARRESTS): [arrests, arrests_by_gender, ... etc]}.

I don't think we have to do this, necessarily -- what you have is probably fine -- but if you want to give this a shot, it might be good backend/Python practice! If you do try it, please do let me know if you have any questions or if you want to pair on it together, etc.

terryttsai:
> what you have is probably fine

As in just using regex to search for which files to download? It seems fine though it's possible there's some case out there where this doesn't work? It doesn't sound hard to do what you're describing in the backend and put it in the published_datapoints endpoint, we're already using it to give not just datapoints but also `dimension_names_by_metric_and_disaggregation` metadata, so I could just add additional metadata for the system to metric files mapping there?

agency-dashboard/src/DashboardView.tsx
```diff
@@ -49,6 +55,33 @@ const DashboardView = () => {
...
+    const a = document.createElement("a");
+    a.href = `/feed/${agencyId}${queryStr}`;
+    a.setAttribute("download", `${filename}.txt`);
+    a.click();
+  };
+
+  const downloadData = async () => {
+    const feedOverview = await fetchDataFeedOverview();
+    const metricStr = metricKey?.split("_").join(".*");
+    const regex = new RegExp(`^.*${metricStr}.*$`, "gim");
```
mxosman, line 77: super nit: is there a way to describe what this regex is in the variable name?

agency-dashboard/src/DashboardView.tsx
```diff
@@ -49,6 +55,33 @@ const DashboardView = () => {
...
+    })) as Response;
+    const text = await result.text();
+    return text;
+  };
+
+  const downloadFeedData = async (queryStr: string, filename: string) => {
+    const a = document.createElement("a");
+    a.href = `/feed/${agencyId}${queryStr}`;
+    a.setAttribute("download", `${filename}.txt`);
+    a.click();
```
mxosman, line 71: nit: could clean this up by adding a.remove() at the end (unless it affects the download)

lilidworkin:
yep that makes sense to me @terryttsai ! Thanks for exploring this avenue though, good to know that you can be our resident RegEx expert :D

agency-dashboard/src/DashboardView.tsx
terryttsai: Good point!

PR #132, opened 2022-11-01
Metric Configuration Store: Refactor & Connect UI Components to the Store (2/2)

### LINKED ISSUE #134: [Frontend][Publisher] Refactor & Connect Metric Config UI to Store

mxosman:
## Description of the change

Equip with a new mobx data store, this PR refactors and connects all Metric Configuration & related UI components to the store. This should be a no-op refactor with regards to behavior (everything here should behave exactly the same as it does in staging). 

## Related issues

Closes #134 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -87,6 +87,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
       Object.keys(
         dimensions[systemMetricKey][activeDisaggregationKey]
       )) as string[];
+
     const metricDisplayName = metrics[systemMetricKey]?.label;
 
     const metricEnabled = Boolean(metrics[systemMetricKey]?.enabled);
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -68,7 +68,6 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
     saveAndUpdateMetricSettings,
   }): JSX.Element => {
     const { metricConfigStore } = useStore();
-
     const { activeMetricKey } = metricConfigStore;
     const systemMetricKey = MetricConfigStore.getSystemMetricKey(
       metricConfigStore.activeMetricKey as string,
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -70,8 +70,8 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
     const { metricConfigStore } = useStore();
     const { activeMetricKey } = metricConfigStore;
     const systemMetricKey = MetricConfigStore.getSystemMetricKey(
-      metricConfigStore.activeMetricKey as string,
-      metricConfigStore.activeSystem as string
+      metricConfigStore.activeSystem as string,
+      metricConfigStore.activeMetricKey as string
     );
     const metricDisplayName = metricConfigStore.metrics[systemMetricKey]?.label;
     const metricEnabled = Boolean(
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -38,7 +38,6 @@ import {
   MetricConfigurationContainer,
   MetricDisaggregations,
   MetricOnOffWrapper,
-  MetricSettings,
   RadioButtonGroupWrapper,
   Subheader,
 } from ".";
@@ -52,7 +51,6 @@ type MetricConfigurationProps = {
   setActiveDisaggregationKey: React.Dispatch<
     React.SetStateAction<string | undefined>
   >;
-  saveUpdatedMetricSettings: (updatedSetting: MetricSettings) => void;
 };
 
 export const Configuration: React.FC<MetricConfigurationProps> = observer(
@@ -61,7 +59,6 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
     setActiveDimensionKey,
     activeDisaggregationKey,
     setActiveDisaggregationKey,
-    saveUpdatedMetricSettings,
   }): JSX.Element => {
     const { metricConfigStore } = useStore();
     const {
@@ -74,6 +71,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
       updateMetricEnabledStatus,
       updateDisaggregationEnabledStatus,
       updateDimensionEnabledStatus,
+      saveMetricSettings,
     } = metricConfigStore;
 
     const systemMetricKey = getActiveSystemMetricKey();
@@ -129,7 +127,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
                   activeMetricKey as string,
                   true
                 );
-                saveUpdatedMetricSettings(updatedSetting);
+                saveMetricSettings(updatedSetting);
               }}
             />
             <BinaryRadioButton
@@ -145,7 +143,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
                   activeMetricKey as string,
                   false
                 );
-                saveUpdatedMetricSettings(updatedSetting);
+                saveMetricSettings(updatedSetting);
               }}
             />
           </RadioButtonGroupWrapper>
@@ -204,7 +202,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
                                   disaggregationKey,
                                   !currentDisaggregation.enabled
                                 );
-                              saveUpdatedMetricSettings(updatedSetting);
+                              saveMetricSettings(updatedSetting);
                             }}
                           />
                           <BlueCheckIcon
@@ -254,7 +252,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
                             dimensionKey,
                             !currentDimension.enabled
                           );
-                          saveUpdatedMetricSettings(updatedSetting);
+                          saveMetricSettings(updatedSetting);
                         }}
                       />
                       <BlueCheckIcon
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -150,7 +150,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
         </MetricOnOffWrapper>
 
         {/* Breakdowns */}
-        {activeDisaggregationKey && (
+        {activeDisaggregationKey && activeDisaggregationKeys?.length > 0 && (
           <MetricDisaggregations enabled={metricEnabled}>
             <BreakdownHeader>Breakdowns</BreakdownHeader>
             <Subheader>
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -19,7 +19,6 @@ import { observer } from "mobx-react-lite";
 import React, { useEffect } from "react";
 
 import { useStore } from "../../stores";
-import MetricConfigStore from "../../stores/MetricConfigStore";
 import { removeSnakeCase } from "../../utils";
 import { ReactComponent as RightArrowIcon } from "../assets/right-arrow.svg";
 import blueCheck from "../assets/status-check-icon.png";
@@ -46,6 +45,9 @@ import {
 
 type MetricConfigurationProps = {
   activeDimensionKey: string | undefined;
+  setActiveDimensionKey: React.Dispatch<
+    React.SetStateAction<string | undefined>
+  >;
   activeDisaggregationKey: string | undefined;
   setActiveDisaggregationKey: React.Dispatch<
     React.SetStateAction<string | undefined>
@@ -54,9 +56,6 @@ type MetricConfigurationProps = {
     updatedSetting: MetricSettings,
     debounce?: boolean
   ) => void;
-  setActiveDimensionKey: React.Dispatch<
-    React.SetStateAction<string | undefined>
-  >;
 };
 
 export const Configuration: React.FC<MetricConfigurationProps> = observer(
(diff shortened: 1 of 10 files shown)


publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -152,93 +216,121 @@ class MetricConfigStore {
...
+                multiple_choice_options: context.multiple_choice_options,
+              }
+            );
+          });
         });
       });
-    });
-  }
+    } catch (error) {
+      return new Error(error as string);
```
mxosman, line 306: Added a try/catch here since `getMetricSettings` can throw an error - this will catch it and return it as an Error object to the UI layer to be handled.

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -152,93 +216,121 @@ class MetricConfigStore {
...
       this.metrics[systemMetricKey] = {};
     }
 
+    /** If provided, add metadata required for rendering */
+    if (metadata) {
+      this.metrics[systemMetricKey].label = metadata.label;
+      this.metrics[systemMetricKey].description = metadata.description;
+      this.metrics[systemMetricKey].frequency =
+        metadata.frequency as ReportFrequency;
+    }
```
mxosman, line 332: This is where we add the necessary metadata required for rendering. During the store's initialization, an optional `metadata` object is passed in and this method loads those details on the metric object we use to update and render from.

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -337,6 +460,38 @@ class MetricConfigStore {
...
+
+      if (isLastDimensionDisabled) {
+        this.disaggregations[systemMetricKey][disaggregationKey].enabled =
+          false;
+      }
+
+      if (isDisaggregationDisabledAndOneDimensionReEnabled) {
+        this.disaggregations[systemMetricKey][disaggregationKey].enabled = true;
+      }
+    }
```
mxosman, line 493: Special handling for when last dimension is turned off and when everything in a breakdown is turned off and you're turning back on one dimension.

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -298,6 +400,26 @@ class MetricConfigStore {
...
+     */
+    if (!metadata && this.dimensions[systemMetricKey]?.[disaggregationKey]) {
+      Object.keys(this.dimensions[systemMetricKey][disaggregationKey]).forEach(
+        (dimensionKey) => {
+          this.dimensions[systemMetricKey][disaggregationKey][
+            dimensionKey
+          ].enabled = enabledStatus;
+        }
+      );
+    }
```
mxosman, line 421: Special handling for when a disaggregation is disabled/enabled.

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -102,16 +119,59 @@ class MetricConfigStore {
...
+            label?: Metric["label"];
+            description?: Metric["description"];
+            frequency?: Metric["frequency"];
+          };
+        }[]
+      );
+
+      return metrics;
+    }
+  };
```
mxosman, line 172: This felt useful to have to render a list of metrics by system.

publisher/src/pages/Settings.tsx
```diff
@@ -46,31 +46,17 @@ const Settings = () => {
...
       <ContentDisplay>
         {activeMenuItem === "Your Account" && <AccountSettings />}
         {activeMenuItem === "Uploaded Files" && <UploadedFiles />}
-        {activeMenuItem === "Metric Configuration" && (
-          <MetricConfiguration
-            activeMetricKey={activeMetricKey}
-            setActiveMetricKey={setActiveMetricKey}
-            setListOfMetrics={setListOfMetrics}
-          />
-        )}
```
mxosman, line 73: 🧹 🧹 🧹

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -52,413 +48,71 @@ import {
...
-          [updatedSetting.key]: {
-            ...prev[metricKey],
-            contexts: updatedContext,
-          },
-        };
-      }
-
-      return prev;
-    });
-  };
```
mxosman, line 402: BYE! 👋 👋 👋

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -52,413 +48,71 @@ import {
...
-    updatedSetting: MetricSettings,
-    debounce?: boolean
-  ) => {
-    updateMetricSettings(typeOfUpdate, updatedSetting);
-    if (debounce) {
-      debouncedSave(updatedSetting);
-    } else {
-      saveMetricSettings(updatedSetting);
-    }
-  };
```
mxosman, line 429: So, we don't need both save functions - the only component so far that has ever used the debouncing is the ContextConfiguration (text field context), so it made sense to just have the `debouncedSave` function live there and clean up the save function.

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -47,222 +44,247 @@ import {
...
-  useEffect(
-    () => {
-      const updatedDisaggregation =
-        activeDisaggregation &&
-        filteredMetricSettings[activeMetricKey]?.disaggregations?.find(
-          (disaggregation) => disaggregation.key === activeDisaggregation.key
-        );
+    const activeDisaggregationKeys =
+      disaggregations[systemMetricKey] &&
+      Object.keys(disaggregations[systemMetricKey]);
```
mxosman, line 83: Not sure if this was wise, but I decided I just needed an array of keys and from there, with our structure (`{ [key: string]: obj }`), I have quick access to the key for components that need it and can quickly look up the object I need to get the rest of the details.

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -47,222 +44,247 @@ import {
...
-      </MetricOnOffWrapper>
+    useEffect(
+      () => {
+        if (activeDisaggregationKeys)
+          setActiveDisaggregationKey(activeDisaggregationKeys[0]);
+        setActiveDimensionKey(undefined);
+      },
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      [systemMetricKey]
+    );
```
mxosman, line 102: Whenever a new metric is selected, reset the disaggregation tabs so the first one is open.

publisher/src/components/MetricConfiguration/Configuration.tsx
mxosman: Could've used `Object.entries` and used that multi-dimensional array, but this felt simple enough. No strong feelings!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -471,14 +125,16 @@ export const MetricConfiguration: React.FC<{
       // return disposer so it is cleaned up if it never runs
       reaction(
         () => userStore.currentAgencyId,
-        async (currentAgencyId, previousAgencyId) => {
+        async (_, previousAgencyId) => {
           if (previousAgencyId !== undefined) {
             setIsLoading(true);
```
terryttsai, line 130: Perhaps the body of the function here can be reused in the previous useEffect? Basically can we create one function that initializes the store and sets the active system and metric?

publisher/src/components/MetricConfiguration/ContextConfiguration.tsx
```diff
@@ -32,189 +33,185 @@ import {
...
+                    <BinaryRadioButton
+                      type="radio"
+                      id={`${contextKey}-yes`}
+                      name={contextKey}
+                      label="Yes"
+                      value="yes"
+                      checked={currentContext.value === "yes"}
+                      onChange={() => {
+                        const updatedSetting = updateContextValue(
+                          activeSystem as string,
```
terryttsai, line 93: Instead of "as string", can the argument type just be defined as MetricSystem?

publisher/src/components/MetricConfiguration/ContextConfiguration.tsx
```diff
@@ -32,189 +33,185 @@ import {
...
+                          activeMetricKey as string,
```
terryttsai, line 94: can this be undefined?

terryttsai: Maybe the function can just throw an error or log a warning if undefined?

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -47,222 +44,248 @@ import {
...
+      Object.keys(disaggregations[systemMetricKey]);
 
-  useEffect(
-    () => {
-      const updatedDisaggregation =
-        activeDisaggregation &&
-        filteredMetricSettings[activeMetricKey]?.disaggregations?.find(
-          (disaggregation) => disaggregation.key === activeDisaggregation.key
-        );
+    const activeDimensionKeys = (activeDisaggregationKey &&
```
terryttsai, line 85: Could write this as
```
    const activeDimensionKeys = (activeDisaggregationKey &&
      dimensions[systemMetricKey]?.[activeDisaggregationKey] ?
      Object.keys(
        dimensions[systemMetricKey][activeDisaggregationKey]
      ) : []);
```

and then you don't need to cast types

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
```diff
@@ -37,200 +33,255 @@ import {
...
-  activeDisaggregation: MetricDisaggregations | undefined;
-  contexts: MetricContext[];
-  saveAndUpdateMetricSettings: (
-    typeOfUpdate: MetricSettingsUpdateOptions,
-    updatedSetting: MetricSettings,
-    debounce?: boolean
-  ) => void;
+  activeDimensionKey: string | undefined;
+  activeDisaggregationKey: string | undefined;
+  saveUpdatedMetricSettings: (updatedSetting: MetricSettings) => void;
```
terryttsai, line 42: I wonder if it's worth making the store handle showing the error toast, then we wouldn't need to pass this in as a prop?

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
```diff
@@ -37,200 +33,255 @@ import {
...
+        dimensionDefinitionSettings[systemMetricKey][activeDisaggregationKey][
+          activeDimensionKey
+        ]
+      )) as string[];
+
+    const activeSettingsKeys = isMetricDefinitionSettings
+      ? metricDefinitionSettingsKeys
+      : dimensionDefinitionSettingsKeys;
+
+    const selectionOptions: MetricConfigurationSettingsOptions[] = [
```
terryttsai, line 95: Can we define MetricConfigurationSettingsOptions as 
```
export const MetricConfigurationSettingsOptions = ["Yes", "No", "N/A"] as const;
export type MetricConfigurationSettingsOptions = typeof MetricConfigurationSettingsOptions;
```

so we have both the array of strings as well as the type?

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -52,413 +48,71 @@ import {
...
 
-export type MetricSettingsUpdateOptions =
-  | "METRIC"
-  | "DISAGGREGATION"
-  | "DIMENSION"
-  | "CONTEXT"
-  | "METRIC_SETTING"
-  | "DIMENSION_SETTING";
-
 export type MetricSettings = {
```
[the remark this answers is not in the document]
mxosman: It could!

publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
mxosman: Great call!

publisher/src/components/MetricConfiguration/ContextConfiguration.tsx
mxosman: Good question - activeSystem could technically be `undefined` (before the initialization step) - though the initialization happens and it is defined as the first system before render this component. I'll have to rework the function a bit to handle the undefined - if I specify AgencySystems, it'll know it can also be undefined and if I include undefined in these functions it throws a cascade of TS errors. Maybe all the updater functions can short circuit in the off chance undefined passes through? Any suggestions?

publisher/src/components/MetricConfiguration/ContextConfiguration.tsx
mxosman: Yeah - great call - this is tricky. Technically they can, but things are initialized by the time it gets to this point. 

Let me think about this - if we update the function to throw an error if undefined, then we'll need an extra step of checking if `updatedSetting` is undefined before saving, right?

publisher/src/components/MetricConfiguration/Configuration.tsx
mxosman: Nice!

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
mxosman: Yeah - I was back and forth on this. Error toast felt more of a responsibility of the UI layer - but if it's in the store, then we get the benefit of not having to prop drill this. Let me try showing the toast in the store - that might work out nicer.

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
mxosman: That's a good idea - I was just using the one in the `common` repo and wanted to avoid having to update that repo in this PR if it wasn't necessary.

I tried what you implemented above and it turned the type into a readonly array - wasn't sure how to shake this TS warning that came as a result:
Screen Shot 2022-11-02 at 6 40 17 PM

Any ideas?

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
mxosman: Feels so much nicer centralizing the save function to the store! Thank you for all of these fantastic suggestions!

publisher/src/components/MetricConfiguration/ContextConfiguration.tsx
mxosman: Worked it out in this commit: [ed985e5](https://github.com/Recidiviz/justice-counts/pull/132/commits/ed985e5bd31081d7c07409f72567d7cd4af03033)

Thank you!

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
mxosman: Let me think about this a little more - I might have an idea of what's going on.

mxosman: Hmm. I do like this pattern and have used it in the past in the app, but as I keep exploring it - it feels like more trouble than its worth. This required updating everything that's typed as `MetricConfigurationSettingsOptions` and was running into the same compatibility issues as above.

mxosman: OK! I think I figured out what was going on - it needs to be used like this to work as intended:

```
export const metricConfigurationSettingsOptions = ["Yes", "No", "N/A"] as const;
export type MetricConfigurationSettingsOptions =
  typeof metricConfigurationSettingsOptions[number];
```

mxosman: [b9418ff](https://github.com/Recidiviz/justice-counts/pull/132/commits/b9418ffdd8a95f4bc7859874a4972ce7ca6f52e7)

Thanks for this optimization! This is cleaning up very nicely!

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -123,11 +123,11 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
               checked={metricEnabled}
               onChange={() => {
                 const updatedSetting = updateMetricEnabledStatus(
-                  activeSystem as string,
+                  activeSystem,
```
terryttsai, line 126: I see, so originally we were doing type casting here because we're only using this if an activeSystem and activeMetricKey are set in the store but the types could be undefined in the store. 

I guess we'll be moving away from setting activeSystem and activeMetricKey in the store anyways so actually I'd be cool with just wrapping this in a conditional like this:

```
              onChange={() => {
                if (activeSystem && activeMetricKey) {
                  const updatedSetting = updateMetricEnabledStatus(
                    activeSystem,
                    activeMetricKey,
                    true
                  );
                  saveMetricSettings(updatedSetting);
                }
              }}
```

Or we could keep the type casting since we expect them to be strings? Sorry I didn't give much thought to this yesterday!

terryttsai: My preference is to only call this `updateMetricEnabledStatus` if we have the required information otherwise don't call it

terryttsai (approved):
Thanks for addressing all the feedback! Left a small comment but everything looks good!

publisher/src/components/MetricConfiguration/Configuration.tsx
mxosman: Ah, OK. Makes sense - let me go back and revert all of those changes.

mxosman:
> Thanks for addressing all the feedback! Left a small comment but everything looks good!

Thank you for all of your help with this - this already feels so much better to interface with!

publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
terryttsai: Thanks for looking into this!

PR #143, opened 2022-11-09
Race & Ethnicity: Logic & Functionality (2/2)

### LINKED ISSUE #138: [Frontend][Publisher] Race & Ethnicities: Logic & Functionality (2/2)

mxosman:
## Description of the change

Implements the required functionality of R&E components according to the updated specification for Race and Ethnicity.

A brief description of the 3 core states users can find themselves in--

There are 3 distinct states (outlined below) that dictate the maximum available dimensions (of the 24 race & ethnicity dimensions) a user can enable/disable. 

### ******************************************State 1: Can Report Ethnicity******************************************

- Answered “Yes” to this question (**Does your case management system allow you to specify an individual’s ethnicity (Hispanic, Non-Hispanic, or Unknown) for this metric?**)
- User can report all 24 dimensions and can choose to disable a race if they are unable to report it
    - Turning off a single race is equivalent to turning off 3 dimensions because a single race has 3 ethnicities (HISPANIC/LATINO, NOT HISPANIC LATINO, UNKNOWN ETHNICITY)

### ******State 2: Cannot Report Ethnicity & Records Hispanic/Latino as Race******

- Answered “No” to this question (**Does your case management system allow you to specify an individual’s ethnicity (Hispanic, Non-Hispanic, or Unknown) for this metric?**)
- Answered “Yes” to (**Which of the following categories does that case management system capture for race?** > **Hispanic/Latino**)
- User can report 9 dimensions [Unknown Race x Hispanic/Latino ethnicity] and [All Races x Not Hispanic/Latino ethnicity]
- Note: if user disables the Unknown Race (equivalent to turning off 2 dimensions:  [**Unknown Race** x **Hispanic/Latino ethnicity**] AND [**Unknown Race** x **Not Hispanic/Latino ethnicity**]) after specifying “No” that they ******Cannot Report Ethnicity,****** then we automatically end up in **************State 3************** because this implies the user is not recording Hispanic/Latino as a race.

### ******State 3: Cannot Report Ethnicity & Does NOT Record Hispanic/Latino as Race******

- Answered “No” to this question (**Does your case management system allow you to specify an individual’s ethnicity (Hispanic, Non-Hispanic, or Unknown) for this metric?**)
- Answered “No” to (**Which of the following categories does that case management system capture for race?** > **Hispanic/Latino**)
- User can report 8 dimensions [All Races x Unknown Ethnicity]

---

### Demo:

https://user-images.githubusercontent.com/59492998/200963006-1699e901-c5b2-465e-bbaf-092e4b19381d.mov

---

### Quick notes for review ease:

  - All of the core functionality of this feature lives in the `MetricConfigStore`, `RaceEthnicities` and `RaceEthnicitiesBreakdown` components. 

  - Everything in the `MetricConfiguration` and `Configuration` conditionally renders the Race/Ethnicities components

  - The `RaceEthnicitiesGridStates` is an object that houses the different grid state truth tables. 

  - The rest are either styling adjustments, typings and adding a constant.

## Related issues

Closes #138 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
+++ b/publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
@@ -217,6 +217,10 @@ export const RaceEthnicitiesForm = observer(() => {
                       );
 
                       if (switchedGridStateUpdatedDimensions) {
+                        /** Add the updated dimension from disabling the Unknown race to the switchedGridStateUpdatedDimensions */
+                        switchedGridStateUpdatedDimensions.disaggregations[0].dimensions.push(
+                          ...updatedDimensions.disaggregations[0].dimensions
+                        );
                         return saveMetricSettings(
                           switchedGridStateUpdatedDimensions
                         );
--- a/publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
+++ b/publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
@@ -15,8 +15,10 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { observer } from "mobx-react-lite";
 import React from "react";
 
+import { useStore } from "../../stores";
 import { BinaryRadioButton } from "../Forms";
 import {
   Header,
@@ -27,7 +29,7 @@ import {
   RaceEthnicitiesDescription,
   RaceEthnicitiesDisplay,
   RaceEthnicitiesTitle,
-  races,
+  raceEthnicityGridStates,
   RaceSelection,
   RaceSelectionButton,
   RadioButtonGroupWrapper,
@@ -35,7 +37,48 @@ import {
   Subheader,
 } from ".";
 
-export const RaceEthnicitiesForm = () => {
+export const RaceEthnicitiesForm = observer(() => {
+  const { metricConfigStore } = useStore();
+  const {
+    ethnicitiesByRace,
+    updateAllRaceEthnicitiesToDefaultState,
+    updateRaceDimensions,
+    saveMetricSettings,
+  } = metricConfigStore;
+  const ethnicitiesByRaceArray = Object.entries(ethnicitiesByRace);
+
+  const canSpecifyEthnicity =
+    ethnicitiesByRaceArray.filter(([_, ethnicities]) => {
+      /** At least one Race where all three Ethnicities are enabled */
+      return (
+        ethnicities.Hispanic.enabled &&
+        ethnicities["Not Hispanic"].enabled &&
+        ethnicities["Unknown Ethnicity"].enabled
+      );
+    }).length > 0;
+
+  const specifiesHispanicAsRace =
+    ethnicitiesByRaceArray.filter(([race, ethnicities]) => {
+      /** Unknown Race has both Hispanic and Not Hispanic enabled */
+      return (
+        race === "Unknown" &&
+        ethnicities.Hispanic.enabled &&
+        ethnicities["Not Hispanic"].enabled
+      );
+    }).length > 0;
+
+  const determineCurrentState = () => {
+    if (canSpecifyEthnicity) {
+      return "CAN_SPECIFY_ETHNICITY";
+    }
+    if (specifiesHispanicAsRace) {
+      return "NO_ETHNICITY_HISPANIC_AS_RACE";
+    }
+    return "NO_ETHNICITY_HISPANIC_NOT_SPECIFIED";
+  };
+
+  const currentState = determineCurrentState();
+
   return (
     <RaceEthnicitiesContainer>
       <RaceEthnicitiesDisplay>
@@ -58,17 +101,35 @@ export const RaceEthnicitiesForm = () => {
           <RadioButtonGroupWrapper>
             <BinaryRadioButton
               type="radio"
-              id="metric-config-specify-ethnicity-yes"
-              name="metric-config-specify-ethnicity"
+              id="specify-ethnicity-yes"
+              name="specify-ethnicity"
               label="Yes"
               value="yes"
+              checked={canSpecifyEthnicity}
+              onChange={() => {
+                const updatedDimensions =
+                  updateAllRaceEthnicitiesToDefaultState(
+                    "CAN_SPECIFY_ETHNICITY",
+                    raceEthnicityGridStates
+                  );
+                saveMetricSettings(updatedDimensions);
+              }}
             />
             <BinaryRadioButton
               type="radio"
-              id="metric-config-specify-ethnicity-no"
-              name="metric-config-specify-ethnicity"
+              id="specify-ethnicity-no"
+              name="specify-ethnicity"
               label="No"
               value="no"
+              checked={!canSpecifyEthnicity}
+              onChange={() => {
+                const updatedDimensions =
+                  updateAllRaceEthnicitiesToDefaultState(
+                    "NO_ETHNICITY_HISPANIC_AS_RACE",
+                    raceEthnicityGridStates
+                  );
+                saveMetricSettings(updatedDimensions);
+              }}
             />
           </RadioButtonGroupWrapper>
         </SpecifyEthnicityWrapper>
@@ -82,17 +143,109 @@ export const RaceEthnicitiesForm = () => {
         </Subheader>
 
         <RaceContainer>
-          {races.map((race) => (
-            <Race key={race}>
-              <RaceDisplayName>{race}</RaceDisplayName>
+          {/* Hispanic/Latino as Race (if user cannot specify ethnicity) */}
+          {!canSpecifyEthnicity && (
+            <Race>
+              <RaceDisplayName>Hispanic/Latino</RaceDisplayName>
               <RaceSelection>
-                <RaceSelectionButton>No</RaceSelectionButton>
-                <RaceSelectionButton>Yes</RaceSelectionButton>
+                <RaceSelectionButton
+                  selected={!specifiesHispanicAsRace}
+                  onClick={() => {
+                    const updatedDimensions =
+                      updateAllRaceEthnicitiesToDefaultState(
+                        "NO_ETHNICITY_HISPANIC_NOT_SPECIFIED",
+                        raceEthnicityGridStates
+                      );
+                    saveMetricSettings(updatedDimensions);
+                  }}
+                >
+                  No
+                </RaceSelectionButton>
+                <RaceSelectionButton
+                  selected={specifiesHispanicAsRace}
+                  onClick={() => {
+                    const updatedDimensions =
+                      updateAllRaceEthnicitiesToDefaultState(
+                        "NO_ETHNICITY_HISPANIC_AS_RACE",
+                        raceEthnicityGridStates
+                      );
+                    saveMetricSettings(updatedDimensions);
+                  }}
+                >
+                  Yes
+                </RaceSelectionButton>
               </RaceSelection>
             </Race>
-          ))}
+          )}
+
+          {/* Races (Enable/Disable) */}
+          {Object.entries(ethnicitiesByRace).map(([race, ethnicities]) => {
+            const raceEnabled = Boolean(
+              Object.values(ethnicities).find((ethnicity) => ethnicity.enabled)
+            );
+
+            return (
+              <Race key={race}>
+                <RaceDisplayName>{race}</RaceDisplayName>
+                <RaceSelection>
+                  <RaceSelectionButton
+                    selected={!raceEnabled}
+                    onClick={() => {
+                      let switchedGridStateUpdatedDimensions;
+                      /**
+                       * When Unknown Race is disabled in NO_ETHNICITY_HISPANIC_AS_RACE state, we automatically switch
+                       * to the NO_ETHNICITY_HISPANIC_NOT_SPECIFIED state because the Unknown Race (Hispanic/Latino Ethnicity)
+                       * dimension is the only dimension an agency can specify their numbers for Hispanic/Latino as a Race (while
+                       * in the NO_ETHNICITY_HISPANIC_AS_RACE state).
+                       */
+                      if (
+                        race === "Unknown" &&
+                        currentState === "NO_ETHNICITY_HISPANIC_AS_RACE"
+                      ) {
+                        switchedGridStateUpdatedDimensions =
+                          updateAllRaceEthnicitiesToDefaultState(
+                            "NO_ETHNICITY_HISPANIC_NOT_SPECIFIED",
+                            raceEthnicityGridStates
+                          );
+                      }
+
+                      const updatedDimensions = updateRaceDimensions(
+                        race,
+                        false,
+                        currentState,
+                        raceEthnicityGridStates
+                      );
+
+                      if (switchedGridStateUpdatedDimensions) {
+                        return saveMetricSettings(
+                          switchedGridStateUpdatedDimensions
+                        );
+                      }
+                      saveMetricSettings(updatedDimensions);
+                    }}
+                  >
+                    No
+                  </RaceSelectionButton>
+                  <RaceSelectionButton
+                    selected={raceEnabled}
+                    onClick={() => {
+                      const updatedDimensions = updateRaceDimensions(
+                        race,
+                        true,
+                        currentState,
+                        raceEthnicityGridStates
+                      );
+                      saveMetricSettings(updatedDimensions);
+                    }}
+                  >
+                    Yes
+                  </RaceSelectionButton>
+                </RaceSelection>
+              </Race>
+            );
+          })}
         </RaceContainer>
       </RaceEthnicitiesDisplay>
     </RaceEthnicitiesContainer>
   );
-};
+});
--- a/publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
+++ b/publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
@@ -15,8 +15,10 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { observer } from "mobx-react-lite";
 import React from "react";
 
+import { useStore } from "../../stores";
 import { ReactComponent as RightArrowIcon } from "../assets/right-arrow.svg";
 import {
   CalloutBox,
@@ -32,10 +34,12 @@ import {
   RaceEthnicitiesBreakdownContainer,
   RaceEthnicitiesRow,
   RaceEthnicitiesTable,
-  races,
 } from ".";
 
-export const RaceEthnicitiesGrid = () => {
+export const RaceEthnicitiesGrid: React.FC = observer(() => {
+  const { metricConfigStore } = useStore();
+  const { ethnicitiesByRace } = metricConfigStore;
+
   return (
     <RaceEthnicitiesBreakdownContainer>
       <CalloutBox>
@@ -43,7 +47,6 @@ export const RaceEthnicitiesGrid = () => {
           Answer the questions on the <span>Race and Ethnicity</span> form; the
           grid below will reflect your responses.
         </Description>
-
         <RightArrowIcon />
       </CalloutBox>
 
@@ -60,17 +63,20 @@ export const RaceEthnicitiesGrid = () => {
       </GridHeaderContainer>
 
       <RaceEthnicitiesTable>
-        {races.map((race, index) => (
+        {Object.entries(ethnicitiesByRace).map(([race, ethnicities]) => (
           <RaceEthnicitiesRow key={race}>
             <RaceCell>{race}</RaceCell>
             <EthnicitiesRow>
-              <EthnicityCell enabled={index % 1 === 0} />
-              <EthnicityCell />
-              <EthnicityCell />
+              {Object.values(ethnicities).map((ethnicity) => (
+                <EthnicityCell
+                  key={ethnicity.key}
+                  enabled={ethnicity.enabled}
+                />
+              ))}
             </EthnicitiesRow>
           </RaceEthnicitiesRow>
         ))}
       </RaceEthnicitiesTable>
     </RaceEthnicitiesBreakdownContainer>
   );
-};
+});
--- a/publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
+++ b/publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
@@ -1,0 +1,172 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2022 Recidiviz, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+export type StateKeys = keyof typeof raceEthnicityGridStates;
+
+export type RaceEthnicitiesGridStates = {
+  [state: string]: {
+    [race: string]: {
+      [ethnicity: string]: boolean;
+    };
+  };
+};
+
+/**
+ * Race x Ethnicity Grid States represent the maximum set of dimensions (of the 24 available
+ * Race & Ethnicity dimensions) a user can enter data for or disable/enable.
+ *
+ * There are 3 states that a user can fall under based on the following conditions:
+ * State 1 (`CAN_SPECIFY_ETHNICITY`): They record & can report each ethnicity per race (Hispanic/Latino, Not Hispanic/Latino, Unknown Ethnicity)
+ * State 2 (`NO_ETHNICITY_HISPANIC_AS_RACE`): They cannot record/report each ethnicity and record Hispanic/Latino as a Race
+ * State 3 (`NO_ETHNICITY_HISPANIC_NOT_SPECIFIED`): They cannot record/report each ethnicity and DO NOT record Hispanic/Latino as a Race
+ *
+ * This object serves as a truth table of the 3 states. The boolean represents whether or not that
+ * particular dimension is available to the user to enter data for or disable/enable.
+ *
+ * @example
+ * State 1 has all 24 available dimensions (Race x All 3 Ethnicities)
+ * State 2 has only 9 available dimensions (Race x Not Hispanic/Latino Ethnicity, Unknown Race x Hispanic/Latino Ethnicity) - the other 15 dimensions are disabled.
+ * State 3 has only 8 available dimensions (Race x Unknown Ethnicity) - the other 16 dimensions are disabled.
+ */
+export const raceEthnicityGridStates = {
+  CAN_SPECIFY_ETHNICITY: {
+    "American Indian / Alaskan Native": {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    Asian: {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    Black: {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    "More than one race": {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    "Native Hawaiian / Pacific Islander": {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    White: {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    Other: {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+    Unknown: {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": true,
+    },
+  },
+  NO_ETHNICITY_HISPANIC_AS_RACE: {
+    "American Indian / Alaskan Native": {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    Asian: {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    Black: {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    "More than one race": {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    "Native Hawaiian / Pacific Islander": {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    White: {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    Other: {
+      Hispanic: false,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
+    },
+    Unknown: {
+      Hispanic: true,
+      "Not Hispanic": true,
+      "Unknown Ethnicity": false,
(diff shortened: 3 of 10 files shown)


publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
```diff
@@ -0,0 +1,172 @@
...
+ * This object serves as a truth table of the 3 states. The boolean represents whether or not that
+ * particular dimension is available to the user to enter data for or disable/enable.
+ *
+ * @example
+ * State 1 has all 24 available dimensions (Race x All 3 Ethnicities)
+ * State 2 has only 9 available dimensions (Race x Not Hispanic/Latino Ethnicity, Unknown Race x Hispanic/Latino Ethnicity) - the other 15 dimensions are disabled.
+ * State 3 has only 8 available dimensions (Race x Unknown Ethnicity) - the other 16 dimensions are disabled.
+ */
+export const raceEthnicityGridStates = {
+  state1: {
```
mxosman, line 46: I'm trying to think of a clearer way to represent these states throughout the app.

Something kinda like this - but better? Clearer? Simpler? All caps, snakecase? orCamelCase? S.O.S

`CAN_SPECIFY_ETHNICITY_STATE`
`NO_ETHNICITY_HISPANIC_AS_RACE_STATE`
`NO_ETHNICITY_HISPANIC_NOT_SPECIFIED_STATE`

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+    saveMetricSettings,
+  } = metricConfigStore;
+
+  const [canSpecifyEthnicity, setCanSpecifyEthnicity] = useState<boolean>();
+  const [specifiesHispanicAsRace, setSpecifiesHispanicAsRace] =
+    useState<boolean>();
+
+  const inferCanSpecifyEthnicityFromDimensions =
+    Object.values(ethnicitiesByRace).filter((ethnicities) => {
+      /** All ethnicities are enabled */
```
lilidworkin, line 58: nit, maybe change comment to "At least one Race where all three Ethnicities are enabled."

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+      return (
+        ethnicities.Hispanic.enabled &&
+        ethnicities["Not Hispanic"].enabled &&
+        ethnicities["Unknown Ethnicity"].enabled
+      );
+    }).length > 0;
+
+  const inferSpecifiesHispanicAsRaceFromDimensions =
```
lilidworkin, line 66: Love these two "infer" functions -- super clean / clear

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+  useEffect(() => {
+    setCanSpecifyEthnicity(inferCanSpecifyEthnicityFromDimensions);
+    setSpecifiesHispanicAsRace(inferSpecifiesHispanicAsRaceFromDimensions);
+  }, [
+    inferCanSpecifyEthnicityFromDimensions,
+    inferSpecifiesHispanicAsRaceFromDimensions,
+  ]);
+
+  const determineCurrentState = () => {
```
lilidworkin, line 84: I actually think "state1" etc is fine to put in the codebase! This function makes it super clear what each state represents.

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+    }
+    if (!canSpecifyEthnicity && specifiesHispanicAsRace) {
+      return "state2";
+    }
+    return "state3";
+  };
+
+  const currentState = determineCurrentState();
+
+  const mergeSwitchingGridStatesAndDisablingUnknownRaceDimensions = (
```
lilidworkin, line 96: Maybe add a comment explaining in a little more detail what this guy is doing?

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+            <BinaryRadioButton
+              type="radio"
+              id="specify-ethnicity-yes"
+              name="specify-ethnicity"
+              label="Yes"
+              value="yes"
+              checked={canSpecifyEthnicity || false}
+              onChange={() => {
+                const updatedDimensions =
+                  updateAllRaceEthnicitiesToDefaultState(
```
lilidworkin, line 152: Such a clear function name!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+            return (
+              <Race key={race}>
+                <RaceDisplayName>{race}</RaceDisplayName>
+                <RaceSelection>
+                  <RaceSelectionButton
+                    selected={!raceEnabled}
+                    onClick={() => {
+                      let combinedUpdatedDimensions;
+
+                      if (race === "Unknown" && currentState === "state2") {
```
lilidworkin, line 237: Maybe add a comment explaining what's happening here?

publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
lilidworkin: I actually don't mind state 1/2/3, but I also like what you have above too! I think either is fine

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -426,6 +440,21 @@ class MetricConfigStore {
...
+
+      /**
+       * For Race & Ethnicities: when disaggregation is enabled/disabled, update the `ethnicitiesByRace` object
+       * with updated dimensions.
+       * */
+      if (
+        Object.values(this.ethnicitiesByRace).length > 0 &&
+        RACE_ETHNICITY_DISAGGREGATION_KEY === disaggregationKey
+      ) {
+        this.mapRaceEthnicitiesDimensionsToEthnicitiesByRace(
```
lilidworkin, line 452: Just to make sure I understand, this is happening when the Race/Ethnicity disaggregation as a whole is enabled/disabled? And/or when a particular breakdown in that disaggregation is enabled/disabled?

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -651,6 +682,125 @@ class MetricConfigStore {
...
+  ): UpdatedDisaggregation => {
+    const updatedDimensions = [] as UpdatedDimension[];
+
+    ethnicities.forEach((ethnicity) => {
+      /** No update if intended update matches the current state (e.g. enabling an already enabled dimension) */
+      if (this.ethnicitiesByRace[race][ethnicity].enabled === enabled) return;
+      /** No update if enabling a disabled dimension that is not available to the user to edit (determined by current grid state) */
+      if (
+        enabled &&
+        this.ethnicitiesByRace[race][ethnicity].enabled ===
```
lilidworkin, line 757: Nice, I love how using a map here makes the update so easy!

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -651,6 +682,125 @@ class MetricConfigStore {
...
+      )
+        return;
+
+      this.ethnicitiesByRace[race][ethnicity].enabled = enabled;
+      updatedDimensions.push(this.ethnicitiesByRace[race][ethnicity]);
+    });
+
+    /**
+     * When last race is disabled, the disaggregation is disabled
+     * When all races are off, and one race is re-enabled, the disaggregation is enabled
```
lilidworkin, line 768: Great comments

lilidworkin (commented):
@mxosman this is fantastic. I'm blown away by how clearly and elegantly you were able to write (and comment) this PR! All your variable and function names are so descriptive, your comments are great, everything reads so easily. 

I may take another slightly more detailed pass later, but overall I think everything here is really great and I barely have any comments at all :)

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+  }, [
+    inferCanSpecifyEthnicityFromDimensions,
+    inferSpecifiesHispanicAsRaceFromDimensions,
+  ]);
+
+  const determineCurrentState = () => {
+    if (canSpecifyEthnicity) {
+      return "state1";
+    }
+    if (!canSpecifyEthnicity && specifiesHispanicAsRace) {
```
terryttsai, line 88: Don't need `!canSpecifyEthnicity` since that's covered in the lines above

terryttsai: Though it would be nice to still have a code comment mentioning that the user doesn't specify ethnicity but specifies Hispanic as race

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
terryttsai: Yeah I'm fine with going with these numbers if there's a comment explaining what the different states are

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+      /** Unknown Race has both Hispanic and Not Hispanic enabled */
+      return (
+        race === "Unknown" &&
+        ethnicities.Hispanic.enabled &&
+        ethnicities["Not Hispanic"].enabled
+      );
+    }).length > 0;
+
+  useEffect(() => {
+    setCanSpecifyEthnicity(inferCanSpecifyEthnicityFromDimensions);
```
terryttsai, line 77: Is there a reason to set the two infers in react state? Can't they just be variables derived from the metric config store? (Or calculated mobx values in the metric config store?

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+          </Header>
+
+          <RadioButtonGroupWrapper>
+            <BinaryRadioButton
+              type="radio"
+              id="specify-ethnicity-yes"
+              name="specify-ethnicity"
+              label="Yes"
+              value="yes"
+              checked={canSpecifyEthnicity || false}
```
terryttsai, line 149: If `canSpecifyEthnicity` is a mobx calculated value, it can easily be a boolean and not "boolean | undefined" like it is right now

terryttsai:
https://www.loom.com/share/a58902a81b134cff90693c80ecb7dc42

This is looking amazing so far! Thanks for planning out and defining the 3 states!

Clicking through this, I couldn't click in the race & ethnicities dialog box to change my race & ethnicities settings, I had to click on the Race / Ethnicity tab, which is not intuitive to me.

Also, when I switch on/off Hispanic as race, it auto-reenables all the other races I had turned off?

publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
```diff
@@ -28,22 +30,49 @@ import {
...
+  } = metricConfigStore;
+
+  const systemMetricKey = getActiveSystemMetricKey();
+  const raceEthnicitiesDimensions =
+    dimensions[systemMetricKey]?.[RACE_ETHNICITY_DISAGGREGATION_KEY] &&
+    (Object.values(
+      dimensions[systemMetricKey][RACE_ETHNICITY_DISAGGREGATION_KEY]
+    ) as UpdatedDimension[]);
+
+  useEffect(
```
terryttsai, line 57: Generally I would avoid using component mount to set mobx state, could the logic from line 50-67 be moved into the MetricConfigStore?

publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
terryttsai: These state names sound good to me! No need for `STATE` at the end

publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
```diff
@@ -0,0 +1,172 @@
...
+      Hispanic: false,
+      "Not Hispanic": false,
+      "Unknown Ethnicity": true,
+    },
+    Other: {
+      Hispanic: false,
+      "Not Hispanic": false,
+      "Unknown Ethnicity": true,
+    },
+    Unknown: {
```
terryttsai, line 166: Love these defined constants!

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -119,6 +132,7 @@ class MetricConfigStore {
     this.disaggregations = {};
     this.dimensions = {};
     this.dimensionDefinitionSettings = {};
+    this.ethnicitiesByRace = {};
```
terryttsai, line 135: So `ethnicitiesByRace` are just the race dimensions but keyed by race and ethnicity? 

I think there should be one single source of truth for this so either they should not be a part of `this.dimensions`, or `ethnicitiesByRace` should be a mobx computed value that's computed from `dimensions`, which is what I do for my DatapointsStore.

publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
terryttsai: Looking into the whole `this.ethnicitiesByRace`, could this just be a mobx computed value from `this.dimensions`?

lilidworkin:
re: @terryttsai 's comments:

> Clicking through this, I couldn't click in the race & ethnicities dialog box to change my race & ethnicities settings, I had to click on the Race / Ethnicity tab, which is not intuitive to me.
Haha yea there's a discussion about this in Slack [here](https://recidiviz.slack.com/archives/C02VDN1LSSW/p1667944951435479). I very much agree and want the box to be clickable! Up to @mxosman if we do that now or after playtesting.

> Also, when I switch on/off Hispanic as race, it auto-reenables all the other races I had turned off?
Yea this is known behavior actually. @mxosman and I discussed it and I thought it was fine. @mxosman I imagine the issue is that it's a little tricky to preserve the race settings due to the complexity with Unknown race? If so, I think this is an edge case that I'm fine punting on.

publisher/src/components/MetricConfiguration/RaceEthnicitiesGridStates.ts
mxosman: Thank you both! I'll try it with the names above.

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: Love it! Thank you for that!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: Absolutely!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: 😅 Naming is so hard!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: Absolutely!

publisher/src/stores/MetricConfigStore.tsx
mxosman: Yes! This is for when the disaggregation as a whole is being enabled/disabled.

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: Great catch! Agreed!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: No particular reason - just amongst the first things that came to mind to solve this. I had them set in the useEffect so they always update based on the user's current state/selections. The selections for specifying an ethnicity and whether or not they specify Hispanic/Latino as a race are not things that we are saving in the backend - they exist temporarily and update based on user's selection/state. I haven't tried moving them to the store because it felt mostly UI.

Moving them to the store - I'm trying to think of how to keep them updated and in sync with the current state. I'll play around with it.

publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
mxosman: > Generally I would avoid using component mount to set mobx state, especially since this seems like data layer stuff. Could the logic from line 50-67 be moved into the MetricConfigStore? That way the view logic isn't setting data logic?

This is more/less initializing the whole `ethnicitiesByRace` on component mount - I feel like we do that for most other components, don't we?

> Looking into the whole `this.ethnicitiesByRace`, could this just be a mobx computed value derived from `this.dimensions`?

I think so? I haven't tried it as a mobx computed value - I'll try that! Might solve for your former comment as well.

publisher/src/stores/MetricConfigStore.tsx
mxosman: > @mxosman this is fantastic. I'm blown away by how clearly and elegantly you were able to write (and comment) this PR! All your variable and function names are so descriptive, your comments are great, everything reads so easily.
> 
> I may take another slightly more detailed pass later, but overall I think everything here is really great and I barely have any comments at all :)

Thank you, Lili! It's because you have all leveled me up - grateful to always be learning from you and the team!

mxosman:
> @mxosman this is fantastic. I'm blown away by how clearly and elegantly you were able to write (and comment) this PR! All your variable and function names are so descriptive, your comments are great, everything reads so easily.
> 
> I may take another slightly more detailed pass later, but overall I think everything here is really great and I barely have any comments at all :)

Thank you, Lili! It's because you have all leveled me up - grateful to always be learning from you and the team!

mxosman:
> > Also, when I switch on/off Hispanic as race, it auto-reenables all the other races I had turned off?
> 
> Yea this is known behavior actually. @mxosman and I discussed it and I thought it was fine. @mxosman I imagine the issue is that it's a little tricky to preserve the race settings due to the complexity with Unknown race? If so, I think this is an edge case that I'm fine punting on.

Good call out! Yeah - exactly, Lili - but, I can definitely give it some more thought! It feels like an edge case as I imagine someone might have an easier time turning the entire disaggregation off if they can't report any race vs go through each one like we do? But, it is definitely not the most optimal behavior for that situation.

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
terryttsai: If `canSpecifyEthnicity` is just defined as the result of `inferCanSpecifyEthnicityFromDimensions`, then it's automatically gonna be in sync with the store right?

terryttsai: since when `ethnicitiesByRace` updates, the react component will recalculate `inferCanSpecifyEthnicityFromDimensions`, no need to set that in react state right?

terryttsai: In general react state should only be used to keep track of some local component-specific state, we don't need to set additional react state computed off of mobx store observable values here.

publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
terryttsai: I don't see a reason that `ethnicitiesByRace` should be recalculated on component mount, it could easily become out of sync if we forget to update it when dimensions are updated, or it could not get set if we stop mounting this component.

terryttsai: Whereas if it's a computed value derived from the single source of truth, it'll automatically get updated when the single source of truth updates

terryttsai:
Yeah I'm totally fine with leaving the edge case alone, it's not a big deal to me at all!

publisher/src/components/MetricConfiguration/RaceEthnicitiesGrid.tsx
mxosman: That makes so much sense! I don't know why I overcomplicated it - just turned `ethnicitiesByRace` into a computed value and it cleaned up SO MUCH! We don't even need to move L50-67 - those can be deleted now! Seriously - thank you so much for this! This was a tough one to figure out and you simplified it so much!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: 100%! Thank you for that!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
mxosman: OK! I turned the `ethnicitiesByRace` into the computed value - and I see exactly what you're seeing! You're absolutely right! Thank you so much for this!

mxosman:
> > > Also, when I switch on/off Hispanic as race, it auto-reenables all the other races I had turned off?

> > Yea this is known behavior actually. @mxosman and I discussed it and I thought it was fine. @mxosman I imagine the issue is that it's a little tricky to preserve the race settings due to the complexity with Unknown race? If so, I think this is an edge case that I'm fine punting on.

> Yeah I'm totally fine with leaving the edge case alone, it's not a big deal to me at all!

After implementing the feedback, I played around with this and it turned out to be less tricky than I thought. Here's what it looks like - curious to hear your thoughts!

https://user-images.githubusercontent.com/59492998/201433248-3c30b64f-e22f-4c3d-adfa-93f2a82ea50f.mov

terryttsai:
The behavior in the video looks awesome! Thanks for taking a stab at it!

mxosman:
> > > > Also, when I switch on/off Hispanic as race, it auto-reenables all the other races I had turned off?
> 
> > > Yea this is known behavior actually. @mxosman and I discussed it and I thought it was fine. @mxosman I imagine the issue is that it's a little tricky to preserve the race settings due to the complexity with Unknown race? If so, I think this is an edge case that I'm fine punting on.
> 
> > Yeah I'm totally fine with leaving the edge case alone, it's not a big deal to me at all!
> 
> After implementing the feedback, I played around with this and it turned out to be less tricky than I thought. Here's what it looks like - curious to hear your thoughts!
> 

Let me think about this some more - there is a weird state you get into when you turn off the Unknown Race in the CAN_SPECIFY_ETHNICITY state and switch the Can Specify Ethnicity answer to No (technically you're supposed to jump to the 3rd state but it keeps some dimensions in the 2nd state)

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -651,6 +682,122 @@ class MetricConfigStore {
...
+      this.activeMetricKey
+    );
+    const dimensions = Object.values(
+      this.dimensions[systemMetricKey][RACE_ETHNICITY_DISAGGREGATION_KEY]
+    ) as UpdatedDimension[];
+    const ethnicitiesByRaceMap = dimensions.reduce(
+      (acc, dimension) => {
+        /** Dimension keys are `Race / Ethnicity` or `Race Description 1 / Race Description 2 / Ethnicity` */
+        const [ethnicity, ...race] = dimension.key.split(" / ").reverse();
+        const normalizedRace = race.reverse().join(" / ");
```
mxosman, line 700: I wasn't sure if there was a better way to do this? This was just the simplest that came to mind that was less verbose than conditionals. 

I was thinking since dimension keys can be `Race / Ethnicity` or `Race Description 1 / Race Description 2 / Ethnicity` - a quick way that allows use of the ...rest destructing is to split the key by " / " and reverse it. The first item in the array will be the `ethnicity`, and the remaining items (one or two) will be the `race` (in reverse for multi-description races). We do a corrective reverse on the `race` array so multi-description races are ordered properly.

I feel like in explaining this, this might not be the clearest way. Any recommendations or does this seem reasonable?

terryttsai: What are some examples of what you're trying to do? Do you want to change `Race / Ethnicity` to `Ethnicity / Race` and `Race Description 1 / Race Description 2 / Ethnicity` to `Ethnicity / Race Description 1 / Race Description 2`?

mxosman: Great question! So I'm just trying to extract the `race` string and the `ethnicity` string from the `dimension.key`. The key will be in the format `Race / Ethnicity` (e.g. `Black / Hispanic`) or `Race Description 1 / Race Description 2 / Ethnicity` (e.g. `American Indian / Alaskan Native / Not Hispanic`).

The logic above allows me to cover both cases and get back the `race` and `ethnicity` strings without conditionals and Regex - but not sure if it's clear.

terryttsai: To extract the ethnicity, you could split the string by "/" and get the last element of the array, that's the ethnicity, then prepend it to the beginning of a string then add in the remaining elements? Then you don't need to reverse twice?

terryttsai: What is the purpose of the string splitting? It's prone to future bugs so if it's for display purposes I wonder if the server can send us the right display names...

mxosman:
> > > > > Also, when I switch on/off Hispanic as race, it auto-reenables all the other races I had turned off?
> > 
> > 
> > > > Yea this is known behavior actually. @mxosman and I discussed it and I thought it was fine. @mxosman I imagine the issue is that it's a little tricky to preserve the race settings due to the complexity with Unknown race? If so, I think this is an edge case that I'm fine punting on.
> > 
> > 
> > > Yeah I'm totally fine with leaving the edge case alone, it's not a big deal to me at all!
> > 
> > 
> > After implementing the feedback, I played around with this and it turned out to be less tricky than I thought. Here's what it looks like - curious to hear your thoughts!
> 
> Let me think about this some more - there is a weird state you get into when you turn off the Unknown Race in the CAN_SPECIFY_ETHNICITY state and switch the Can Specify Ethnicity answer to No (technically you're supposed to jump to the 3rd state but it keeps some dimensions in the 2nd state)

OK, I think I figured it out and a lot of the edge cases covered - let me know your thoughts:


https://user-images.githubusercontent.com/59492998/201765415-82a1e48c-8cae-41dc-9595-52bc455bf7f2.mov

publisher/src/stores/MetricConfigStore.tsx
mxosman: > To extract the ethnicity, you could split the string by "/" and get the last element of the array, that's the ethnicity, then prepend it to the beginning of a string then add in the remaining elements? Then you don't need to reverse twice?

Correct - but you'd also then have to check the array length to get the race incase of multi-description race (e.g. `Black / Hispanic` becomes `["Black", "Hispanic"]` vs. `American Indian / Alaskan Native / Not Hispanic` which becomes `["American Indian", "Alaskan Native", "Not Hispanic"]`).

I was using the `, ...race` spread destructuring to bypass the need for a conditional, but needed to reverse it to get the last either 1 or 2 array items.

Example:

- dimension.key is `American Indian / Alaskan Native / Not Hispanic`
- Split at " / " = ["American Indian", "Alaskan Native", "Not Hispanic"]
- Reverse = ["Not Hispanic", "Alaskan Native", "American Indian"]
- Now the first item is always the `ethnicity` and the last 1 or 2 items (I don't have to check for that with ...spread) is the race reversed
- Normalize race by corrective reverse: ["Alaskan Native", "American Indian"] becomes  ["American Indian", "Alaskan Native"] which becomes `American Indian / Alaskan Native`

mxosman: > What is the purpose of the string splitting? It's prone to future bugs so if it's for display purposes I wonder if the server can send us the right display names...

This is to map the dimensions to the ethnicities by race object. I don't think backend needs to change the key structure since we can easily infer it on the frontend. 

I was just more wondering if this was an OK way to express this - or if there was a clearer way?

terryttsai: Can we just pop the last element off the array to get the ethnicity, and the remaining elements are the race(s)?

mxosman: Yes, of course! Wow, my brain! `pop` - I _finally_ have an excuse to use `pop`! Thank you, Terry! LMK if you would do this any other way - open to all ideas!

lilidworkin: I also don't love this string splitting, though I'm not opposed to keeping it. To echo @terryttsai  -- could we have the backend just change the response from:

```

                    {
                      "key": "NOT_HISPANIC_WHITE",
                      "label": "Not Hispanic / White",
                      "enabled": True,
                      "settings": [],
                    },
```

to 

```
                    {
                      "key": "NOT_HISPANIC_WHITE",
                      "label": "Not Hispanic / White",
                      "enabled": True,
                      "settings": [],
"race": "White",
"ethnicity": "Not Hispanic",
                    },
```

lilidworkin: I imagine this would be pretty quick for @nichelle-hall  to do? Will let you guys decide, the way you have it is fine @mxosman ,but agreed that string splitting on a display name is fragile because there's nothing guaranteeing that format.

terryttsai: Would love if the backend makes it explicit what's race or ethnicity!

mxosman: I have no strong feelings at all - just going off of our conversation of not needing the backend to make changes for this feature! Always happy to be flexible.

lilidworkin: @mxosman can you discuss with @nichelle-hall and see if she can knock it out? If she thinks it'll take more than an hour or so, not worth it

mxosman: Will do

nichelle-hall: Yes! This is doable. I already have some helper methods that would make that easy :)

nichelle-hall: [PR](https://github.com/Recidiviz/recidiviz-data/pull/16676) for this change is up!

mxosman: Thank you so much, Nichelle! 🙏

mxosman:
@terryttsai / @lilidworkin - whenever you have a moment, please let me know if there's any other feedback I can address in this PR (or in a follow up one). Thank you for all of the feedback so far!

mxosman:
> @terryttsai / @lilidworkin - whenever you have a moment, please let me know if there's('re?) any other feedback I can address in this PR (or in a follow up one). Thank you for all of the feedback so far!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -27,15 +29,56 @@ import {
...
+      return (
+        race === "Unknown" &&
+        ethnicities.Hispanic.enabled &&
+        ethnicities["Not Hispanic"].enabled
+      );
+    }).length > 0;
+
+  const determineCurrentState = () => {
+    if (canSpecifyEthnicity) {
+      return "CAN_SPECIFY_ETHNICITY";
```
lilidworkin, line 72: I like these string names!

publisher/src/components/MetricConfiguration/RaceEthnicitiesForm.tsx
```diff
@@ -82,17 +143,113 @@ export const RaceEthnicitiesForm = () => {
...
+            return (
+              <Race key={race}>
+                <RaceDisplayName>{race}</RaceDisplayName>
+                <RaceSelection>
+                  <RaceSelectionButton
+                    selected={!raceEnabled}
+                    onClick={() => {
+                      let switchedGridStateUpdatedDimensions;
+                      /**
+                       * When Unknown Race is disabled in NO_ETHNICITY_HISPANIC_AS_RACE state, we automatically switch
```
lilidworkin, line 196: Great comment!

lilidworkin (approved):
Looks great to me! Not sure if @terryttsai has any additional feedback but I'm fine with merging this and then addressing any other feedback in a subsequent PR

PR #148, opened 2022-11-10
[Justice Counts] Fix data entry form bug when navigating back from publish confirm page

### LINKED ISSUE #152: [Frontend][Bug] Edit metric -> review and publish -> go back to data entry -> changes lost
@lilidworkin commented on [Wed Nov 09 2022](https://github.com/Recidiviz/recidiviz-data/issues/16570)

From Matt at CSG: "I got myself into a weird loop where when I tried to change a metric previously saved to a new number, then review and publish, then back to data entry, the metric reverts to the previously saved value" ([screen capture](https://csgorg-my.sharepoint.com/:v:/g/personal/mherman_csg_org/Efbj1Yo8JeBLiBxZ8qGD_WUBChTvPRN2GVqv50OJni8Wew?e=9DqJWZ))

---

@terryttsai commented on [Wed Nov 09 2022](https://github.com/Recidiviz/recidiviz-data/issues/16570#issuecomment-1309479428)

Thanks for the informative video capture! 

---

@lilidworkin commented on [Thu Nov 10 2022](https://github.com/Recidiviz/recidiviz-data/issues/16570#issuecomment-1310382086)

That was all from Matt actually! :D

terryttsai:
## Description of the change

See video for reproducing the bug here: https://app.zenhub.com/workspaces/justice-counts--622109ea3bb6e0001d2d06ad/issues/recidiviz/recidiviz-data/16570

This bug is caused by `formStore.validatePreviouslySavedInputs` being called when the user is navigating from the Publish Confirm page back to the data entry page, which replaces all user inputs with what was originally fetched from the server when the data entry page was first loaded.

The `DataEntryForm` component gets unmounted and remounted when navigating to and away from the Publish Confirm page, which causes `formStore.validatePreviouslySavedInputs` to get triggered. I believe `formStore.validatePreviouslySavedInputs` is intended to be triggered once on loading the data entry page and not when navigating from Publish Confirm back to the Data Entry Form.

My solution is to move the `formStore.validatePreviouslySavedInputs` trigger out of the `DataEntryForm` component and into the parent `ReportDataEntry` component. The parent component is loaded once when the data entry page is loaded, and calls `formStore.validatePreviouslySavedInputs`. It doesn't get called again until the user navigates away from the data entry section entirely. 

## Type of change

## Related issues

Closes #152

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/ReportDataEntry.tsx
+++ b/publisher/src/components/Reports/ReportDataEntry.tsx
@@ -29,7 +29,6 @@ import styled from "styled-components/macro";
 
 import { trackReportUnpublished } from "../../analytics";
 import { useStore } from "../../stores";
-import FormStore from "../../stores/FormStore";
 import { printReportTitle } from "../../utils";
 import { PageWrapper } from "../Forms";
 import { Loading } from "../Loading";
@@ -62,7 +61,7 @@ const ReportDataEntry = () => {
   >(undefined);
   const [showConfirmation, setShowConfirmation] = useState(false);
   const [showDataEntryHelpPage, setShowDataEntryHelpPage] = useState(false);
-  const { reportStore, userStore } = useStore();
+  const { formStore, reportStore, userStore } = useStore();
   const params = useParams();
   const reportID = Number(params.id);
   const reportOverview = reportStore.reportOverviews[reportID] as Report;
@@ -103,7 +102,7 @@ const ReportDataEntry = () => {
     }
   };
 
-  const checkMetricForErrors = (metricKey: string, formStore: FormStore) => {
+  const checkMetricForErrors = (metricKey: string) => {
     let foundErrors = false;
 
     if (formStore.metricsValues[reportID]?.[metricKey]?.error) {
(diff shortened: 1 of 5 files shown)


publisher/src/components/Reports/ReportDataEntry.tsx
```diff
@@ -155,6 +154,13 @@ const ReportDataEntry = () => {
     []
   );
 
+  useEffect(() => {
+    /** Runs validation of previously saved inputs on load */
+    if (reportMetrics) {
+      formStore.validatePreviouslySavedInputs(reportID);
```
terryttsai, line 160: Alternatively, we could also call this whenever a report is fetched? It could just be part of the process of fetching a report, storing it in the report store, and also storing inputs + validation in the form store?

lilidworkin: That makes a lot of sense actually! I don't have strong feelings though -- if you feel that this solution is robust enough for now @terryttsai , I'm fine to merge this in as-is. Up to you.

lilidworkin (approved):
@terryttsai thank you so much for investigating this and for writing such a clear and well-explained PR description! This solution makes perfect sense to me. Curious what @mxosman thinks about your suggestion of moving this validation logic up to happen right after a report is fetched -- I'll leave it up to you both if you think that's worth doing now. I'm happy with this as-is though :)

publisher/src/components/Reports/ReportDataEntry.tsx
mxosman: No strong feelings either - both make a lot of sense! I'm good with your solution here!

mxosman (approved):
Nice debugging and solution, Terry!!

publisher/src/components/Reports/ReportDataEntry.tsx
terryttsai: I'll go with this for now!

PR #172, opened 2022-11-21
update metric settings badges to reflect custom reporting frequency

nichelle-hall:
## Description of the change

This PR updates the badges used in the metric settings pages so that they reflect the custom reporting frequency of a metric if the custom frequency is not `undefined`.

The default reporting frequency for `Civilians Complaints Sustained` is annual, but a custom reporting frequency has been added such that it is now reported monthly. 

Screen Shot 2022-11-21 at 2 05 31 PM

Screen Shot 2022-11-21 at 2 05 33 PM

## Related issues

Related to #16218

### Development

### Code review

### DIFF
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -56,6 +56,7 @@ class MetricConfigStore {
       label?: string;
       description?: Metric["description"];
       frequency?: Metric["frequency"];
+      customFrequency?: Metric["custom_frequency"];
     };
   };
 
@@ -165,6 +166,7 @@ class MetricConfigStore {
             label?: string;
             description?: Metric["description"];
             frequency?: Metric["frequency"];
+            customFrequency?: Metric["custom_frequency"];
           };
         }[]
       );
@@ -242,6 +244,7 @@ class MetricConfigStore {
               label: metric.label,
               description: metric.description,
               frequency: metric.frequency || "",
+              customFrequency: metric.custom_frequency || "",
             }
           );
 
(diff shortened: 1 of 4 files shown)


publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -358,6 +361,8 @@ class MetricConfigStore {
       this.metrics[systemMetricKey].description = metadata.description;
       this.metrics[systemMetricKey].frequency =
         metadata.frequency as ReportFrequency;
+      this.metrics[systemMetricKey].customFrequency =
```
lilidworkin, line 364: curious what @mxosman and @terryttsai think, but I'm wondering if we should define a property like `frequencyToUse` here, which is `customFrequency` if not null else `frequency`, and then just use this everywhere when rendering. That way, up above we could replace

```
                    {metrics[systemMetricKey]?.customFrequency
                      ? metrics[
                          systemMetricKey
                        ]?.customFrequency?.toLocaleLowerCase()
                      : metrics[systemMetricKey]?.frequency?.toLowerCase()}
```

with 
```
{metrics[systemMetricKey]?.frequencyToUse?.toLowerCase()}
```

lilidworkin (commented):
@nichelle-hall can you do a search to see if there's anywhere else we need to update the rendering of frequency? I just remembered for instance that the badge also shows up on the Data page. Will your change here take care of that, or do we need to make an additional change for that?

publisher/src/stores/MetricConfigStore.tsx
mxosman: I really like that idea - love how clean it makes rendering the frequency! Let me know if TypeScript gives you any trouble, Nichelle!

nichelle-hall: For example, I wonder if we would want to include the default in the data tab design since we may need to distinguish that the reporting frequency has changed if we want to change the x axis of the charts.

mxosman: That's a great call - makes sense why you'd need them separated in the cases you mentioned!

mxosman (approved):
LGTM! Thank you, Nichelle!

lilidworkin:
Sounds good, thanks for explaining your reasoning @nichelle-hall !

PR #173, opened 2022-11-21
Replace all instances of "report" in Publisher

### LINKED ISSUE #165: Replace all instances of "report" in Publisher
**What needs to be done? Why does it need to be done?**
**tl;dr:** The Bureau of Justice Administration insists that agencies are not "reporting" and we are not "collecting" data (ping for details). **As such, we need to remove all instances of the word “report” from Publisher.**  

**Additional context**
All instances of the word Report have been documented [here (in Notion)](https://www.notion.so/recidiviz/Changing-Instances-of-Report-298ccb3cdc3844868f60cfd79bc38967), along with proposed alternatives. General rules for the name changes: 
- Where possible, "report" (used as a verb) should be replaced with "share" or "enter". 
- Reports will be renamed to Records (also taking suggestions for alternative names if this doesn't feel right). 
    - Variable names, functions, databases, etc could still reference "Reports"; will leave up to the engineering team as to what to do about those.

Please look out for any other instances of the word "report" that I might have missed! Ideally this is completed by **Tuesday, November 22** to help unblock conversations with BJA.

mxosman:
## Description of the change

Replaces all user-facing instances of the word "report" (as outlined in this [doc](https://www.notion.so/recidiviz/Changing-Instances-of-Report-298ccb3cdc3844868f60cfd79bc38967)) with relevant replacement constants. There were other components outside of the ones listed in the doc that needed updating - I went through ~60 files that had the word "report" in it, and found some hidden hotspots like toasts/error messages, etc. 

If you have a moment, please feel free to test the link below and let me know if there are areas I may have missed!

Demo: https://mahmoudtest---justice-counts-web-qqec6jbn6a-uc.a.run.app/

## Related issues

Closes #165 

### Development

### Code review


publisher/src/components/Menu/Menu.tsx
```diff
@@ -32,8 +33,8 @@ import {
 } from ".";
 
 enum MenuItems {
-  Reports = "REPORTS",
-  CreateReport = "CREATE REPORT",
+  Reports = "RECORDS",
+  CreateReport = "CREATE RECORD",
```
mxosman, line 37: TypeScript is strict about enums not having computed values - only strings.

publisher/src/components/Global/constants.ts
```diff
@@ -0,0 +1,36 @@
...
+export const REPORTS_UPPERCASE = "RECORDS";
+
+export const REPORT_CAPITALIZED = "Record";
+export const REPORTS_CAPITALIZED = "Records";
+export const REPORTING_CAPITALIZED = "Data Sharing";
+export const REPORTED_CAPITALIZED = "Shared";
+export const REPORT_PERIOD_CAPITALIZED = "Time Period";
+
+export const REPORT_LOWERCASE = "record";
+export const REPORT_LOWERCASE2 = "share";
```
terryttsai, line 28: What's the difference between `REPORT_LOWERCASE2`, `REPORT_VERB_LOWERCASE`, `GATHER_LOWERCASE`? Since they are all defined as "share"?

terryttsai (commented):
Awesome! Thanks for going through the whole app!

I'm wondering if it would be worth renaming all our constants / file names to Record instead of Report, so that we don't refer to Report internally but have to remember to say Record externally, or new developers have to learn that Report = Record? 

Or maybe it's not worth such a big constant renaming push especially if the name changes again?

lilidworkin:
> I'm wondering if it would be worth renaming all our constants / file names to Record instead of Report, so that we don't refer to Report internally but have to remember to say Record externally, or new developers have to learn that Report = Record?

@terryttsai this is a good callout -- @mxosman and I discussed this and I actually suggested keeping the code as Report. It does seem likely/possible that it will change again, and I think it's okay for our code and our internal mental model to stay with "Report". I don't feel strongly and I see the argument both ways, but given how much easier it is to keep as-is, that would be my suggestion.

publisher/src/components/Global/constants.ts
lilidworkin: Yea I agree that REPORT_LOWERCASE and REPORT_VERB_LOWERCASE can probably be consolidated? And REPORT_LOWERCASE3 should maybe be REPORT_VERB_LOWERCASE2, and consolidated with GATHER_LOWERCASE?

lilidworkin (commented):
> I went through ~60 files that had the word "report" in it, and found some hidden hotspots like toasts/error messages, etc.

@mxosman that's amazing -- thank you SO MUCH for doing such impressive due diligence!

mxosman:
> > I'm wondering if it would be worth renaming all our constants / file names to Record instead of Report, so that we don't refer to Report internally but have to remember to say Record externally, or new developers have to learn that Report = Record?
> 
> @terryttsai this is a good callout -- @mxosman and I discussed this and I actually suggested keeping the code as Report. It does seem likely/possible that it will change again, and I think it's okay for our code and our internal mental model to stay with "Report". I don't feel strongly and I see the argument both ways, but given how much easier it is to keep as-is, that would be my suggestion.

Yeah - this was also more/less the quickest solution to unblock Humphrey. I think there is somewhere around 1k instances of "report" to replace - probably best to feel confident that it won't change before fully switching over.

publisher/src/components/Global/constants.ts
mxosman: Great call! Was super tunnel-visioned here - will consolidate this.

terryttsai:
Keeping the internal mental model as Report sounds good to me, thanks for explaining the thought process!

publisher/src/components/Global/constants.ts
mxosman: Cleaned up here: [d418ae6](https://github.com/Recidiviz/justice-counts/pull/173/commits/d418ae6de6d88c2b8342ef36483b8872e18e37ee)

mxosman:
> Keeping the internal mental model as Report sounds good to me, thanks for explaining the thought process!

I do think that eventually (once the naming is stable) it would be worth it to rename everything for all the reasons you mentioned!

terryttsai (approved):
Looks awesome!

PR #174, opened 2022-11-23
Race & Ethnicities (Playtesting Followup): Debounce Saving & Send All Dimensions

mxosman:
## Description of the change

Debounce saving and send all 24 dimensions instead of the deltas. Also, optimize saving by using `input`s for the buttons and utilizing `onChange`.

Demo: https://playtesting---justice-counts-web-qqec6jbn6a-uc.a.run.app/

## Related issues

Contributes to #180 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Forms/BinaryRadioButton.tsx
+++ b/publisher/src/components/Forms/BinaryRadioButton.tsx
@@ -20,7 +20,7 @@ import {
   typography,
 } from "@justice-counts/common/components/GlobalStyles";
 import React, { InputHTMLAttributes } from "react";
-import styled from "styled-components/macro";
+import styled, { css } from "styled-components/macro";
 
 export const BinaryRadioGroupContainer = styled.div`
   display: flex;
@@ -43,7 +43,7 @@ export const BinaryRadioGroupQuestion = styled.div`
   color: ${palette.solid.darkgrey};
 `;
 
-export const RadioButtonWrapper = styled.div`
+export const RadioButtonWrapper = styled.div<{ lastOptionBlue?: boolean }>`
   display: flex;
   flex: 1 1 0;
   margin: 15px 0 0 0;
@@ -51,6 +51,22 @@ export const RadioButtonWrapper = styled.div`
   &:first-child {
     margin: 15px 10px 0 0;
   }
+
+  ${({ lastOptionBlue }) =>
+    lastOptionBlue &&
+    css`
+      margin: 0;
+
+      &:first-child {
+        margin: 0px 10px 0 0;
+      }
+
+      &:first-child input:checked + label {
+        background-color: ${palette.highlight.grey9};
+        border-color: unset;
+        color: ${palette.solid.white};
+      }
+    `}
 `;
 
 export const RadioButtonElement = styled.input<{
@@ -83,6 +99,7 @@ export const RadioButtonElement = styled.input<{
 
 export const RadioButtonLabel = styled.label<{
   disabled?: boolean;
+  buttonSize?: string;
 }>`
   ${typography.sizeCSS.medium}
   width: 100%;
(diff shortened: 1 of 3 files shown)


lilidworkin (approved):
Looks great! Thanks for making this change. It actually seems to simplify the code a bit too!

> Also, optimize saving by using inputs for the buttons and utilizing onChange.

Just out of curiosity, would you mind expanding on this a little?

mxosman:
> Looks great! Thanks for making this change. It actually seems to simplify the code a bit too!

Thank you for the suggestion - it definitely simplified a lot of the code! 

> 
> > Also, optimize saving by using inputs for the buttons and utilizing onChange.
> 
> Just out of curiosity, would you mind expanding on this a little?

Absolutely! Before I just used a `div` to act as a mini button and used an `onClick` handler - every time you clicked on an option that was already selected, it would fire off the onClick function. Here, I am extending the current binary radio input  component and using the input's `onChange` event handler so nothing fires off when the user clicks an already selected option. It's better practice in general to be using `input`s for buttons. Since our buttons are continuing to evolve, I'll make a task to go through at some point and centralize/clean them up.

publisher/src/components/Forms/BinaryRadioButton.tsx
```diff
@@ -127,16 +156,22 @@ export const BinaryRadioButton: React.FC<RadioButtonProps> = ({
   context,
   metricKey,
   disabled,
+  buttonSize,
+  lastOptionBlue,
   ...props
 }): JSX.Element => {
   return (
-    <RadioButtonWrapper>
+    <RadioButtonWrapper lastOptionBlue={lastOptionBlue}>
```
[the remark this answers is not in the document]
mxosman: In the designs, it looks like the mini buttons are all grey when selected, except for the last button which is blue when selected. 
Screenshot 2022-11-30 at 10 43 51 AM

Since I'm extending our BinaryRadioButton component, I needed to add this to have it deviate from it's current behavior throughout the app (all selected options are blue):

Screenshot 2022-11-30 at 10 43 04 AM
Screenshot 2022-11-30 at 10 42 58 AM

Let me know if you think I should have a clearer name/add a comment in the BinaryRadioButton component describing this.

lilidworkin:
Nice, that makes perfect sense @mxosman ! Thanks for explaining!

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -819,17 +805,25 @@ class MetricConfigStore {
...
+      (Object.values(raceEthnicitiesDimensions) as UpdatedDimension[]);
+
+    /** Return an object w/ all dimensions in the desired backend data structure for saving purposes */
     return {
       key: metricKey,
       disaggregations: [
         {
           key: RACE_ETHNICITY_DISAGGREGATION_KEY,
-          dimensions: updatedDimensions,
+          dimensions,
```
terryttsai, line 826: What if dimensions is undefined?

terryttsai: If in line 817 raceEthnicitiesDimensions is falsey

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -770,21 +756,22 @@ class MetricConfigStore {
...
```
terryttsai, line 774: Same here, what if dimensions is undefined?

publisher/src/stores/MetricConfigStore.tsx
mxosman: Always a great question - part of the reason I added the `raceEthnicitiesDimensions` and `dimensions` below the updater function is to ensure that `dimensions` will always have a value. In fact, `dimensions` doesn't even really need the `raceEthnicitiesDimensions &&` check.

If somehow it's `undefined` and we get to this point, it'll just return the object without the `dimensions` property - but I can't figure out how it could end up as `undefined` by the time we get here in either function.

PR #182, opened 2022-11-28
[Fix] Settings: Metric Configuration Link

mxosman:
## Description of the change

Terry pointed out an issue with the Metric Configuration link in the Settings side menu where it shows up nothing when clicked (after you've selected a metric). After debugging, it looks like it navigates back to `/settings/metric-config` without the additional params. The solution is to update the Settings menu navigation to include the system params for the Metric Configuration link.

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/Settings/SettingsMenu.tsx
+++ b/publisher/src/components/Settings/SettingsMenu.tsx
@@ -54,6 +54,9 @@ export const SettingsMenu: React.FC = observer(() => {
           <MenuItem
             selected={location.pathname === path}
             onClick={() => {
+              if (displayName === "Metric Configuration") {
+                return navigate(`${path}?system=${systemSearchParam}`);
+              }
               navigate(path);
             }}
           >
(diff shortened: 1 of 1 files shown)


publisher/src/components/Settings/SettingsMenu.tsx
```diff
@@ -54,6 +54,9 @@ export const SettingsMenu: React.FC = observer(() => {
           <MenuItem
             selected={location.pathname === path}
             onClick={() => {
+              if (displayName === "Metric Configuration") {
```
terryttsai, line 57: Would this be prone to breaking if the displayName changes?

terryttsai: Another possible solution is perhaps if we navigate to `/settings/metric-config` without params, then the component can find the default system and set the param?

mxosman:
> This looks like a good fix for the problem at hand, though here are some other cases to think about:
> 
> What if a user shares a link like `/settings/metric-config` without the system? If someone clicks on that link, will they see the blank screen? Should we re-route them to the right system in that case?
> 
> Or what if a user shares a link like `/settings/metric-config?system=SOME_SYSTEM` and the person that clicks on the link doesn't have that system? Should we also re-route them to the right system in that case?

Thanks, Terry - it's currently handling both cases - Ivan already solved for that. Let me know when you get a chance to run it!

terryttsai (approved):
Fix works for me! LGTM!

PR #186, opened 2022-11-29
[Agency Dashboard] DataVizStore implementation

terryttsai:
## Description of the change

Creation of DataVizStore to share data viz state across multiple different components, namely to use with the existing DatapointsView component and the MetricInsights component which I'm rendering on the left panel of the agency dashboard.

If Publisher and Agency Dashboard's charts continue to diverge, we could break up the entire DatapointsView component (which renders the chart, metric title, legend, insights, dropdowns, select metrics button) to its Publisher version and Agency Dashboard version. Right now it's one big component that conditionally renders some of the components depending on how its configured. But for now I'm going to leave this alone.

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/common/components/DataViz/DatapointsView.tsx
+++ b/common/components/DataViz/DatapointsView.tsx
@@ -48,7 +48,7 @@ import {
   DataVizAggregateName,
   DataVizTimeRangeDisplayName,
   DataVizTimeRangesMap,
-  DataVizViewSetting,
+  DataVizCountOrPercentageView,
   DimensionNamesByDisaggregation,
   NoDisaggregationOption,
   ReportFrequency,
@@ -92,11 +92,11 @@ export const DatapointsView: React.FC<{
   datapointsGroupedByAggregateAndDisaggregations?: DatapointsGroupedByAggregateAndDisaggregations;
   dimensionNamesByDisaggregation?: DimensionNamesByDisaggregation;
   timeRange: DataVizTimeRangeDisplayName;
-  disaggregation: string;
-  viewSetting: DataVizViewSetting;
+  disaggregationName: string;
+  countOrPercentageView: DataVizCountOrPercentageView;
   setTimeRange: (timeRange: DataVizTimeRangeDisplayName) => void;
-  setDisaggregation: (disaggregation: string) => void;
-  setViewSetting: (viewSetting: DataVizViewSetting) => void;
+  setDisaggregationName: (disaggregation: string) => void;
+  setCountOrPercentageView: (viewSetting: DataVizCountOrPercentageView) => void;
   metricName?: string;
   metricFrequency?: ReportFrequency;
   metricNames?: string[];
@@ -107,11 +107,11 @@ export const DatapointsView: React.FC<{
   datapointsGroupedByAggregateAndDisaggregations,
   dimensionNamesByDisaggregation,
   timeRange,
-  disaggregation,
-  viewSetting,
+  disaggregationName,
+  countOrPercentageView,
   setTimeRange,
-  setDisaggregation,
-  setViewSetting,
+  setDisaggregationName,
+  setCountOrPercentageView,
   metricName,
   metricFrequency,
   metricNames,
@@ -123,10 +123,10 @@ export const DatapointsView: React.FC<{
     React.useState<boolean>(false);
 
   const selectedData =
-    (disaggregation !== NoDisaggregationOption &&
+    (disaggregationName !== NoDisaggregationOption &&
       Object.values(
         datapointsGroupedByAggregateAndDisaggregations?.disaggregations[
-          disaggregation
+          disaggregationName
         ] || {}
       )) ||
     datapointsGroupedByAggregateAndDisaggregations?.aggregate ||
@@ -136,8 +136,8 @@ export const DatapointsView: React.FC<{
   const disaggregationOptions = [...disaggregations];
   disaggregationOptions.unshift(noDisaggregationOption);
   const dimensionNames =
-    disaggregation !== noDisaggregationOption
-      ? (dimensionNamesByDisaggregation?.[disaggregation] || [])
+    disaggregationName !== noDisaggregationOption
+      ? (dimensionNamesByDisaggregation?.[disaggregationName] || [])
           .slice() // Must use slice() before sorting a MobX observableArray
           .sort(sortDatapointDimensions)
       : [DataVizAggregateName];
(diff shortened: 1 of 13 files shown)


common/components/DataViz/DatapointsView.tsx
```diff
@@ -209,49 +211,38 @@ export const DatapointsView: React.FC<{
...
-      datapointsGroupedByAggregateAndDisaggregations?.aggregate || [],
-      selectedTimeRangeValue
-    )
-  );
-  const percentChange = getPercentChangeOverTime(dataSelectedInTimeRange);
-  const avgValue = getAverageTotalValue(dataSelectedInTimeRange, isAnnual);
-  const mostRecentValue = getLatestDateFormatted(
-    dataSelectedInTimeRange,
-    isAnnual
+  const filteredAggregateData = transformDataForMetricInsights(
```
lilidworkin, line 241: Love this new helper function!

common/components/DataViz/utils.ts
```diff
@@ -255,30 +255,39 @@ export const fillTimeGapsBetweenDatapoints = (
   return dataWithGapDatapoints;
 };
 
-export const transformData = (
-  d: Datapoint[],
+// transforms data into the right display format for the data viz chart
+export const transformDataForBarChart = (
```
lilidworkin, line 259: Thanks for renaming this to differentiate with the new helper!

lilidworkin (commented):
This is really cool!

> If Publisher and Agency Dashboard's charts continue to diverge, we could break up the entire DatapointsView component (which renders the chart, metric title, legend, insights, dropdowns, select metrics button) to its Publisher version and Agency Dashboard version. 

I agree that we should keep these together for now! We should push back against too much divergence here. Anytime the designs or functionality spec differs significantly between the two, we should call it out and discuss with Humphrey and Juan and see if there's a way to keep things more similar. (If you feel like things are already diverging a lot, it might be helpful to start keeping track of a list of things that are different?)

This is probably a very naive question, but I'm curious how this new store works when you switch metrics. Does the selected time range, disaggregation, etc. reset? Or does it persist?

common/components/DataViz/utils.ts
```diff
@@ -255,30 +255,39 @@ export const fillTimeGapsBetweenDatapoints = (
...
+  datapoints: Datapoint[],
+  monthsAgo: DataVizTimeRange
+) => {
+  if (datapoints.length === 0) {
+    return datapoints;
+  }
+  return filterNullDatapoints(filterByTimeRange(datapoints, monthsAgo));
+};
+
 // get insights from data
```
lilidworkin, line 291: super nit, should this comment be moved above `transformDataForMetricInsights` now?

common/components/DataViz/utils.ts
```diff
@@ -255,30 +255,39 @@ export const fillTimeGapsBetweenDatapoints = (
...
 };
 
+export const transformDataForMetricInsights = (
+  datapoints: Datapoint[],
+  monthsAgo: DataVizTimeRange
+) => {
+  if (datapoints.length === 0) {
+    return datapoints;
+  }
+  return filterNullDatapoints(filterByTimeRange(datapoints, monthsAgo));
```
lilidworkin, line 288: This is probably a bit of a nit, but both of these helper functions (transform for bar and transform for metric insights) do the same null filtering and time range filtering. Is that redundant work? Is there a way to just do that once and then share it?

common/types.ts
```diff
@@ -278,15 +278,24 @@ export interface RawDatapointsByMetric {
...
+} = {
   All: 0,
   "6 Months Ago": 6,
   "1 Year Ago": 12,
   "5 Years Ago": 60,
   "10 Years Ago": 120,
 };
 
-export type DatapointsViewSetting = "Count" | "Percentage";
+export type DataVizViewSetting = "Count" | "Percentage";
```
lilidworkin, line 298: nit, but this name confused me, I wonder if we can just explicitly call this DataVizCountOrPercentageView or something like that to be explicit?

common/stores/DataVizStore.ts
```diff
@@ -0,0 +1,58 @@
...
+  setDisaggregation = (disaggregation: string) => {
+    runInAction(() => {
+      this.disaggregation = disaggregation;
+    });
+  };
+
+  setViewSetting = (viewSetting: DataVizViewSetting) => {
+    runInAction(() => {
+      this.viewSetting = viewSetting;
+    });
```
mxosman, line 54: Do these need to be wrapped in a `runInAction` if the function is not async?

common/components/DataViz/DatapointsView.tsx
```diff
@@ -92,8 +89,14 @@ const SelectMetricButtonDropdown: React.FC<{
 );
 
 export const DatapointsView: React.FC<{
-  datapointsGroupedByAggregateAndDisaggregations: DatapointsGroupedByAggregateAndDisaggregations;
-  dimensionNamesByDisaggregation: DimensionNamesByDisaggregation;
+  datapointsGroupedByAggregateAndDisaggregations?: DatapointsGroupedByAggregateAndDisaggregations;
+  dimensionNamesByDisaggregation?: DimensionNamesByDisaggregation;
+  timeRange: DataVizTimeRangeDisplayName;
+  disaggregation: string;
```
mxosman, line 95: small nit: not sure if the `disaggregation` string is the disaggregation key or name - but might be worth noting in the variable incase it might be confused with a disaggregation object.

common/stores/DataVizStore.ts
terryttsai: Good call, removing

common/types.ts
terryttsai: Renamed!

common/components/DataViz/DatapointsView.tsx
terryttsai: Renamed!

terryttsai:
The key parts that are different are the addition of the Select Metrics button and showing the Metric Insights in a separate component in the Agency Dashboard, and rendering the metric title in Publisher (which I think should probably be separated out from the DatapointsView component)

terryttsai:
It was difficult to figure out how to place the metric insights with the datapoints view so trying out putting it in the left panel made sense at the time

common/components/DataViz/utils.ts
terryttsai: The bar chart has the extra step of transforming for percent view but i'll share logic between them so thanks for pointing that out!

common/components/DataViz/utils.ts
terryttsai: Renamed comment to be more clear that these are calculations from already transformed data

mxosman (approved):
Looks good to me, Terry! Awesome to see how you're breaking this out and thinking about it.

lilidworkin:
@terryttsai naive question but I'm curious how this new store works when you switch metrics. Does the selected time range, disaggregation, etc. reset? Or does it persist?

terryttsai:
Oh sorry I missed answering your question earlier—the behavior stays the same as before, since we're relying on react useEffects on the data viz component to reset these fields

lilidworkin:
Ah thanks, makes sense!!

PR #199, opened 2022-12-03
Custom Report Frequency

### LINKED ISSUE #204: [Frontend] Custom Reporting Frequency

mxosman:
## Description of the change

Allows user to specify and update reporting frequency for each metric.

Quick Demo: 

https://user-images.githubusercontent.com/59492998/205953863-40c2098b-54df-4cd0-8a80-8e9f406c29cb.mov

## Related issues

Closes #204 

### Development

### Code review


mxosman:
There's a weird flakey bug that only appears to be happening on my end (curious if anyone else can repro) where some of the metrics aren't saving properly:


https://user-images.githubusercontent.com/59492998/205954548-f6b70715-4325-4cd0-bfb9-ec54b45e0a60.mov

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -324,7 +326,7 @@ class MetricConfigStore {
     system: AgencySystems,
     metricKey: string,
     enabledStatus: boolean,
-    metadata?: { [key: string]: string }
+    metadata?: { [key: string]: string | number | null }
```
terryttsai, line 329: Should the keys of the metadata be typed? like 
```
label: string;
description: string;
...
```

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -356,6 +362,50 @@ class MetricConfigStore {
...
 
+  updateMetricReportFrequency = (
+    system: AgencySystems,
+    metricKey: string,
+    update: {
+      frequency?: ReportFrequency;
+      customFrequency?: ReportFrequency | string;
+      startingMonth?: number | null;
+    },
+    reenableMetric?: boolean
```
terryttsai, line 373: Maybe we can just have `updateMetricReportFrequency` always set enabled to true? is there a case of updating the metric frequency but we want the metric to stay disabled?

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -356,6 +362,50 @@ class MetricConfigStore {
...
+      this.metrics[systemMetricKey].frequency = update.frequency;
+    }
+    if (update.customFrequency !== undefined) {
+      this.metrics[systemMetricKey].customFrequency = update.customFrequency;
+    }
+    if (update.startingMonth !== undefined) {
+      this.metrics[systemMetricKey].startingMonth = update.startingMonth;
+    }
+    if (reenableMetric) {
+      this.updateMetricEnabledStatus(system, metricKey, true);
```
terryttsai, line 396: It's a little confusing to me that `updateMetricEnabledStatus` is also used to initialize the metric config's info values, I'm thinking there should be a `updateMetric` that takes in any keys, then `updateMetricEnabledStatus` uses that function to specifically set the `enabled` key. 

Like
```
updateMetricEnabledStatus = (
    system: AgencySystems,
    metricKey: string,
    enabledStatus: boolean,
  ): MetricSettings => {
  return this.updateMetric(system, metricKey, { enabled: enabledStatus });
}
```

Then `updateMetric` can be used when initializing all the other values

terryttsai (approved):

terryttsai:
Sorry, I accidentally approved when I meant to leave comment :P

terryttsai (changes requested):
Sorry, I meant to comment rather than approve!

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -52,7 +52,8 @@ class MetricConfigStore {
       label?: string;
       description?: Metric["description"];
       frequency?: Metric["frequency"];
-      customFrequency?: Metric["custom_frequency"];
+      customFrequency?: Metric["custom_frequency"] | string;
```
terryttsai, line 55: is this because these fields are optional? do they have to be optional?

publisher/src/stores/MetricConfigStore.tsx
terryttsai: Or we can just have the `updateMetric` function, we don't need `updateMetricEnabledStatus`

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -73,24 +78,32 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
...
     const activeDimensionKeys =
       activeDisaggregationKey &&
       dimensions[systemMetricKey]?.[activeDisaggregationKey]
         ? Object.keys(dimensions[systemMetricKey][activeDisaggregationKey])
         : [];
 
-    const metricDisplayName = metrics[systemMetricKey]?.label;
-
     const metricEnabled = Boolean(metrics[systemMetricKey]?.enabled);
+    const frequency = metrics[systemMetricKey]?.frequency;
```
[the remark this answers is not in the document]
mxosman: Yeah, I ran into a crash because `metrics[systemMetricKey]` - total guess - but I think now that we have everything routed through URLs, there maybe a brief moment where metrics[systemMetricKey] is `undefined` before it goes through initialization.

publisher/src/stores/MetricConfigStore.tsx
mxosman: When there is no `customFrequency` the backend sends an empty string.

publisher/src/stores/MetricConfigStore.tsx
mxosman: I don't think so? I think we kept this purposefully typed in a flexible way so we can pass in any new property that needs rendering - this is just the first time and only in this function's `metadata` that we will receive `number` and `null`.

publisher/src/stores/MetricConfigStore.tsx
mxosman: Great idea! I wasn't sure if always sending `enabled` even when the status is already enabled (say you've already selected "Monthly" - and you go to select "Annual") would be suboptimal.

publisher/src/stores/MetricConfigStore.tsx
mxosman: I'm not sure I 100% agree. Originally when we set up these functions, we realized a need to initialize values in the store that will be rendered - and the simplest solution we agreed on was passing in an optional `metadata` (which is never run if not passed in). All of the other pieces aside from the metadata logic are crucial - not sure we can just get away with an `updateMetric` that doesn't have the same core pieces we thought out beforehand.

Please let me know if I'm misunderstanding!

publisher/src/stores/MetricConfigStore.tsx
lilidworkin: Would it help if the backend sent None instead? That feels more correct to me?

publisher/src/components/MetricConfiguration/Configuration.tsx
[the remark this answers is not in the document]
lilidworkin: I'm also not sure about this, we just kept it the same on the backend to minimize the number of changes, but it is true that it could cause confusion. I don't feel strongly.

publisher/src/stores/MetricConfigStore.tsx
lilidworkin: I think that's fine -- hopefully the backend is smart enough not to do extra work here, and if not, that's really the backend's problem!

lilidworkin:
@mxosman no rush on this but let me / @nichelle-hall  know if you're able to deploy to playtesting and repro the bug there! And if so, tell us which agency/metrics to use to repro and we'll look into it!

publisher/src/components/MetricConfiguration/Configuration.tsx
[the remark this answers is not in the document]
mxosman: I'm personally leaning towards minimizing the # of changes - but don't feel strongly either.

publisher/src/stores/MetricConfigStore.tsx
mxosman: We can do `None`! I agree with it feeling more correct than an empty string.

mxosman:
> @mxosman no rush on this but let me / @nichelle-hall know if you're able to deploy to playtesting and repro the bug there! And if so, tell us which agency/metrics to use to repro and we'll look into it!

Will do!

publisher/src/stores/MetricConfigStore.tsx
lilidworkin: Great! @nichelle-hall can you make that change?

mxosman:
> @mxosman no rush on this but let me / @nichelle-hall know if you're able to deploy to playtesting and repro the bug there! And if so, tell us which agency/metrics to use to repro and we'll look into it!

OK - great news - I've tried a few agencies and all their metrics and haven't been able to repro! You will see in some areas when you first load a metric - it might by in a funky state, but you can disable and reenable the metric and things work just fine. Would love a sanity check if folks have time/bandwidth!

Here's the playtesting link: https://playtesting---justice-counts-web-qqec6jbn6a-uc.a.run.app

publisher/src/stores/MetricConfigStore.tsx
mxosman: Only if it doesn't complicate things in the backend - both empty string and null will be falsy - so it won't make a huge difference!

lilidworkin:
Awesome, thanks @mxosman ! I also checked it out briefly and wasn't able to repro, but I'll spend a little more time playing around with it later.

publisher/src/stores/MetricConfigStore.tsx
terryttsai: Maybe this proposal makes more sense: 

Instead of having `updateMetricEnabledStatus` handle both changing the metric enabled AND initializing metric metadata, we create a separate `initializeMetric` function like so:

```
  initializeMetric = (
    system: AgencySystems,
    metricKey: string,
    metadata: { [key: string]: string | number | null }
  ) => {
    const systemMetricKey = MetricConfigStore.getSystemMetricKey(
      system,
      metricKey
    );

    /** Initialize nested object for quick lookup and update and reduce re-renders */
    if (!this.metrics[systemMetricKey]) {
      this.metrics[systemMetricKey] = {};
    }

    /** Add metadata required for rendering */
    this.metrics[systemMetricKey].label = metadata.label as string;
    this.metrics[systemMetricKey].description = metadata.description as string;
    this.metrics[systemMetricKey].defaultFrequency =
      metadata.frequency as ReportFrequency;
    this.metrics[systemMetricKey].customFrequency =
      metadata.customFrequency as ReportFrequency;
    this.metrics[systemMetricKey].startingMonth = metadata.startingMonth as
      | number
      | null;
  };
```

then `updateMetricEnabledStatus` can be simplified to just only updating the metric enabled state?

Originally I was thinking of some base-level `updateMetric` function that is shared between these different update functions but how does this proposal sound?

publisher/src/stores/MetricConfigStore.tsx
terryttsai: So I see a lot of casting as string and null going on here, and I also see the `label, description, frequency, customFrequency, startingMonth` params being redefined often. I think in a follow-up PR this could be cleaned up, something like: https://github.com/Recidiviz/justice-counts/commit/e7a5ac75147af607b0ee956ab1064bc4da6c5abd

By typing metadata into these specific keys, then I no longer have to cast all the metadata as string or ReportFrequency for each key. I hadn't taken a good look to understand the system in place so let me know if this is totally off the mark with what you're going for!

publisher/src/stores/MetricConfigStore.tsx
terryttsai: Hey so perhaps in a follow-up PR I'm wondering if we can change the type around such that we won't need to cast as string here: https://github.com/Recidiviz/justice-counts/commit/e7a5ac75147af607b0ee956ab1064bc4da6c5abd

terryttsai (approved):
This system looks good to me, thanks for reasoning out and dealing w all the complexity here! I answered your question to my feedback and left a quick proposal for possible further type improvements in the future, though I hadn't looked at this long enough to know if what I'm proposing makes sense for this feature. Let's get this merged in!

terryttsai:
Oh the commit I put up https://github.com/Recidiviz/justice-counts/commit/e7a5ac75147af607b0ee956ab1064bc4da6c5abd also includes the `initializeMetric` I was alluding to [here](https://github.com/Recidiviz/justice-counts/pull/199#discussion_r1042787730), the next step down that route is to use that method for initialization, and remove `metadata` from `updateMetricEnabledStatus`

publisher/src/components/MetricConfiguration/Configuration.tsx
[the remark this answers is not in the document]
mxosman: I wonder if we'll run into `systemMetricKey` being undefined with the new changes in main - since entering an invalid/empty value for metric/system in params gets autocorrected.

publisher/src/stores/MetricConfigStore.tsx
terryttsai: Looks good!

publisher/src/stores/MetricConfigStore.tsx
terryttsai: Thanks for making the task for it! Glad it makes sense after I had some thought to write it out!

PR #261, opened 2022-12-27
Guidance: mobX Data Store

### LINKED ISSUE #249: [Frontend][Publisher] Guidance: mobX Data Store

mxosman:
## Description of the change

Adds basic mobX data store for the Guidance component and sets up basic store functionality (e.g. computing `hasCompletedOnboarding`, `currentTopicID` and updating a topic's status.

This does not contain all of the store members needed for this feature - just a few to start and I'll be adding more as we go along.

## Related issues

Closes #249

### Development

### Code review


publisher/src/components/Guidance/types.ts
```diff
@@ -0,0 +1,82 @@
...
+    pathToTask?: string;
+  };
+};
+
+export type OnboardingTopicsStatus = {
+  topicID: TopicID;
+  topicCompleted: boolean;
+};
+
+export const onboardingTopicsMetadata: OnboardingTopicsMetadata = {
```
terryttsai, line 38: Is this mock data?

publisher/src/stores/GuidanceStore.ts
```diff
@@ -0,0 +1,75 @@
...
+    makeAutoObservable(this);
+
+    this.api = api;
+    this.userStore = userStore;
+    this.onboardingTopicsMetadata = onboardingTopicsMetadata;
+    this.onboardingTopicsStatus = mockTopicsStatus;
+  }
+
+  get hasCompletedOnboarding() {
+    if (this.onboardingTopicsStatus.length === 0) return;
```
terryttsai, line 49: What does it mean to return undefined? Can it just return false?

publisher/src/components/Guidance/types.ts
mxosman: The topics metadata is not mock data - it's more of a template of things to render for each step. I figured it would give us one central place to make updates to the copy/links/etc.

publisher/src/stores/GuidanceStore.ts
mxosman: Ah - nice catch! I think so! Having the `onboardingTopicsStatus` be [] empty is a weird state and should probably never happen - I figured we'd just short circuit - but returning `false` makes sense!

publisher/src/stores/GuidanceStore.ts
```diff
@@ -0,0 +1,75 @@
...
+    return this.onboardingTopicsStatus[topicIndex].topicID;
+  }
+
+  updateTopicStatus = (topicID: TopicID, status: boolean) => {
+    this.onboardingTopicsStatus = this.onboardingTopicsStatus.map((topic) => {
+      if (topic.topicID !== topicID) return topic;
+      return { ...topic, topicCompleted: status };
+    });
+
+    return this.onboardingTopicsStatus;
```
terryttsai, line 71: Does this need to return anything?

mxosman: I kept it there incase we decide to show overall guidance progress - thought it would be nice to get a fresh updated object. I can remove it for now and add it back in if we ever decide to go that route.

mxosman: [9f5fcf6](https://github.com/Recidiviz/justice-counts/pull/261/commits/9f5fcf606e16ec89f7f2e98889600313a71b3cbb)

publisher/src/stores/GuidanceStore.ts
```diff
@@ -0,0 +1,73 @@
...
+  get currentTopicID() {
+    if (this.onboardingTopicsStatus.length === 0) return;
+    const topicIndex = this.onboardingTopicsStatus.findIndex(
+      (topic) => !topic.topicCompleted
+    );
+    if (topicIndex < 0) return;
+    return this.onboardingTopicsStatus[topicIndex].topicID;
+  }
+
+  updateTopicStatus = (topicID: TopicID, status: boolean) => {
```
lilidworkin, line 65: What if mockTopicStatus was a map of topic ID to status, then we could update the map directly, this.onboardingTopicsStatus[topicID] = status, making this a O(1) operation instead of O(n)?

publisher/src/router/Router.tsx
```diff
@@ -34,15 +34,14 @@ import { useStore } from "../stores";
...
+  const { userStore, guidanceStore } = useStore();
 
   // using this to indicate whether user has agency id of which is provided in the url
   // if not then show placeholder (InvalidAgency page) from which you can go to homepage
   // e.g. reports page with initial available user agency
   // or maybe display some text since header is available and user can pick available agency
   const isAgencyIdInUserAgencies = userStore.getAgency(agencyId);
-  // TODO(#249): Move boolean to mobX data store
-  const hasCompletedOnboarding = true;
+  const { hasCompletedOnboarding } = guidanceStore;
```
lilidworkin, line 44: Was this previously based on FRE? How does guidance interact with FRE?

lilidworkin (approved):
Looks great! Left a comment but not at all blocking -- let's maybe go ahead and just merge some of these earlier PRs in, we can always address the comments in later PRs in the stack

terryttsai (approved):

publisher/src/stores/GuidanceStore.ts
mxosman: Great idea! I thought about that initially and opted to do it this way because there were only going to be ~4 topics and I couldn't imagine us adding a ton of new topics, but I can absolutely rework this to be a mapped object!

publisher/src/router/Router.tsx
mxosman: Ah, this was fleshing out what I had in the previous Routing PR. In the previous routing PR, I put a temporary `hasCompletedOnboarding` boolean until I implemented a data store to sort out the initial routing.

But, this is a great question about the original FRE. I was under the assumption that this new Guidance flow will replace that - but now, I'm not 100% sure. The state for the original FRE lives in Auth0 metadata. Do you know if it be replaced by this flow?

publisher/src/stores/GuidanceStore.ts
mxosman: Added to Task #329

lilidworkin: Actually yea you raise a great point -- it's not really O(n) since we only have 4 topics and that list is unlikely to grow! It's O(4) which is O(1) :) I was being overeager with my optimization again, lol. I still think a map might be cleaner, but it won't really make a performance difference :D

publisher/src/router/Router.tsx
lilidworkin: Ah makes sense! Yea I'm not sure either -- maybe ask Humphrey in #justice-counts? I could see them both existing at the same time, it feels like they could maybe be compatible or at least not mutually exclusive?

publisher/src/stores/GuidanceStore.ts
mxosman: I do think it's a really good habit for me to prefer the mapped object like you suggested! I always appreciate this advice, Lili! It definitely doesn't hurt to always optimize this way!

publisher/src/router/Router.tsx
mxosman: Will do!!

PR #266, opened 2023-01-03
Guidance: Add Data

### LINKED ISSUE #258: [Frontend][Publisher] Guidance: Add Data

mxosman:
## Description of the change

Implements the Add Data step in the guidance flow which allows a user to choose between uploading a spreadsheet or manually filling out a record. This step unlocks the `Records` page as well as the `Upload Data` button (in the header). 

NOTE: I've commented out the line that updates the topic status based on whether or not the user has at least one report. Since we all have atleast one report - I figured I'll disable it for now so you can experience this step (otherwise, it'll skip this step because you've already added data and go to the Publish Data step).

Demo:

https://user-images.githubusercontent.com/59492998/210604999-0bdea03f-1ef5-487f-b51d-b1c047fc2ef8.mov

## Related issues

Closes #258 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Guidance/Guidance.tsx
+++ b/publisher/src/components/Guidance/Guidance.tsx
@@ -22,6 +22,7 @@ import { useNavigate } from "react-router-dom";
 import { useStore } from "../../stores";
 import {
   ActionButton,
+  ActionButtonWrapper,
   ContentContainer,
   GuidanceContainer,
   ProgressStepBubble,
@@ -88,17 +89,36 @@ export const Guidance = observer(() => {
           <TopicTitle>{currentTopicDisplayName}</TopicTitle>
           <TopicDescription>{currentTopicDescription}</TopicDescription>
 
-          {/* TODO(#) Replace the || "Next" and only display ActionButton if there is a buttonDisplayName property while mocking */}
-          <ActionButton
-            onClick={() => {
-              if (currentTopicID) {
-                if (pathToTask) navigate(pathToTask);
-                updateTopicStatus(currentTopicID, true);
-              }
-            }}
-          >
-            {buttonDisplayName || "Next"}
-          </ActionButton>
+          {currentTopicID === "ADD_DATA" ? (
+            <ActionButtonWrapper>
+              <ActionButton
+                kind="primary"
+                onClick={() => navigate("../upload")}
+              >
+                Upload spreadsheet
+              </ActionButton>
+              <ActionButton
+                kind="bordered"
+                onClick={() => navigate("../records")}
+              >
+                Fill out report
+              </ActionButton>
+            </ActionButtonWrapper>
+          ) : (
+            <>
+              {/* TODO(#) Replace the || "Next" and only display ActionButton if there is a buttonDisplayName property while mocking */}
+              <ActionButton
+                onClick={() => {
+                  if (currentTopicID) {
+                    if (pathToTask) navigate(pathToTask);
+                    updateTopicStatus(currentTopicID, true);
+                  }
+                }}
+              >
+                {buttonDisplayName || "Next"}
+              </ActionButton>
+            </>
+          )}
 
           {skippable && (
             <SkipButton
--- a/publisher/src/components/Guidance/Guidance.tsx
+++ b/publisher/src/components/Guidance/Guidance.tsx
@@ -20,6 +20,7 @@ import React, { useEffect } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 
 import { useStore } from "../../stores";
+import { REPORTS_LOWERCASE } from "../Global/constants";
 import {
   ActionButton,
   ActionButtonWrapper,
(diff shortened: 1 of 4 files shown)


publisher/src/router/Router.tsx
```diff
@@ -58,7 +58,29 @@ export const Router = () => {
               <>
                 <Route path="/settings/*" element={<Settings />} />
                 {isAddDataOrPublishDataStep && (
-                  <Route path="/upload/*" element={<DataUpload />} />
+                  <>
+                    <Route path="/upload/*" element={<DataUpload />} />
```
lilidworkin, line 62: Yea, so this is what I was worried about / hinting at in earlier PRs -- is there any way to share this logic between the two cases (guidance and not guidance) so we don't have to copy/paste and worry about forgetting to copy/paste a certain case, etc?

publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -88,17 +110,36 @@ export const Guidance = observer(() => {
...
-                updateTopicStatus(currentTopicID, true);
-              }
-            }}
-          >
-            {buttonDisplayName || "Next"}
-          </ActionButton>
+          {currentTopicID === "ADD_DATA" ? (
+            <ActionButtonWrapper>
+              <ActionButton
+                kind="primary"
```
lilidworkin, line 116: Random but I'm curious why this button is the primary one? I guess that was the case in Figma? Would you mind clarifying this with @agronomic and @hobuobi -- is it intentional that we're encouraging folks towards bulk upload?

publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -51,6 +54,25 @@ export const Guidance = observer(() => {
   const topLeftPositionedTopic =
     currentTopicID === "WELCOME" || currentTopicID === "METRIC_CONFIG";
 
+  useEffect(() => {
+    const initialize = async () => {
+      reportStore.resetState();
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      await reportStore.getReportOverviews(agencyId!);
```
terryttsai, line 61: Do we need to do this here?

publisher/src/components/Guidance/Guidance.tsx
mxosman: Oh `primary` was just a variable name to distinguish between the different colored buttons (blue vs no-background-grey-border). I could rename it `blue`? As far as which one should be blue and which one should be no-background but with a grey border - I just went with what was on the design and didn't really think about them as one pushing a user towards a particular button. Similar to other buttons side by side, usually their not both blue nor both no-background w/ a grey border.

publisher/src/components/Guidance/Guidance.tsx
mxosman: Yeah, I needed to initialize the report overviews to get a list of reports a user has uploaded to then do a check and see if there is at least one report - if so, then the topic is completed. Open to any suggestions you have if this doesn't feel right!

publisher/src/router/Router.tsx
mxosman: Yeah - this is a good question! I was having a hard time doing that and having it restrict access so I took this naive approach. I can make another attempt and note the exact issues I'm running into and see if there might not be a better solution that combines the two. Working out the routing was a bit of a doozy.

publisher/src/components/Guidance/Guidance.tsx
lilidworkin: Yea this was totally a product/design question, not an engineering one! Your implementation definitely looks right.

publisher/src/components/Guidance/Guidance.tsx
lilidworkin: I wonder if we can instead have the backend be responsible for determining which guidance topic we're currently on, and send that over with an existing GET request (like /users probably), and therefore reduce the number of round trips between FE and BE? That also lets us pull just the info we need to determine the guidance topic, whereas report overviews might include extraneous information and therefore waste resources. We don't have to do this in this PR -- would make more sense when you get around to working on the other backend changes -- but something to note for later?

publisher/src/router/Router.tsx
lilidworkin: Might be worth spending a little more time on it, but timebox it -- we can always return to it later.

lilidworkin (approved):
Will stamp this so we can merge and start shrinking this stack :D but make tasks for the two followups here (investigating shared routing and removing the getReportOverviews call). Thanks!

publisher/src/components/Guidance/Guidance.tsx
mxosman: Ah - this is brilliant! I love this idea - will include this in #329 to follow up on when I implement the backend! Thank you both for the feedback here!

publisher/src/router/Router.tsx
mxosman: Will do!

PR #269, opened 2023-01-04
Guidance: Publish Data

### LINKED ISSUE #259: [Frontend][Publisher] Guidance: Publish Data

mxosman:
## Description of the change

Implements the Publish Data step in the guidance flow. Here a user will see a list of their draft/unpublished records and be encouraged to publish data. The `Data` link to view the visualizations of the uploaded metrics is unlocked which means the rest of the app functionality is available to the user. If they haven't published data yet, every time they return to the app, they will be greeted by the "Publish Data" step of guidance.

Demo:

https://user-images.githubusercontent.com/59492998/210675307-38561b4d-7d1d-4d4c-a9dd-fec5df38529c.mov

NOTE: the "Next" button won't exist here in the final version (it's just to get you through to the next screen - which is completing the guidance journey)

## Related issues

Closes #259 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/PublishConfirmation.tsx
+++ b/publisher/src/components/Reports/PublishConfirmation.tsx
@@ -249,7 +249,7 @@ const MetricsDisplay: React.FC<{
 const PublishConfirmation: React.FC<{ reportID: number }> = ({ reportID }) => {
   const [isPublishable, setIsPublishable] = useState(false);
   const [metricsPreview, setMetricsPreview] = useState<MetricWithErrors[]>();
-  const { formStore, reportStore, userStore } = useStore();
+  const { formStore, reportStore, userStore, guidanceStore } = useStore();
   const { agencyId } = useParams();
   const navigate = useNavigate();
   const checkMetricForErrors = useCheckMetricForErrors(reportID);
@@ -268,6 +268,12 @@ const PublishConfirmation: React.FC<{ reportID: number }> = ({ reportID }) => {
       )) as Response;
 
       if (response.status === 200) {
+        // For users who have not completed the onboarding flow and are publishing for the first time.
+        if (
+          guidanceStore.currentTopicID === "PUBLISH_DATA" &&
+          !guidanceStore.hasCompletedOnboarding
+        )
+          guidanceStore.updateTopicStatus("PUBLISH_DATA", true);
         navigate(`/agency/${agencyId}/${REPORTS_LOWERCASE}`);
         showToast(
           `Congratulations! You published the ${printReportTitle(
@@ -298,6 +304,7 @@ const PublishConfirmation: React.FC<{ reportID: number }> = ({ reportID }) => {
     const { metrics, isPublishable: publishable } =
       formStore.validateAndGetAllMetricFormValues(reportID);
     const enabledMetrics = metrics.filter((metric) => metric.enabled);
+
     setMetricsPreview(enabledMetrics);
     setIsPublishable(publishable);
   }, [formStore, reportID]);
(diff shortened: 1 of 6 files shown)


publisher/src/components/Reports/ReviewReportDataEntry.tsx
```diff
@@ -44,9 +44,9 @@ const ReviewReportDataEntry = () => {
       }
     };
 
-    if (Object.keys(reportStore.reportOverviews).length === 0) {
-      initialize();
-    }
+    // if (Object.keys(reportStore.reportOverviews).length === 0) {
+    initialize();
+    // }
```
mxosman, line 49: I need to look into this more and make sure it's OK to remove or find an alternative solution. Since we already initialized the reportOverviews - this is never run, so you're not able to go directly from the Publish Data screen to review and publish. I haven't tested to make sure this is OK - mostly leaving a note to myself and calling it out incase there are other thoughts.

publisher/src/components/Reports/PublishConfirmation.tsx
```diff
@@ -268,6 +268,12 @@ const PublishConfirmation: React.FC<{ reportID: number }> = ({ reportID }) => {
       )) as Response;
 
       if (response.status === 200) {
+        // For users who have not completed the onboarding flow and are publishing for the first time.
```
lilidworkin, line 271: Ooh good call on this change!

lilidworkin (commented):
This looks good and non controversial to me!! 

Two tiny things: I noticed some weird nav behavior with the screen shaking a little, not sure if you're able to repro:
https://user-images.githubusercontent.com/99899591/214367895-f3ee0b22-d877-4933-8216-1bcd23346017.mov

I'm seeing scrollbars (because I have them explicitly turned on) but I wonder if we can remove the horizontal one?

publisher/src/components/Reports/ReviewReportDataEntry.tsx
```diff
@@ -39,14 +39,13 @@ const ReviewReportDataEntry = () => {
   useEffect(() => {
     const initialize = async () => {
       const result = await reportStore.getReport(reportID);
+      formStore.validatePreviouslySavedInputs(reportID);
```
terryttsai, line 42: Do we need to validate if the result is an Error?

mxosman: Good question! Should I move it below L45 and add a `return` on L44?

mxosman:
> This looks good and non controversial to me!!
> 
> Two tiny things: I noticed some weird nav behavior with the screen shaking a little, not sure if you're able to repro: https://user-images.githubusercontent.com/99899591/214367895-f3ee0b22-d877-4933-8216-1bcd23346017.mov

Ah - I think the screen shaking might be due to the loading animation of the report overviews appearing and disappearing very quickly once the reports load from the BE.

> I'm seeing scrollbars (because I have them explicitly turned on) but I wonder if we can remove the horizontal one? Screenshot 2023-01-24 at 12 05 36 PM

Great catch! I'll work on removing the horizontal bar.

publisher/src/components/Reports/PublishConfirmation.tsx
mxosman: Thank you, Lili! This one bit me and took up so much time to debug!

publisher/src/components/Reports/ReviewReportDataEntry.tsx
lilidworkin: That makes sense to me!

lilidworkin (dismissed):
Followups here seem small so this feels good to go!

PR #270, opened 2023-01-05
Agency Dashboard playtesting followups

terryttsai:
## Description of the change

<div>

- "24 months 2 years" becomes "4 years", "13 months" becomes "1 year 1 month"
Screen Shot 2023-01-05 at 9 47 59 AM
Screen Shot 2023-01-05 at 9 48 15 AM
Screen Shot 2023-01-05 at 9 50 28 AM
Screen Shot 2023-01-05 at 9 50 36 AM

- Fix legend in Race & Ethnicity overflow
- Add selected metric state for Select Metrics dropdown
- Fix “No reported data for this metric” placement in tablet and mobile
- Scrolling issue when navigating
- Make the url all lower case
- Fix issue with monthly values formatted as yearly (Agency 152 Readmissions show 4 2022 values)
  - I just made each bar say "month year" or "year" depending on whether that datapoint is annual, or monthly
Screen Shot 2023-01-05 at 9 48 22 AM

Screen Shot 2023-01-05 at 9 48 04 AM

- [Agency 152] Funding says 1 year of data in the overview, but the chart says No Data Available
  - The issue is that the frontend is receiving datapoints with Null values. In the aggregate case, it seems like we don't need this, but in the disaggregations case, that might be potentially useful because then we'd know what the dimensions are, even if they have no published values?
  - Or maybe we don't want the backend to serve any datapoints w only null values? Then how does the frontend know which dimensions exist?
  - Link to bug: https://dash---justice-counts-agency-dashboard-web-qqec6jbn6a-uc.a.run.app/agency/152
  - I changed my backend to point to https://dash---justice-counts-agency-dashboard-web-qqec6jbn6a-uc.a.run.app/ to verify the fix works
  - So now the overview should say No Data
- Back to Agency Overview button clickable area is too long
- Select Metrics button gap should be 8px
- Add “Close” to mobile modals

</div>

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/AgencyOverview.tsx
+++ b/agency-dashboard/src/AgencyOverview.tsx
@@ -77,35 +77,43 @@ const AgencyOverview = () => {
 
   const handleNavigate = (isPublished: boolean, metricKey: string) => {
     if (isPublished) {
-      navigate(`/agency/${agencyId}/dashboard?metric=${metricKey}`);
+      navigate(
+        `/agency/${agencyId}/dashboard?metric=${metricKey.toLocaleLowerCase()}`
+      );
     }
   };
 
   const getPublishCount = (
     datapoint: DatapointsGroupedByAggregateAndDisaggregations
   ): string => {
-    if (!datapoint.aggregate.length) {
+    const aggregateWithNonNullValues = datapoint.aggregate.filter(
+      (dp) => dp[DataVizAggregateName] !== null
+    );
+    if (!aggregateWithNonNullValues.length) {
       return "No Data";
     }
 
     let result = "";
 
-    const monthlyPublishes = datapoint.aggregate.filter(
+    const monthlyPublishes = aggregateWithNonNullValues.filter(
       (aggregation) => aggregation.frequency === "MONTHLY"
     );
-    const yearlyPublishes = datapoint.aggregate.filter(
+    const yearlyPublishes = aggregateWithNonNullValues.filter(
       (aggregation) => aggregation.frequency === "ANNUAL"
     );
 
-    if (monthlyPublishes.length > 0) {
-      result = `${monthlyPublishes.length} ${
-        monthlyPublishes.length > 1 ? "Months" : "Month"
-      } `;
+    const yearCount =
+      yearlyPublishes.length + Math.floor(monthlyPublishes.length / 12);
+
+    if (yearCount > 0) {
+      result += `${yearCount} ${yearCount > 1 ? "Years" : "Year"} `;
     }
 
-    if (yearlyPublishes.length > 0) {
-      result += `${yearlyPublishes.length} ${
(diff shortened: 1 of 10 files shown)


terryttsai:
One thing to note is the change in this commit: https://github.com/Recidiviz/justice-counts/pull/270/commits/4849bcebf838d1dd21d0c8c5c247651c9622a67a

I just implemented their suggestion to always scroll to the top of the page when navigating to a different page. See https://v5.reactrouter.com/web/guides/scroll-restoration for their suggestion.

lilidworkin (commented):
@terryttsai nice, thanks for knocking through all of these!

> Fix issue with monthly values formatted as yearly (Agency 152 Readmissions show 4 2022 values). I just made each bar say "month year" or "year" depending on whether that datapoint is annual, or monthly.

Would you mind adding a screenshot showing both before and after?

> [Agency 152] Funding says 1 year of data in the overview, but the chart says No Data Available

Can you explain what the bug / fix was? Was there actually data for this agency or not?

terryttsai:
Updated the description with screenshots and an explanation for "No Data Available" bug!

lilidworkin (approved):
> The issue is that the frontend is receiving datapoints with Null values. In the aggregate case, it seems like we don't need this, but in the disaggregations case, that might be potentially useful because then we'd know what the dimensions are, even if they have no published values?

Got it! So filtering out the datapoints with null values was the fix.

I think this seems reasonable. An alternative is to not send datapoints with null values, but the backend sends over the dimensions separately. In fact I thought I remembered that you made that change (or something similar) previously -- does that ring a bell?

In any case, this all looks great to me! I'll accept since @mxosman is out but he can take a look when he gets back.

lilidworkin:
The fix with the label under the bars in the bar chart looks so good and was very clever!!

terryttsai:
Yes I remember sending the dimensions over separately in the old API, though I no longer am with the new API. I just looked at the API payload and actually this is no longer an issue since we send over all the dimensions, even if there are no datapoints for them. This makes sense since this is similar to the metric config API.

So yeah, maybe the better fix would be for the backend to just not send datapoints with no values? Would that lead to any issues? Why are these datapoints being sent over in the first place? How much time should I spend digging into this, maybe I should write up a ticket to investigate this later?

lilidworkin:
@terryttsai yea it doesn't seem especially high priority -- I think filtering out null datapoints is totally fine for now! Not sure why we're sending them and if omitting would break anything, would need to be investigated.

terryttsai:
Made a ticket to investigate this: https://github.com/Recidiviz/justice-counts/issues/274

When I think about it, when you publish a report but some of the fields are not filled out, perhaps it's okay for those to show up as null values. So maybe this is expected? Is there any issue with the backend sending that over to the frontend?

mxosman:
LGTM!

PR #276, opened 2023-01-06
Share data viz with filters applied in the url params

### LINKED ISSUE #219: [Agency Dashboard] Put data viz filters and selected breakdown in query params
Put data viz filters and selected breakdown in query params so that the URL can be shared and users automatically see the data viz with the filters and selected breakdown.

terryttsai:
## Description of the change

Screen Shot 2023-01-05 at 10 21 16 PM
Screen Shot 2023-01-05 at 10 21 27 PM
Screen Shot 2023-01-05 at 10 21 33 PM

When changing the data viz filter settings, we store the selected option as a url search param. When loading the data viz page in both Publisher and Agency Dashboard, we apply the filter settings as stored in the url search params.

New "Share current view" option in the Share screen that shares the whole url with the filter settings applied. If unchecked, the filter settings are removed from the search params in the share url link.

When switching between metrics on both platforms, the filters should stay applied, if applicable, and the URL should be updated. When navigating away from data viz view, filters should be reset.

## Type of change

## Related issues

Closes #219

### Development

### Code review


common/components/DataViz/DatapointsView.tsx
```diff
@@ -163,16 +163,24 @@ export const DatapointsView: React.FC<{
     }
     if (!disaggregationOptions.includes(disaggregationName)) {
       setDisaggregationName(noDisaggregationOption);
-      setCountOrPercentageView("Count");
+      if (countOrPercentageView !== "Count") {
```
lilidworkin, line 166: should this be == rather than !==? or am I misunderstanding?

lilidworkin (approved):
This looks great @terryttsai !!

common/components/DataViz/DatapointsView.tsx
terryttsai: The changes in this file can be reverted actually. The purpose of this function is to reset the view back to Count in certain cases.

Originally I was making it so we don't set the search param to make the URL look cleaner but now I'm always setting the search param so we don't need this check anymore.

PR #294, opened 2023-01-11
Get rid of "N/A" in Includes/Excludes

### LINKED ISSUE #293: [Frontend] Get rid of "N/A" in Includes/Excludes
We'd like to get rid of the "N/A" option in each Includes/Excludes set. Instead, this should be a nullable setting, and if the value is null, the frontend should render the component with neither box checked.

In particular, after the backend makes its changes, the API should now look like:
```
    "settings": 
          [
              {
                  "key": "SETTING 1",
                  "included": "Yes",  # render Yes box checked
              },
          ],
          [
              {
                  "key": "SETTING 1",
                  "included": "No",  # render No box checked
              },
          ],
          [
              {
                  "key": "SETTING 1",
                  "included": null,  # render neither box checked
              },
          ]
```

(Original task link for reference: https://github.com/Recidiviz/recidiviz-data/issues/17760)

mxosman:
## Description of the change

Removes the "N/A" option in the Metric Definitions (Includes/Excludes). The backend may provide `null` for some of the includes/excludes fields and that will render as no options selected.

We used to have 3 visible options:
Screenshot 2023-01-11 at 5 16 59 PM

Now, there are only 2 (Yes / No):
Screenshot 2023-01-11 at 5 16 48 PM

Demo (what it looks like if the backend were to send `null` (in the `included` property) for the first two includes/excludes:
Screenshot 2023-01-11 at 5 07 41 PM

Since I had to hack together a mock response from the backend - it was tough to record a video of the behavior because I would have had to create some temp. state to update those null'd fields (totally doable). Let me know if the above works as a visual - or if you'd rather see something more interactive!

## Related issues

Closes #293

### Development

### Code review

### DIFF
--- a/common/types.ts
+++ b/common/types.ts
@@ -15,6 +15,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { MetricConfigurationSettingsOptions } from "../publisher/src/components/MetricConfiguration";
+
 export enum Permission {
   RECIDIVIZ_ADMIN = "recidiviz_admin",
   SWITCH_AGENCIES = "switch_agencies",
@@ -114,8 +116,6 @@ export type MetricDisaggregationDimensionsWithErrors =
   MetricDisaggregationDimensions & {
     error?: string;
   };
-
-export type MetricConfigurationSettingsOptions = "Yes" | "No" | "N/A";
 
 export type MetricConfigurationSettings = {
   key: string;
--- a/publisher/src/components/MetricConfiguration/types.ts
+++ b/publisher/src/components/MetricConfiguration/types.ts
@@ -21,7 +21,7 @@ import {
   ReportFrequency,
 } from "@justice-counts/common/types";
 
-export const metricConfigurationSettingsOptions = ["N/A", "No", "Yes"] as const;
+export const metricConfigurationSettingsOptions = ["No", "Yes"] as const;
 export type MetricConfigurationSettingsOptions =
   typeof metricConfigurationSettingsOptions[number];
 


lilidworkin (commented):
Oh wow, this was easier than I thought, haha! Thanks for knocking it out so quickly @mxosman 

cc @terryttsai  to maybe help look into the lint issues?

Out of curiosity @mxosman , just thinking about backwards compatibility, what would happen if the backend were to still send "N/A"? It shouldn't ever do this (once we make our changes) but I'm wondering if it's safe to merge this in before those are made

common/types.ts
```diff
@@ -15,6 +15,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { MetricConfigurationSettingsOptions } from "../publisher/src/components/MetricConfiguration";
```
terryttsai, line 18: Try moving `MetricConfigurationSettingsOptions` out of `publisher` and into the `common` directory? Would that help with the lint issues?

terryttsai: `agency-dashboard` and `publisher` imports from `common` via `"@justice-counts/common/*"`, but if you try to import via the relative directory (or try to import `publisher` from outside `publisher` or `agency-dashboard` from outside `agency-dashboard`, then you get problems

terryttsai: it looks like `agency-dashboard` is trying to import a file from `common`, and that file in `common` is trying to import something from `publisher` and that's causing the lint issues?

mxosman:
> Oh wow, this was easier than I thought, haha! Thanks for knocking it out so quickly @mxosman
> 
> cc @terryttsai to maybe help look into the lint issues?
> 
> Out of curiosity @mxosman , just thinking about backwards compatibility, what would happen if the backend were to still send "N/A"? It shouldn't ever do this (once we make our changes) but I'm wondering if it's safe to merge this in before those are made

Haha - I know, I thought the same thing! Took me a lot longer to mock things out to test it - I was so surprised to only see 3 lines changed!

Great question! If "N/A" is the value, it will treat it the same way it does `null` - it will show nothing selected. Got super lucky with this one - it only marks an option as selected if the `included` value is a "Yes" or "No" (whatever options we specify in `MetricConfigurationSettingsOptions`) - all other values for `included` will be treated as if no selection was made. Is that OK?

To test this, I saved the first 3 definitions as N/A - then ran it with the changes here:

Before:
Screenshot 2023-01-11 at 11 20 18 PM

After:
Screenshot 2023-01-11 at 11 20 28 PM

common/types.ts
mxosman: 🤦  Ugh - thank you, Terry. I did the "Quick Fix" import here in VSCode and didn't think to double check the import... I shouldn't be importing into common from publisher. Thank you for pointing me in the right direction!

lilidworkin:
> Great question! If "N/A" is the value, it will treat it the same way it does null - it will show nothing selected. Got super lucky with this one - it only marks an option as selected if the included value is a "Yes" or "No" (whatever options we specify in MetricConfigurationSettingsOptions) - all other values for included will be treated as if no selection was made. Is that OK?

This is perfect behavior! So great!

lilidworkin (approved):
🙌

terryttsai (approved):

PR #319, opened 2023-01-18
UI to display badge next to users that are admins

terryttsai:
## Description of the change

Display badge next to users that are admins, excluding the Agency Settings page which is under revamp

This displays the list of editors in rows instead of a comma-separated list in the data entry page, I should probably revert that change?

Screen Shot 2023-01-18 at 3 22 58 PM
Screen Shot 2023-01-18 at 3 23 04 PM
Screen Shot 2023-01-18 at 3 23 11 PM

Once the api is implemented I can hook this PR up with live data!

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review


publisher/src/components/Reports/ReportSummaryPanel.tsx
```diff
@@ -285,7 +287,12 @@ const ReportSummaryPanel: React.FC<{
         <EditDetailsTitle>Editors</EditDetailsTitle>
         <EditDetailsContent>
           {editors.length
-            ? printCommaSeparatedList(editors)
+            ? editors.map((editor) => (
+                <TeamMemberNameWithBadge
+                  name={editor}
+                  permission={Permission.RECIDIVIZ_ADMIN}
```
terryttsai, line 293: This is just to demonstrate, will remove this before merging (or hook up with live data if we have it

publisher/src/components/DataUpload/UploadedFiles.tsx
```diff
@@ -142,9 +145,13 @@ export const UploadedFileRow: React.FC<{
 
         {/* Date Uploaded */}
         <ExtendedCell capitalize>
-          <span>
-            {uploadedBy} / {dateUploaded}
-          </span>
+          <UploadedContainer>
+            <TeamMemberNameWithBadge
+              name={uploadedBy}
+              permission={Permission.AGENCY_ADMIN}
```
terryttsai, line 151: This is just to demonstrate, will remove this before merging (or hook up with live data if we have it

publisher/src/pages/Reports.tsx
```diff
@@ -226,7 +227,10 @@ const Reports: React.FC = () => {
                       "-"
                     ) : (
                       <>
-                        <span>{report.editors[0]}</span>
+                        <TeamMemberNameWithBadge
+                          name={report.editors[0]}
+                          permission={Permission.RECIDIVIZ_ADMIN}
```
terryttsai, line 232: This is just to demonstrate, will remove this before merging (or hook up with live data if we have it

publisher/src/components/DataUpload/UploadedFiles.tsx
lilidworkin: Makes sense! I wonder if it makes sense to add a (nullable) field to the UploadedFile type (and similarly, add a field to all the other types throughout) to make it clear to you and @nichelle-hall which APIs will need to start returning this information? Also that way we can think about how we convert a string like "agency-admin" (which is what we'd get from the API) to Permission.AGENCY_ADMIN

publisher/src/pages/Reports.tsx
lilidworkin: Looks like this change only applies to the first editor, do we also need to do something with the line `{printCommaSeparatedList(report.editors)}` below?

lilidworkin (commented):
This is great @terryttsai , thanks for taking a stab at this even before the backend was ready! 

regarding comma-separation -- is that going to be tricky now due to the way the divs work? If so we can probably live without it, but maybe give it a try?

Also, I can't quite tell from the spec, but are these icons supposed to be tooltips?
Screenshot 2023-01-18 at 8 56 37 AM

publisher/src/pages/Reports.tsx
terryttsai: So this is for the tooltip on hover, and for some reason my brain thought I wouldn't need to add badges to that but we do, in case one of the "others" is an admin, I'll do that!

terryttsai:
I think I should get the comma separation working, I'll work on that!

terryttsai:
As for the tooltip on hover, sorry I didn't mention it in my description. I think we can implement it pretty easily, was initially worried that we might have to implement it differently than the way we do it now if it doesn't fit in the parent div.

publisher/src/pages/Reports.tsx
lilidworkin: Got it that makes sense!!

publisher/src/pages/Reports.tsx
```diff
@@ -226,7 +227,10 @@ const Reports: React.FC = () => {
                       "-"
                     ) : (
                       <>
-                        <span>{report.editors[0]}</span>
+                        <TeamMemberNameWithBadge
+                          name={report.editors[0]}
```
terryttsai, line 231: So Lili with your suggestion on thinking how the role will come from the api, I think we'll need to either modify the ReportOverview response to include roles for each editor, or include the id and then look up the editor name and  role in the agency store.

terryttsai: Probably easiest to just do the former?

terryttsai: https://app.zenhub.com/workspaces/justice-counts--622109ea3bb6e0001d2d06ad/issues/gh/recidiviz/recidiviz-data/17963 ticket for this issue

terryttsai:
I'm going to put tooltips in a follow-up! Let me know how this is looking so far. I filed tickets so we can hook up the UI to live data later.

publisher/src/pages/Reports.tsx
lilidworkin: Awesome, thanks! Yea, I agree with that. In fact I'd go a step further -- _wherever_ the backend returns a list of name or id currently, it should be refactored to return a list of objects that include name and role. cc @nichelle-hall

lilidworkin (commented):
This looks great to me @terryttsai ! Are the issues that Humphrey pointed out [here](https://github.com/Recidiviz/justice-counts/pull/324#issuecomment-1396331139) issues with this PR, or the tooltip one on top of it? 

Before merging, you can take out the fake permissions and instead add a TODO in the code that points to the tickets you created?

terryttsai:
The issues are in this PR, so I'll be working on it here! (Or maybe I'll merge my tooltip changes into this branch?)

terryttsai:
Sounds good, I'll add TODOs to point to the tickets!

lilidworkin (approved):
Looks good to me! Before merging though let's take out the fake permissions and add TODOs in the code

publisher/src/components/primitives/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+    )}
+    {permission === Permission.AGENCY_ADMIN && (
+      <Tooltip
+        anchorId={badgeId}
+        content="Admin"
+        place="right"
+        noArrow
+        offset={6}
+        style={tooltipStyles}
+      />
```
mxosman, line 103: How did you like working with `react-tooltip`?

publisher/src/index.tsx
```diff
@@ -15,6 +15,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import "react-tooltip/dist/react-tooltip.css";
```
mxosman, line 18: Just curious - is it required to import the CSS for the tooltip in the index to get this library working?

terryttsai: Yeah it's part of the setup: https://react-tooltip.com/docs/getting-started#set-the-default-styling

publisher/src/components/primitives/index.tsx
terryttsai: It was good! Freed me from thinking about designing a tooltip system!

PR #327, opened 2023-01-21
[Bug] Prevent rendering of metric level context component when the context array is empty

### LINKED ISSUE #326: [Frontend][Publisher] Prevent rendering of metric level context component when the context array is empty

mxosman:
## Description of the change

It looks like there are some metrics that do not have a metric level context. This fix prevents us from rendering any of the context component (right now it just shows the header) when the metric level context array is empty. I thought that all metric level contexts have at least one free text box - but I'm not 100% sure. Either way, this helped expose the unnecessary rendering.

Screenshot 2023-01-20 at 8 04 43 PM

## Related issues

Closes #326 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/MetricContextConfiguration.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricContextConfiguration.tsx
@@ -78,10 +78,11 @@ export const ContextConfiguration: React.FC = observer(() => {
     }
   };
 
+  if (activeContextKeys.length === 0) return null;
+
   return (
     <MetricContextContainer enabled={metrics[systemMetricKey]?.enabled}>
       <MetricContextHeader>Context</MetricContextHeader>
-
       {activeContextKeys.map((contextKey) => {
         const currentContext = contexts[systemMetricKey][contextKey];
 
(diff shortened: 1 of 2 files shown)


lilidworkin (approved):
Nice, thanks for catching this @mxosman ! This is actually a valid state. We got rid of the "additional context" text box that accompanied all metrics. Now, a text box will only appear if an includes/excludes form also appears (the text box serves as a way for you to clarify your includes/excludes input). In this case, this metric doesn't have an includes/excludes defined for it yet (since the TIG is in progress), so no context is correctly appearing.

I thought in this case we used to show something like "No definitions are available for this metric yet" or something like that, do you know if that still appears or what happened to that? It might be nice to have a placeholder. Or else will the right sidebar just be totally blank right now? That's probably fine too.

mxosman:
> 



> Nice, thanks for catching this @mxosman ! This is actually a valid state. We got rid of the "additional context" text box that accompanied all metrics. Now, a text box will only appear if an includes/excludes form also appears (the text box serves as a way for you to clarify your includes/excludes input). In this case, this metric doesn't have an includes/excludes defined for it yet (since the TIG is in progress), so no context is correctly appearing.

Gotcha! Thank you for clarifying that! 

> I thought in this case we used to show something like "No definitions are available for this metric yet" or something like that, do you know if that still appears or what happened to that? It might be nice to have a placeholder. Or else will the right sidebar just be totally blank right now? That's probably fine too.

Ah - you know what? I think it only appears for the dimensions that don't have an includes/excludes - not the metric level! That's my bad - I'll make the necessary adjustments here to fix that! Thank you for calling this out!!

mxosman:
Now all metrics (not dimensions/breakdowns) without includes/excludes will include this message:

Screenshot 2023-01-23 at 8 14 36 PM

PR #341, opened 2023-01-26
[Fix] Adjustments to "Technical Definitions are not available for this metric yet." logic

### LINKED ISSUE #336: Tweaks to "Technical Definitions are not available for this metric yet." logic
After playing around a bit, I came up with a few ideas to make this a bit clearer -- lmk what you think and if this is easy to implement! 

**Case 1: Top-level metric does not have any includes/excludes defined.**
We should show "Technical Definitions are not available for this metric yet." (I believe this is the current behavior already!)

**Case 2: Dimension does not have any includes/excludes define, BUT it does have at least one context (i.e. all Other dimensions will fall in this category).**
We should _not_ show the "Technical Definitions are not available for this metric yet.", and should just show the text box. This will require a small change -- currently it looks like this:

**Case 3: Dimension does not have any includes/excludes define, and doesn't have a context either (all Unknown dimensions will fall in this category).**
We should say ""There are no technical definitions for this breakdown." instead of  "Technical Definitions are not available for this metric yet." This will require a small change.

mxosman:
## Description of the change

Adjusts the logic that displays the "Technical definitions are not available for this metric yet." for the top level metric and dimensions.

Description of the different cases as outlined in issue #336:

Case 1: Top-level metric does not have any includes/excludes defined.
We should show "Technical Definitions are not available for this metric yet." (I believe this is the current behavior already!)

Case 2: Dimension does not have any includes/excludes define, BUT it does have at least one context (i.e. all Other dimensions will fall in this category).
We should not show the "Technical Definitions are not available for this metric yet.", and should just show the text box.

Case 3: Dimension does not have any includes/excludes define, and doesn't have a context either (all Unknown dimensions will fall in this category).
We should say ""There are no technical definitions for this breakdown." instead of "Technical Definitions are not available for this metric yet." This will require a small change.

https://user-images.githubusercontent.com/59492998/214896450-88608331-a01c-4e6d-bb54-0d320c98f32a.mov

## Related issues

Closes #336 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricDefinitions.tsx
@@ -108,7 +108,6 @@ export const MetricDefinitions: React.FC<MetricDefinitionsProps> = observer(
 
     const hasMinOneContext =
       dimensionContextsMap && Object.values(dimensionContextsMap).length > 0;
-    console.log(hasMinOneContext);
 
     const noSettingsAvailable =
       !activeSettingsKeys ||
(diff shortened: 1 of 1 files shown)


lilidworkin (approved):
Logic looks perfect! Thanks!

PR #343, opened 2023-01-26
New frontend handling of supervision disaggregations

### LINKED ISSUE #321: Frontend changes to go along with "Change how supervision disaggregation works" #17802
This task is to make the frontend changes that go along with [this backend PR](https://github.com/Recidiviz/recidiviz-data/pull/17802).

Do the following:
1) Read that backend PR description and the actual code, to make sure you understand what changes are being made.
2) Check out that PR and run it locally. Go to a Supervision system. Make sure you've selected some supervision subsystems in Agency Settings, and then and try toggling on/off the disaggregated-by-supervision-subsystems flag for some metrics. Make sure you can see how (after a manual refresh) the enabled/disabled statuses are changing.
3) Think about what the minimal frontend changes we need to support this change are (i.e. remove the toast and redirect, update the enabled/disabled statuses in place without a manual refresh).
4) Think about what the *best* frontend experience would be. See Humphrey's comments in Slack [here](https://recidiviz.slack.com/archives/C02VDN1LSSW/p1674025862137109?thread_ts=1673993096.344509&cid=C02VDN1LSSW).

terryttsai:
## Description of the change

https://www.loom.com/share/9813a11e0a424d1cafcd11db85fb6cc2

Went through this with Humphrey: now that we see all supervision disaggregated metrics, we simply enable/disable them when they are toggled between combined and disaggregated.

This is enough to avoid the refresh experience. There is an even bigger design exercise in the future to make this more intuitive.

## Type of change

## Related issues

Closes #321

### Development

### Code review

### DIFF
--- a/common/types.ts
+++ b/common/types.ts
@@ -40,8 +40,9 @@ export type AgencyTeam = {
   email: string;
 };
 
-export const SupervisionSystems: AgencySystems[] = [
-  "SUPERVISION",
+export const SupervisionSystem: AgencySystems = "SUPERVISION";
+
+export const SupervisionSubsystems: AgencySystems[] = [
   "PAROLE",
   "PROBATION",
   "POST_RELEASE",
--- a/publisher/src/components/MetricConfiguration/Configuration.tsx
+++ b/publisher/src/components/MetricConfiguration/Configuration.tsx
@@ -16,8 +16,10 @@
 // =============================================================================
 
 import blueCheck from "@justice-counts/common/assets/status-check-icon.png";
-import { showToast } from "@justice-counts/common/components/Toast";
-import { SupervisionSystems } from "@justice-counts/common/types";
+import {
+  SupervisionSubsystems,
+  SupervisionSystem,
+} from "@justice-counts/common/types";
 import { printCommaSeparatedList } from "@justice-counts/common/utils";
 import { Dropdown } from "@recidiviz/design-system";
 import { observer } from "mobx-react-lite";
@@ -48,7 +50,6 @@ import {
   DropdownButton,
   Header,
   MetricConfigurationContainer,
-  MetricDisaggregations,
   MetricOnOffWrapper,
   PromptWrapper,
   RACE_ETHNICITY_DISAGGREGATION_KEY,
@@ -56,6 +57,7 @@ import {
   RadioButtonGroupWrapper,
   ReportFrequencyUpdate,
   Subheader,
+  TogglableSection,
 } from ".";
 
 type MetricConfigurationProps = {
@@ -67,7 +69,7 @@ type MetricConfigurationProps = {
   setActiveDisaggregationKey: React.Dispatch<
     React.SetStateAction<string | undefined>
   >;
-  supervisionSubsystems?: string[];
+  enabledSupervisionSubsystems?: string[];
 };
 
 export const Configuration: React.FC<MetricConfigurationProps> = observer(
@@ -76,7 +78,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
     setActiveDimensionKey,
     activeDisaggregationKey,
     setActiveDisaggregationKey,
-    supervisionSubsystems,
+    enabledSupervisionSubsystems,
   }): JSX.Element => {
     const { agencyId } = useParams();
     const [settingsSearchParams] = useSettingsSearchParams();
@@ -92,6 +94,7 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
       updateMetricReportFrequency,
       updateDisaggregatedBySupervisionSubsystems,
       saveMetricSettings,
+      initializeMetricConfigStoreValues,
     } = metricConfigStore;
 
     const { system: systemSearchParam, metric: metricSearchParam } =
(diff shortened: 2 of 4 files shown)


publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -144,34 +147,23 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
...
 
-        setTimeout(
-          () => showToast({ message: toastMessage, timeout: 5000 }),
-          1000
-        );
-        setTimeout(() => {
-          navigate("../metric-config");
-          window.location.reload();
-        }, 5000);
+        // After saving disaggregation selection, re-fetch metric settings
```
lilidworkin, line 163: Oh interesting way to solve this! I'm fine with this, but I'm wondering, what backend returned the new metric config values in the saveMetricSettings response? That way we'd avoid another round trip? I think we could make that change. (Would we need to return all of them, or could we just return the ones that changed?)

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -239,177 +231,191 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
...
-                <p>
-                  Disaggregations include the populations you selected in{" "}
-                  <BlueLinkSpan onClick={() => navigate("../agency-settings")}>
-                    Agency Settings
-                  </BlueLinkSpan>{" "}
-                  ({printCommaSeparatedList(capitalizedSupervisionSubsystems)}
-                  ).
-                </p>
-                <p>
-                  NOTE: Changing this option will refresh the page to reflect
```
lilidworkin, line 377: Do we still want to include a note explaining how this toggle will enable/disable this metric and enable/disable the other metrics in the other supervision subpopulations? Might be nice. cc @hobuobi

lilidworkin (approved):
@terryttsai this looks awesome! This is *exactly* what I had in mind. Just left two questions inline. I think regardless we should merge this in and can address the other items in a followup PR!

publisher/src/components/MetricConfiguration/Configuration.tsx
terryttsai: Down if that helps! I'm not sure if "Supervision subpopulation" maps to "Revocation (Parole)", "Revocation (Probation)" etc on its own or if users would already understand that

publisher/src/components/MetricConfiguration/Configuration.tsx
terryttsai: Since this whole experience is being redesigned anyways, I don't know if it's worth putting much more effort into making this cleaner. 

If we did want to make this cleaner, I was thinking if there's someway for the front-end to know how to map between a metric's combined and disaggregated versions, then the front-end can just modify the mobx store accordingly, without needing any additional information from the backend

mxosman: Ah - this is a super clever solution! Love both ideas here as well as far as reducing the # of round trip API calls.

publisher/src/components/MetricConfiguration/Configuration.tsx
```diff
@@ -239,177 +231,191 @@ export const Configuration: React.FC<MetricConfigurationProps> = observer(
...
-          )}
+          <TogglableSection
+            enabled={
+              !(
+                (disaggregatedBySupervisionSubsystems === true &&
+                  systemSearchParam === SupervisionSystem) ||
+                (disaggregatedBySupervisionSubsystems === false &&
+                  systemSearchParam &&
+                  SupervisionSubsystems.includes(systemSearchParam))
+              )
```
mxosman, line 242: super tiny nit & only if you feel moved: maybe worth putting this in a descriptive variable?

common/types.ts
```diff
@@ -40,8 +40,9 @@ export type AgencyTeam = {
   email: string;
 };
 
-export const SupervisionSystems: AgencySystems[] = [
-  "SUPERVISION",
+export const SupervisionSystem: AgencySystems = "SUPERVISION";
+
+export const SupervisionSubsystems: AgencySystems[] = [
```
mxosman, line 45: Love the way you broke these two apart!

mxosman (approved):

mxosman:
Awesome solution here Terry & Lili!

publisher/src/components/MetricConfiguration/Configuration.tsx
terryttsai: Good idea, it's a bit confusing!

publisher/src/components/MetricConfiguration/Configuration.tsx
lilidworkin: Hm yea, maybe let's discuss this in playtesting tomorrow and see if folks ahve thoughts!

publisher/src/components/MetricConfiguration/Configuration.tsx
lilidworkin: > If we did want to make this cleaner, I was thinking if there's someway for the front-end to know how to map between a metric's combined and disaggregated versions, then the front-end can just modify the mobx store accordingly, without needing any additional information from the backend

@terryttsai yes! There should be a way to do this, indeed. Something like this:
Aggregated:
SUPERVISION_STAFF enabled
PAROLE_STAFF disabled
PROBATION_STAFF disabled

Disaggregated: reverse

Basically, the metric key is SYSTEM_<METRIC_NAME>. The SYSTEMs of interest are SUPERVISION and then whichever other ones the agency has specified (which we should be already surfacing in the UI here, @mxosman would know about that). We should be able to determine the other metric keys by just doing a string replacement. LMK if that makes sense!

lilidworkin:
@terryttsai is it okay if I merge this so I can get it into a deploy for playtesting? We can address any followups in a separate PR!

terryttsai:
I'll merge this in for playtesting now!

PR #350, opened 2023-01-27
Guidance: Metric Configuration Tooltip Progress Tracker

### LINKED ISSUE #257: [Frontend][Publisher] Guidance: Configure Metrics - Tooltip Progress Tracker/Toast

mxosman:
## Description of the change

Improves the progress calculation logic and implements the tooltip progress tracker on the overview page and in the Metric Configuration as a user completes any of the 4 steps required.

Notes: 
* This was super tricky to implement and right now is only a working solution and not the most optimal solution. I plan to further optimize it and reduce the # of progress recalculations in future PRs - and as always, I'm open to your optimization suggestions.
* I disabled the tooltip in the overview screen for metrics that are marked as "Not Available" so it doesn't only show one of the 4 items checkmarked.
  *  Example (if you hover over Expenses, there will be no tooltip): Screenshot 2023-02-09 at 12 08 48 PM
  * If I were to let it display the tooltip, it will just show this (which I felt might be confusing?): Screenshot 2023-02-09 at 12 11 14 PM

Demo:

https://user-images.githubusercontent.com/59492998/217887672-39fa82a4-7ca5-4c0d-b5c5-07b63721a646.mov

**Question**: 
* When a user turns off a disaggregation - that disables all of the dimensions. Since we don't count a user not filling out the dimension definitions against them when a dimension is disabled/marked as unavailable, what should the tooltip treatment look like when all disaggregations are disabled?
  * Right now, the tooltip gives them a checkmark that they completed the "Confirm breakdown definitions" step, which might be misleading
 
    * Screenshot 2023-02-09 at 11 56 08 AM
  
  * When a user turns back on those disaggregations, the tooltip updates and removes the checkmark for the "Confirm breakdown definitions" 
    * Screenshot 2023-02-09 at 11 55 59 AM

## Related issues

Closes #257

### Development

### Code review


publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -338,23 +284,38 @@ export const Guidance = observer(() => {
...
                           "Action Required"}
                       </MetricStatus>
                       <RightArrowIcon />
+
+                      {/* Progress Tooltip */}
+                      <ProgressTooltipContainer>
+                        {metricConfigurationProgressSteps.map((step) => (
+                          <ProgressItemWrapper key={step}>
+                            <CheckIconWrapper>
+                              {(metricCompletionProgress[step] === 25 ||
```
lilidworkin, line 303: Why are we checking for either 25 or 100 percent here? (If it's valid, maybe add a comment to explain?)

publisher/src/components/Guidance/types.ts
```diff
@@ -68,6 +68,16 @@ export const onboardingTopicsMetadata: OnboardingTopicsMetadata = {
   },
 };
 
+export const metricConfigurationProgressSteps = [
+  "Confirm metric availability",
+  "Confirm metric definition",
+  "Confirm breakdowns",
```
lilidworkin, line 74: nit, but I wonder if this should be "Confirm breakdown availability" to be consistent? If you agree feel free to change, otherwise check with Humphrey?

publisher/src/stores/GuidanceStore.ts
```diff
@@ -68,6 +95,191 @@ class GuidanceStore {
...
+          ] = 0;
+        });
+      });
+      return this.metricConfigurationProgressStepsTracker[systemMetricKey][
+        "Confirm metric availability"
+      ];
+    }
+
+    if (
+      metrics[systemMetricKey]?.enabled !== null &&
```
lilidworkin, line 135: And the logic here is, if you enable a metric, we set completion of this step to 25? Why not 100?

publisher/src/stores/GuidanceStore.ts
```diff
@@ -68,6 +95,191 @@ class GuidanceStore {
...
+
+    if (
+      (metricDefinitionsCompleted ||
+        !metricDefinitionSettings[systemMetricKey]) &&
+      metrics[systemMetricKey]?.enabled
+    ) {
+      runInAction(() => {
+        this.metricConfigurationProgressStepsTracker[systemMetricKey][
+          "Confirm metric definition"
+        ] = 25;
```
lilidworkin, line 167: Also not sure why this is 25 and not 100

publisher/src/stores/GuidanceStore.ts
```diff
@@ -68,6 +95,191 @@ class GuidanceStore {
...
+    disaggregationValues?.forEach((disaggregation) => {
+      Object.values(disaggregation).forEach((dimension) => {
+        if (dimension.enabled === null) nullDimensions.push(dimension);
+      });
+    });
+
+    if (nullDimensions.length === 0 && metrics[systemMetricKey]?.enabled) {
+      runInAction(() => {
+        this.metricConfigurationProgressStepsTracker[systemMetricKey][
+          "Confirm breakdowns"
```
lilidworkin, line 192: Does it make sense to use an enum or constant here rather than repeating this string, in case of typos?

publisher/src/stores/GuidanceStore.ts
```diff
@@ -68,6 +95,191 @@ class GuidanceStore {
...
+      this.metricConfigurationProgressStepsTracker[systemMetricKey] = {
+        ...Object.fromEntries(
+          metricConfigurationProgressSteps.map((step) => [step, 0])
+        ),
+      };
+    });
+    this.isInitialized = true;
+  };
+
+  getMetricAvailabilityFrequencyProgress = (systemMetricKey: string) => {
```
lilidworkin, line 111: I think it would be great to give every method in here a docstring!

lilidworkin (commented):
Just took a brief look at this @mxosman , it looks great! A few high-level things:

* Some of the logic I'm not super clear on, and would appreciate some comments/docstrings. After that I can take a closer pass. 
* I think the design decisions you called out in the PR description were exactly right. We can discuss further in playtesting, but for now I think you made the exact right calls.
* Should we add unit tests for any of the complex/fragile logic?

@terryttsai should also take a look!

publisher/src/components/Guidance/GuidanceHeader.tsx
```diff
@@ -45,22 +64,86 @@ export const GuidanceHeader = observer(() => {
...
+  useEffect(
+    () => handleMetricConfigToastDisplay(),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [
+      metricProgress,
+      metricDefinitionProgress,
+      breakdownProgress,
+      breakdownDefinitionProgress,
+    ]
+  );
```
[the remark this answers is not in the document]
mxosman: Moreso performance - it recomputes everything after each render and I think we can do better utilizing mobX's full potential to minimize the amount of computation and take advantage of cached computations. I'm still in the weeds with using `autorun` here to observe the changes - it'll require refactoring the store and these methods a bit - but I'll test out how this feels in staging and see if there are any noticeable lags.

I can add a TODO here to revisit and improve this via mobX.

publisher/src/components/Guidance/Guidance.tsx
mxosman: I'll definitely add a comment if this sounds like a reasonable explanation/way of doing this:

`25`% is the total weight for completion in each of the 4 categories. But -- marking a metric unavailable is considered `100`% completion of the entire metric configuration. I made an exception and gave the "Confirm metric availability" category the ability to be `100`% (and all other categories `0`%) so when we total all 4 steps, it can display a check mark marking its completion if a user decides to turn the entire metric off. 

Does it make sense to do it this way?

publisher/src/components/Guidance/types.ts
mxosman: Agreed!

publisher/src/stores/GuidanceStore.ts
```diff
@@ -68,6 +95,191 @@ class GuidanceStore {
...
+    const { metrics } = this.metricConfigStore;
+
+    /** Confirm the metric’s availability/frequency */
+    if (metrics[systemMetricKey]?.enabled === false) {
+      runInAction(() => {
+        metricConfigurationProgressSteps.forEach((step) => {
+          if (step === "Confirm metric availability") {
```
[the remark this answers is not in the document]
mxosman: Exactly - I'll add a comment here to explain this.

publisher/src/stores/GuidanceStore.ts
mxosman: LMK if the above explanation made any sense!

publisher/src/stores/GuidanceStore.ts
mxosman: LMK if the above explanation made any sense!

publisher/src/stores/GuidanceStore.ts
mxosman: Great idea!

publisher/src/stores/GuidanceStore.ts
mxosman: Absolutely!

mxosman:
> * Some of the logic I'm not super clear on, and would appreciate some comments/docstrings. After that I can take a closer pass.
I'll add comments to clarify!

> * Should we add unit tests for any of the complex/fragile logic?

Can do!

publisher/src/components/Guidance/types.ts
mxosman: Done: [c38f85b](https://github.com/Recidiviz/justice-counts/pull/350/commits/c38f85bb9b21be3369800e1e345c7cc457060f6b)

publisher/src/stores/GuidanceStore.ts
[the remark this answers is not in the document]
mxosman: Done: [2c779db](https://github.com/Recidiviz/justice-counts/pull/350/commits/2c779db7818e22b4fe1430bc5e804d2580ec05f2)

publisher/src/stores/GuidanceStore.ts
mxosman: Done: [c38f85b](https://github.com/Recidiviz/justice-counts/pull/350/commits/c38f85bb9b21be3369800e1e345c7cc457060f6b)

publisher/src/stores/GuidanceStore.ts
mxosman: Done: [2c779db](https://github.com/Recidiviz/justice-counts/pull/350/commits/2c779db7818e22b4fe1430bc5e804d2580ec05f2)

publisher/src/components/Guidance/GuidanceHeader.tsx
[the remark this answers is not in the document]
lilidworkin: Yea, let's not worry about it _unless_ we notice a lag on staging! I think it's fine.

publisher/src/components/Guidance/Guidance.tsx
mxosman: > I think so ... so to make sure I'm understanding, the Confirm Metric Availability step is the _only_ one that could ever have a non-25 value? And that's why we don't have to check _which_ step we're on here?

Correct! Having the "Confirm metric availability" step be 25 means that this step is complete and the user still needs to complete the remaining 3 steps. Having the "Confirm metric availability" step be 100 means that the metric has been marked "Not Available", this step is complete and the user does not need to complete the remaining 3 steps.

> I think it might make more sense for the value of this dictionary to be a boolean -- completed or not. It seems strange to me that setting the value of metricCompletionProgress[step] to 25 means that this step is completed. I would expect metricCompletionProgress[step] = 100 to mean the step is completed. But really we don't need the full range of values 0 to 100 here -- we just need a 0 or 1. We actually don't even need the notion of percentages anywhere. When we fill up the progress bar, we can just sum up the values and divide by 4.

Completely understandable. We can do it that way.

> To handle the exception with metric unavailability, could we just have some special case logic at the point where we perform the sum, and explicitly check whether we're in this state, i.e. metricCompletionProgress[availability] == 1, and metricIsUnavailable? And then instead of setting sum to be sum of all availability (which would be 1 and not 4) we just override and set the sum to 4?
> 
> I think what you have is fine, and so if what I'm suggesting doesn't work, just lmk and we can keep this as is! Not worth spending a crazy amount of time on.

I'll go back and refactor this to operate in the way you're suggesting and give it a shot.

mxosman: Is this in line with what you were thinking? [43057c4](https://github.com/Recidiviz/justice-counts/pull/350/commits/43057c49124aab5ce725fdac521ecd904b0e6a18)

publisher/src/stores/GuidanceStore.ts
```diff
@@ -68,6 +96,233 @@ class GuidanceStore {
...
+      });
+    }
+
+    return this.metricConfigurationProgressStepsTracker[systemMetricKey]?.[
+      ProgressSteps.CONFIRM_BREAKDOWN_DEFINITIONS
+    ];
+  };
+
+  /** Computes progress on all 4 categories and returns the overall progress tracker object */
+  getOverallMetricProgress = (systemMetricKey: string) => {
```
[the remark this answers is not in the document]
mxosman: Thank you for the detailed feedback and for the sketched out proposal! My original thinking was that we wouldn't need to reflect those changes unless one of these functions are called - and that they will always be called when trying to figure out/display the current progress. But, your solution makes a lot of sense and is certainly more optimal and clearer than what I was going for!

I only wish I read this before hammering those tests on all the functions.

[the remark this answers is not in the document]
mxosman: > We could also split this computed value up into 4 computed values representing the 4 categories too

Let me try the other way first - where it's just one getter function. If we were to split them up, we'd need to find a way to persist/get the `systemMetricKey` since the getters can't take any params (e.g., useEffect gets the current `systemMetricKey` and stores it in the GuidanceStore, etc.).

mxosman:
> Love the design decision you made that you pointed out in your PR description!
> 
> For your design question, good job finding that case. What if we don't show the `Confirm breakdown definitions` step if there are none to confirm?

Great idea! Let me give that a shot.

mxosman:
> > Love the design decision you made that you pointed out in your PR description!
> > For your design question, good job finding that case. What if we don't show the `Confirm breakdown definitions` step if there are none to confirm?
> 
> Great idea! Let me give that a shot.

Hmm - this is much trickier than I thought. I'll need to map through the disaggregations to see which ones are turned off - or go through all dimension definitions to see if there are any to determine whether or not to disable that. Will think about this some more.

publisher/src/stores/GuidanceStore.ts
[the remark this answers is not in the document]
mxosman: This worked out really well - thank you for the suggestion, Terry! Let me know your thoughts: [58a5967](https://github.com/Recidiviz/justice-counts/pull/350/commits/58a59679e5ceb97bb30c0aa18b8d88dd0dbb8ada)

lilidworkin (approved):
Great, in that case I shall approve!

PR #352, opened 2023-01-30
Actually save draft with "Save as Draft" button

### LINKED ISSUE #346: "Save as Draft" button should actually save
I'm seeing some weird behavior where reports aren't getting saved properly. I *think* (but am not 100% sure) that this happens if you click the "Save as Draft" button quickly after making an edit, before autosave has a chance to kick in. "Save as Draft" doesn't actually save, so then you end up back on the reports back, but the save hasn't actually occurred. The value is saved in the frontend store, but not on the backend, so if you refresh, it's gone. 

Here's a repro:
https://images.zenhubusercontent.com/529290376/b0dca3a1-08ed-460b-9d5e-8ecb5a9ce896/screen_recording_2023_01_27_at_10_25_35_am.mov

terryttsai:
## Description of the change

Right now the "Save as Draft" button just navigates you back to the Reports page. Modify it to actually save the report as well. 

I wonder if we want to enforce saving when the user navigates away, like if they press the back button or home icon before autosave kicks in? This PR doesn't do that, but it's something we could do. 

## Type of change

## Related issues

Closes #346

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/DataEntryForm.tsx
+++ b/publisher/src/components/Reports/DataEntryForm.tsx
@@ -266,7 +266,10 @@ const DataEntryForm: React.FC<{
           </Button>
           <Button
             type="border"
-            onClick={() => navigate(`/agency/${agencyId}/${REPORTS_LOWERCASE}`)}
+            onClick={() => {
+              saveUpdatedMetrics();
+              navigate(`/agency/${agencyId}/${REPORTS_LOWERCASE}`);
+            }}
           >
             Save as Draft
           </Button>


mxosman:
> I wonder if we want to enforce saving when the user navigates away, like if they press the back button or home icon before autosave kicks in? This PR doesn't do that, but it's something we could do.

I believe this is accomplished in the useEffect here - I _think_ that navigating away or pressing the back button will trigger the `beforeunload` event:

https://github.com/Recidiviz/justice-counts/blob/c4e2eb60a561c94bbffe93fbeb8166bdae614bdf/publisher/src/components/Reports/DataEntryForm.tsx#L217-L231

terryttsai:
@mxosman If I press the JC icon or the back button really quickly, it doesn't save my latest changes, so I think I would probably add a useEffect hook on unmount to save?

mxosman:
> @mxosman If I press the JC icon or the back button really quickly, it doesn't save my latest changes, so I think I would probably add a useEffect hook on unmount to save?

Yeah - that's weird. I'm getting the same effect. If I add a `beforeunload` window event listener to any random website (in the browser console) and hit the back button, it triggers the callback function. I'm wondering if the unmount `removeEventListener` is called before it finishes saving - or if it even triggers it at all. Either way, I agree - it doesn't save when you hit the back button quickly!

Since that useEffect works for refreshing and other things - would you just add to this useEffect's unmount callback? 
```
     return () => {
       saveBeforeExiting(); // or just saveUpdatedMetrics();
       window.removeEventListener("beforeunload", saveBeforeExiting); 
     }
```
Or add a separate useEffect?

mxosman (approved):

lilidworkin (approved):
Thanks for this fix @terryttsai ! I'll let the two of you figure out this other issue with the back button -- that can be fixed in a different PR if we do decide to go for it.

terryttsai:
If I add saving to the unmount callback, since React in dev mode renders everything twice, it causes the save to be triggered when I navigate to the page. Not sure what's the best solution there. Maybe we can not show the toast prompt in this case? If backend is okay with the extra meaningless save calls

terryttsai:
Gonna merge this fix in for now, but we can continue the discussion

lilidworkin:
@terryttsai thanks to looking into this!

> If I add saving to the unmount callback, since React in dev mode renders everything twice, it causes the save to be triggered when I navigate to the page. Not sure what's the best solution there. Maybe we can not show the toast prompt in this case? If backend is okay with the extra meaningless save calls

Hmm this is tricky. I think given the current situation with the backend -- i.e. it not being in very good shape, lol -- we should probably do our best to avoid unnecessary API calls. I'm not 100% sure about this decision but that's my hunch.

I'm curious to get a bit of clarity on when we do save on leaving the page vs. when we don't. As @mxosman mentioned above, the `beforeunload` event handles some of these cases, but I guess from your conversation above there are at least two cases when it doesn't -- when you press the back button or JC home icon quickly? Is that correct? Are all other cases handled by it? Is there a way to detect _just_ those two actions? Mostly pressing the back button -- the JC home icon I'm a little less worried about since that seems like a rare case.

terryttsai:
Okay, good to know! We'll avoid doing unnecessary API calls. 

So I think the beforeunload event handles the case of closing the window or tab, but it doesn't handle the case of navigating away from the page (like pressing the back button for example). Should that implicitly save? 

Another thing we could possibly do is listen to route changes in the root router and if we detect that we're navigating away from this route, save it? But I don't like how unmaintainable that is, we're bound to forget about this.

Lastly, maybe we can call a function that checks if there are any new changes to save before sending the network request? But it sounds like that could lead to more complications if things get out of sync trying to track whether or not there are new changes.

lilidworkin:
Interesting ... this is tricky! Thanks so much for the detailed thoughts, @terryttsai -- I'm going to think about it a bit further and get back to you.

mxosman:
Yeah - this is tricky. I'm not sure how 100% reliable the "beforeunload" event is from what I've been reading - but here's a quick test to show that it does listen to the back button:

https://user-images.githubusercontent.com/59492998/216112399-5b49f9ed-8512-4abb-929c-d3bbda80c986.mov

I've been looking at the "visibilitychange" event - but haven't been able to get it to fire on back button... I'll keep thinking about this as well. 

RE: the unmount callback in a useEffect - would it be OK if it fires twice in dev mode knowing that it won't do that in production? Or would the issue be saving a blank report once it loads (before it gets all of the data from the backend)? It feels like the unmount callback covers all of our cases and only fires that extra unnecessary save in dev mode.

terryttsai:
Huh, I wasn't able to reproduce what you're showing in the video, I only see the console logs triggered when I refresh or close the tab.

lilidworkin:
@mxosman / @terryttsai I'm actually not too worried about this. I feel like if the user presses the back button instead of the "Save as Draft", they actually might not even _want_ their results to save, lol. I think as long as the "Save" button works (which this PR fixes!), I'm not especially worried about anything else.

terryttsai:
that's a good point! we'll leave it as-is for now, and if it's an issue we can revisit?

lilidworkin:
Yea that sounds great to me! Thanks both for digging into it so deeply!!

PR #367, opened 2023-02-01
Transition from auth0 permissions to roles

### LINKED ISSUE #344: [Agency Permissions][Frontend] Change logic for isRecidivizAdmin and isAgencyAdmin
Code [here](https://github.com/Recidiviz/justice-counts/blob/31f3e7f1d6472eb835badc2b7faed3893aa99e50/publisher/src/stores/UserStore.ts#L168)

These methods should instead be based on the `role` we receive from the backend.

@terryttsai I don't think we already had a task for the frontend part of this, so I'm adding one!

terryttsai:
## Description of the change

Transition from auth0 permissions to a user's agency role instead.

I also went ahead and removed using the assertion `agencyId!` in our codebase as well

## Type of change

## Related issues

Closes #344

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/CreateReport.test.tsx
+++ b/publisher/src/components/Reports/CreateReport.test.tsx
@@ -19,7 +19,7 @@ import { AgencyTeamMemberRole } from "@justice-counts/common/types";
 import { act, render, screen } from "@testing-library/react";
 import { runInAction } from "mobx";
 import React from "react";
-import { BrowserRouter } from "react-router-dom";
+import { BrowserRouter, MemoryRouter, Route, Routes } from "react-router-dom";
 
 import Reports from "../../pages/Reports";
 import { rootStore, StoreProvider } from "../../stores";
@@ -111,7 +111,7 @@ describe("test create report button", () => {
 
     await act(async () => {
       runInAction(() => {
-        rootStore.userStore.permissions = [""];
+        rootStore.userStore.userAgencies = [];
       });
     });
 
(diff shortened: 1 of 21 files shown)


publisher/src/components/Reports/CreateReport.test.tsx
```diff
@@ -125,15 +125,40 @@ describe("test create report button", () => {
   test("created reports button should be displayed if user has permission", async () => {
     render(
       <StoreProvider>
-        <BrowserRouter>
-          <Reports />
-        </BrowserRouter>
+        <MemoryRouter initialEntries={["/agency/590/reports"]}>
```
terryttsai, line 128: @mxosman @ivan-kishko I followed this example to mock the agencyId browser param: https://tomalexhughes.com/blog/testing-components-that-use-react-router-hooks

Only the Decoupled solution worked for me, when I tried mocking useParams I wasn't able to get it to return the mocked value for some reason

publisher/src/stores/UserStore.ts
```diff
@@ -138,7 +132,17 @@ class UserStore {
     return undefined;
   }
 
-  getAgency(agencyId: string | undefined): UserAgency | undefined {
+  get userAgenciesById(): { [agencyId: string]: UserAgency } {
```
lilidworkin, line 135: Just out of curiosity / for my own learning, what's the difference between having this here like you do now, vs. how it was before, where it was in the `runInAction`?

publisher/src/stores/UserStore.ts
```diff
@@ -165,12 +169,27 @@ class UserStore {
     return this.authStore.user?.email_verified;
   }
 
-  get isRecidivizAdmin(): boolean {
-    return this.permissions.includes(Permission.RECIDIVIZ_ADMIN);
+  getUserAgencyRole(agencyId: string): AgencyTeamMemberRole | undefined {
```
lilidworkin, line 172: This logic looks great to me!

lilidworkin (approved):
This looks great to me @terryttsai ! Thanks for taking the initiative to fix up the agencyID issue too!

publisher/src/components/Reports/CreateReport.test.tsx
ivan-kishko: isn't it called "records" now? instead of reports?

publisher/src/stores/UserStore.ts
terryttsai: So we already store `userAgencies` as an array in the UserStore. Instead of also manually setting `userAgenciesById`, which maps the agencies by id, we can just have mobx compute it automatically whenever `userAgencies` is updated: https://mobx.js.org/computeds.html

That way, we only have to update one place and not have to remember to update both places

lilidworkin: Oh fantastic, that's great!

publisher/src/components/Reports/CreateReport.test.tsx
ivan-kishko: I meant value You put in initialEntries, now in our app url looks like this with word records. I know that component name is Reports but url is changed to contain word records. Or am I missing something?
Screen Shot 2023-02-02 at 8 46 20 PM

terryttsai: Yes, the component's name is still "Reports" but the app displays "Records" in the app and in the app url. We renamed Reports to Records because the requirements changed, but we didn't rename the code to say "Records", since we're not sure if it's going to change again in the future. 

Do you think we should rename "Reports" to "Records" to avoid confusion? It must be kind of confusing to see this change happen without knowing much about it!

mxosman: I see what @ivan-kishko is referring to. I think the URL path `/agency/590/reports` won't work unless its `/agency/590/records` since we've updated all of the routes (we do have a re-direct in place - so if you go to `.../reports` it will replace the URL with `.../records`). Throughout the code, I believe we have routes to the records page as something like this `/agency/:agencyId/${REPORTS_LOWERCASE}`

`REPORTS_LOWERCASE` translating to `records`

PR #370, opened 2023-02-02
Fix initializing metric config values

terryttsai:
## Description of the change

Fix a couple issues with initializing metric config store values from this refactor: https://github.com/Recidiviz/justice-counts/pull/315

- Make sure to create nested objects so we can access them later
- Make sure to only create new nested objects if they hadn't been created yet

This fixes the two issues of not being able to load the metric config page and accidentally only setting one dimension in each metric.

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
@@ -130,7 +130,6 @@ export const MetricConfiguration: React.FC = observer(() => {
 
   useEffect(() => {
     const initialize = async () => {
-      setIsLoading(true);
       await initializeMetricConfiguration();
     };
 
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -366,7 +366,9 @@ class MetricConfigStore {
     );
 
     /** Initialize nested objects for quick lookup and update and reduce re-renders */
-    this.metricDefinitionSettings[systemMetricKey] = {};
+    if (!this.metricDefinitionSettings[systemMetricKey]) {
+      this.metricDefinitionSettings[systemMetricKey] = {};
+    }
     this.metricDefinitionSettings[systemMetricKey][settingKey] =
       metricDefinitionSettings;
   };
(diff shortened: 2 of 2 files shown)


publisher/src/components/MetricConfiguration/MetricConfiguration.tsx
```diff
@@ -130,7 +130,6 @@ export const MetricConfiguration: React.FC = observer(() => {
 
   useEffect(() => {
     const initialize = async () => {
-      setIsLoading(true);
```
terryttsai, line 133: Removed this since it is not necessary since it's already set to true on mount

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -404,9 +408,15 @@ class MetricConfigStore {
     );
 
     /** Initialize nested objects for quick lookup and update and reduce re-renders */
-    this.dimensions[systemMetricKey] = {};
-    this.dimensions[systemMetricKey][disaggregationKey] = {};
-    this.dimensions[systemMetricKey][disaggregationKey][dimensionKey] = {};
+    if (!this.dimensions[systemMetricKey]) {
```
terryttsai, line 411: If we don't check for existence first, then we're always setting dimensions[systemMetricKey] back to {} for each metric we add

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -702,6 +729,24 @@ class MetricConfigStore {
...
+      !this.dimensionDefinitionSettings[systemMetricKey][disaggregationKey][
+        dimensionKey
+      ]
+    ) {
+      this.dimensionDefinitionSettings[systemMetricKey][disaggregationKey][
+        dimensionKey
+      ] = {};
+    }
+
     this.dimensionDefinitionSettings[systemMetricKey][disaggregationKey][
```
terryttsai, line 750: Since the above was missing we were getting key errors trying to access the nested object that doesn't exist

publisher/src/stores/MetricConfigStore.tsx
terryttsai: Which meant this wipes out every dimension we set except for the last one

mxosman (approved):
Nice catch & fix - completely missed this in my review!

publisher/src/stores/MetricConfigStore.tsx
mxosman: Yup - makes sense!

terryttsai:
Haha same for me!

lilidworkin:
Thanks for fixing this @terryttsai !!

PR #372, opened 2023-02-03
Update Supervision name to be "Supervision (Combined)" for supervision agencies with subsystems

### LINKED ISSUE #356: [Frontend][Publisher] Small refactor to the Supervision tab names
A given Supervision agency can oversee both Prisons and Supervisions. Supervisions can then be broken down into Parole, Probation, Pretrial, and Other.
 
Currently in the Metric Configuration page, these 6 categories are shown in a 'flat' row. To distinguish this hierarchy, we want to implement a quick fix and rename this tabs as follows:
Prisons, Supervisions: Combined,  Supervisions: Parole,  Supervisions: Probation,  Supervisions: Pretrial,  Supervisions: Other

mxosman:
## Description of the change

In general, this formats all areas the render a system name and specifically formats the Supervision system name to "Supervision (Combined)" if the Supervision system has subsystems (e.g. Parole, Probation, etc.)

Thanks to @lilidworkin's suggestion - there's a helper function that helps format the system name and if provided a list of systems can determine whether or not to add "(Combined)" to the Supervision system name. We'll just have to remember any time we're rendering a system name to feed it into the helper function for consistency.

Please let me know if there are areas I may have missed that render the system name. The only areas I was unsure about updating were 3 of the bulk upload screens:

Screenshot 2023-02-03 at 4 21 32 PM

Screenshot 2023-02-03 at 4 21 42 PM

Screenshot 2023-02-03 at 4 23 01 PM

Demo (Supervision agency with subsystems):

https://user-images.githubusercontent.com/59492998/216714044-7f0bb2e2-f46b-4d52-972d-f2b6c254f046.mov

Playtesting Link: https://mahmoudtest---justice-counts-web-qqec6jbn6a-uc.a.run.app

It's not as straightforward to test a Supervision only agency because with a single system agency, things aren't split up/categorized by system name. Nevertheless, the system names should appear formatted as expected.

## Related issues

Closes #356 

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataUpload/UploadedFiles.tsx
+++ b/publisher/src/components/DataUpload/UploadedFiles.tsx
@@ -257,10 +257,10 @@ export const UploadedFiles: React.FC = observer(() => {
       badgeText: fileStatus?.toLowerCase() || "Uploading",
       dateUploaded: formatDate(file.uploaded_at),
       dateIngested: file.ingested_at ? formatDate(file.ingested_at) : "--",
-      system: formatSystemName(
-        file.system,
-        currentAgency?.systems
-      ) as AgencySystems,
+      system: formatSystemName(file.system, {
+        allUserSystems: currentAgency?.systems,
+        hideCombined: true,
+      }) as AgencySystems,
       uploadedBy: file.uploaded_by,
     };
   };
--- a/publisher/src/components/DataUpload/UploadedFiles.tsx
+++ b/publisher/src/components/DataUpload/UploadedFiles.tsx
@@ -21,12 +21,13 @@ import {
   BadgeColors,
 } from "@justice-counts/common/components/Badge";
 import { showToast } from "@justice-counts/common/components/Toast";
+import { AgencySystems } from "@justice-counts/common/types";
 import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 
 import { useStore } from "../../stores";
-import { removeSnakeCase } from "../../utils";
+import { formatSystemName } from "../../utils";
 import downloadIcon from "../assets/download-icon.png";
 import { Title, TitleWrapper } from "../Forms";
 import { Loader } from "../Loading";
@@ -61,7 +62,7 @@ export const UploadedFileRow: React.FC<{
     badgeText: string;
     dateUploaded: string;
     dateIngested: string;
-    system?: string;
+    system?: AgencySystems;
     uploadedBy: string;
   };
   deleteUploadedFile: (spreadsheetID: number) => void;
@@ -209,7 +210,8 @@ export const UploadedFileRow: React.FC<{
 
 export const UploadedFiles: React.FC = observer(() => {
   const { agencyId } = useParams() as { agencyId: string };
-  const { reportStore } = useStore();
+  const { reportStore, userStore } = useStore();
+  const currentAgency = userStore.getAgency(agencyId);
   const dataUploadColumnTitles = [
     "Filename",
     "Uploaded",
@@ -255,7 +257,10 @@ export const UploadedFiles: React.FC = observer(() => {
       badgeText: fileStatus?.toLowerCase() || "Uploading",
       dateUploaded: formatDate(file.uploaded_at),
       dateIngested: file.ingested_at ? formatDate(file.ingested_at) : "--",
-      system: removeSnakeCase(file.system).toLowerCase(),
+      system: formatSystemName(
+        file.system,
+        currentAgency?.systems
+      ) as AgencySystems,
       uploadedBy: file.uploaded_by,
     };
   };
--- a/publisher/src/components/Forms/Form.styles.tsx
+++ b/publisher/src/components/Forms/Form.styles.tsx
@@ -98,7 +98,7 @@ export const OnePanelBackLinkContainer = styled(PreTitle)`
   }
 `;
 
-export const MetricsSectionTitle = styled.div`
+export const MetricSummarySectionTitle = styled.div`
   ${typography.sizeCSS.normal}
   text-transform: capitalize;
   margin-top: 6px;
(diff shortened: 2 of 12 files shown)


publisher/src/components/Reports/DataEntryForm.tsx
```diff
@@ -336,7 +341,10 @@ const DataEntryForm: React.FC<{
                 <Fragment key={system}>
                   {showMetricSectionTitles && (
                     <MetricSystemTitle firstTitle={systemIndex === 0}>
-                      {system}
+                      {formatSystemName(
+                        system as AgencySystems,
```
mxosman, line 345: Had to typecast in these areas where we are mapping `Object.entries` since keys can only be of `string` type

publisher/src/components/Forms/Form.styles.tsx
```diff
@@ -143,6 +144,7 @@ export const MetricSectionTitleWrapper = styled.div`
 
 export const MetricSectionTitle = styled.div<{ notReporting?: boolean }>`
```
mxosman, line 145: Just noticed that we have a `MetricsSectionTitle` and a `MetricSectionTitle` - but both are differently styled. 😆 I'll figure out a different name for these as they're used in different areas for different things.

publisher/src/components/DataUpload/UploadedFiles.tsx
```diff
@@ -255,7 +257,10 @@ export const UploadedFiles: React.FC = observer(() => {
       badgeText: fileStatus?.toLowerCase() || "Uploading",
       dateUploaded: formatDate(file.uploaded_at),
       dateIngested: file.ingested_at ? formatDate(file.ingested_at) : "--",
-      system: removeSnakeCase(file.system).toLowerCase(),
+      system: formatSystemName(
```
lilidworkin, line 260: Let's actually not change it here. I believe the rule of thumb should be as follows: any place where it's _possible_ that the word Parole or Probation could appear too, we should use Supervision (combined). If Parole or Probation could never appear, we just use Supervision. Since users upload a single spreadsheet for all of Supervision (i.e. the same spreadsheet contains rows for combined, parole, and probation) it's impossible for an uploaded file to be of type Parole or Probation, so those words could never appear here, so according to the above rule of thumb we just keep it as Supervision. LMK if that makes any sense. Thanks!

lilidworkin: With that said, maybe we still call the helper function (for consistency, so we can always feel comfortable using that helper everywhere) but we use a flag to determine whether we add (Combined) or not?

publisher/src/utils/helperUtils.ts
```diff
@@ -232,3 +236,30 @@ export function removeAgencyFromPath(location: string) {
...
+ * @note formats SUPERVISION system name to "Supervision (Combined)" if supervision subsystems detected
+ * @param systemName {string} system name
+ * @param allUserSystems {string[]} (optional) a list of all systems a user belongs to
+ * @returns lowercase, non-snakecase string
+ *
+ * @example formatSystemName("SUPERVISION", ["SUPERVISION", "PAROLE", "PROBATION", "POST_RELEASE"]) returns "supervision (combined)"
+ * @example formatSystemName("SUPERVISION", ["SUPERVISION"]) returns "supervision"
+ * @example formatSystemName("COURTS_AND_PRETRIAL") returns "courts and pretrial"
+ */
+export const formatSystemName = (
```
lilidworkin, line 252: Looks great!

lilidworkin (commented):
Looks great @mxosman ! Just two changes:

* Let's _not_ show "combined" on the UploadedFiles page (explained inline)
* Let's _do_ show "combined" on the Bulk Upload "we encountered X errors page". Reason being, there we DO show Parole/Probation, so it makes sense to show Supervision (combined) to differentiate.

LMK if this rule of thumb makes sense.

I think in all cases we should call the helper function, so that we can centralize the logic, but the function should take a parameter that indicates whether "combined" should be added or not. LMK if that makes sense too.

Thanks!

mxosman:
> Looks great @mxosman ! Just two changes:
> 
> * Let's _not_ show "combined" on the UploadedFiles page (explained inline)
> * Let's _do_ show "combined" on the Bulk Upload "we encountered X errors page". Reason being, there we DO show Parole/Probation, so it makes sense to show Supervision (combined) to differentiate.
> 
> LMK if this rule of thumb makes sense.
> 
> I think in all cases we should call the helper function, so that we can centralize the logic, but the function should take a parameter that indicates whether "combined" should be added or not. LMK if that makes sense too.
> 
> Thanks!

Thank you for the direction here! That makes sense - I'll update accordingly!

publisher/src/components/DataUpload/UploadedFiles.tsx
mxosman: Thank you for the explanation, Lili! I like the idea of continuing to call this function - I'll add an additional optional parameter to include the mentioned flag!

mxosman:
> * Let's _not_ show "combined" on the UploadedFiles page (explained inline)
> * Let's _do_ show "combined" on the Bulk Upload "we encountered X errors page". Reason being, there we DO show Parole/Probation, so it makes sense to show Supervision (combined) to differentiate.
>
> I think in all cases we should call the helper function, so that we can centralize the logic, but the function should take a parameter that indicates whether "combined" should be added or not. LMK if that makes sense too.

All changes were made in this commit: [6ff2ac7](https://github.com/Recidiviz/justice-counts/pull/372/commits/6ff2ac7c158721c9e845a5e6677290e617b35506)

Would love to hear your thoughts! I created an optional `options` param object that will house any customizations we need - it includes a `hideCombined` boolean property that will allow us to override displaying the "(Combined)" for those special cases.

Now, the Error/Warnings page will show the "(Combined)":
Screenshot 2023-02-06 at 2 50 48 PM


And, the UploadedFiles component will NOT show the "(Combined)":
Screenshot 2023-02-06 at 2 52 01 PM

lilidworkin (approved):
Perfect, thanks @mxosman !

PR #376, opened 2023-02-03
Adjust ReportStore tests to reduce the run time to < 5s

### LINKED ISSUE #377: [Frontend][Publisher] Fix flaky ReportStore test

mxosman:
## Description of the change

Simplifies and removes `async` in one of our tests to prevent it from taking longer than 5 seconds to pass (which fails the CI test).

## Related issues

Closes #377

### Development

### Code review

### DIFF
--- a/publisher/src/stores/ReportStore.test.tsx
+++ b/publisher/src/stores/ReportStore.test.tsx
@@ -92,7 +92,7 @@ test("displayed reports", async () => {
     </StoreProvider>
   );
 
-  await act(() => {
+  act(() => {
     runInAction(() => {
       rootStore.reportStore.loadingOverview = false;
       rootStore.reportStore.reportOverviews = mockUnorderedReportsMap;
--- a/publisher/src/stores/ReportStore.test.tsx
+++ b/publisher/src/stores/ReportStore.test.tsx
@@ -92,7 +92,7 @@ test("displayed reports", async () => {
     </StoreProvider>
   );
 
-  await act(() => {
+  await act(async () => {
     runInAction(() => {
       rootStore.reportStore.loadingOverview = false;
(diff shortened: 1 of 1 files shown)


terryttsai:
Thanks for looking into this Mahmoud! Do you know or have theories as to why this fix works?

publisher/src/stores/ReportStore.test.tsx
```diff
@@ -92,7 +92,7 @@ test("displayed reports", async () => {
     </StoreProvider>
   );
 
-  await act(async () => {
+  await act(() => {
```
terryttsai, line 95: Do we still need the `await` if we remove the async function?

mxosman: It failed when I removed it.

mxosman:
> Thanks for looking into this Mahmoud! Do you know or have theories as to why this fix works?

Honestly - no clue. I'm not sure how removing `async` reduces the time it takes to run. Perhaps it runs the function synchronously in the same thread and that reduces the test time by a couple seconds?

publisher/src/stores/ReportStore.test.tsx
mxosman: Actually - I just tried it again and it worked - must've been something else I was doing when I tried removing it. Pushing up the changes now.

mxosman:
OK - it looks like having it run synchronously with no `await` did the trick.

terryttsai (approved):

terryttsai:
This looks good to me! If there's no need to make act async then let's simplify the code!

mxosman:
Thank you for your eyes, Terry! These weird cases always throw me for a loop!

PR #378, opened 2023-02-03
Disambiguate annual metrics in bar chart

terryttsai:
## Description of the change

Annual metrics may show up on the bar chart of the same year but different starting months. Right now they would just display the year but it's confusing what's the time range of these years.

I tried out a few ideas and decided on the last image. 

First, I tried adding "(CY)" and "(FY)" to annual datapoints, but it may not be clear what "CY" or "FY" means, plus it doesn't support custom starting months.

Screen Shot 2023-02-03 at 12 45 11 PM

Then I tried adding the starting month back in and adding "(Annual)", but what does it mean to be "Jan 2022 (Annual)"?

Screen Shot 2023-02-03 at 12 46 01 PM

Lastly, in this PR, I tried just listing the date range "Jan 2022 - Dec 2022", "Jul 2022 - Jun 2023" and it's the most clear but wordiest.

Screen Shot 2023-02-03 at 12 55 45 PM

Thoughts?

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/common/components/DataViz/utils.ts
+++ b/common/components/DataViz/utils.ts
@@ -43,7 +43,7 @@ export const nextMonthMap = new Map<string, string>([
   ["Dec", "Jan"],
 ]);
 
-const abbreviatedMonths = [
+export const abbreviatedMonths = [
   "Jan",
   "Feb",
   "Mar",
--- a/common/components/DataViz/BarChart.tsx
+++ b/common/components/DataViz/BarChart.tsx
@@ -31,7 +31,7 @@ import { Datapoint } from "../../types";
 import { rem } from "../../utils";
 import { palette } from "../GlobalStyles";
 import Tooltip from "./Tooltip";
-import { splitUtcString } from "./utils";
+import { abbreviatedMonths, splitUtcString } from "./utils";
 
 const MAX_BAR_SIZE = 150;
 
@@ -215,9 +215,21 @@ const ResponsiveBarChart: React.FC<{
                 return "";
               }
               const [, , month, year] = splitUtcString(value);
-              return data[index].frequency === "ANNUAL"
-                ? year
-                : `${month} ${year}`;
+              if (data[index].frequency === "ANNUAL") {
+                const previousMonth =
+                  abbreviatedMonths[
+                    (abbreviatedMonths.findIndex((m) => m === month) + 11) % 12
+                  ];
+                if (previousMonth === "Dec") {
+                  return `${month} ${year} - ${previousMonth} ${parseInt(
+                    year
+                  )}`;
+                }
+                return `${month} ${year} - ${previousMonth} ${
+                  parseInt(year) + 1
+                }`;
+              }
+              return `${month} ${year}`;
             }}
             tickMargin={12}
           />


hobuobi:
I actually really like this treatment! Two comments: 
1. What does this look like at smaller window sizes? The chart component in Publisher doesn't completely adjust to the size of the window yet, so it's hard for me to test this. 
2. I think we should still highlight the starting month (i.e., make "- [MONTH YEAR]" gray or something). 
3. Haven't yet tested what this looks like if there are both monthly and yearly data for that metric.

terryttsai:
Screen Shot 2023-02-03 at 3 29 43 PM
At smaller bar widths, our charting library just starts hiding labels but you can still see them on hover

common/components/DataViz/utils.ts
```diff
@@ -349,3 +349,18 @@ export const formatDateShort = (dateStr: string) => {
 export const formatExternalLink = (url: string) => {
   return url.match(/^http[s]?:\/\//) ? url : `http://${url}`;
 };
+
+export const getDatapointBarLabel = (datapoint: Datapoint) => {
```
lilidworkin, line 353: I like that you moved this logic into its own function, so it'll be easy to adjust in a centralized place later if needed!

lilidworkin (approved):
Cool, I'm happy with this @terryttsai ! I love how you tried multiple things and explained your though process in the PR description. 😍

My main goal is to make things easy for us to debug right now, and this for sure accomplishes that. We can always tweak the treatment later if we want.

common/components/DataViz/utils.ts
mxosman: +1

common/components/DataViz/utils.ts
```diff
@@ -349,3 +349,18 @@ export const formatDateShort = (dateStr: string) => {
...
+  const [, , month, year] = splitUtcString(datapoint.start_date);
+  if (datapoint.frequency === "ANNUAL") {
+    const previousMonth =
+      abbreviatedMonths[
+        (abbreviatedMonths.findIndex((m) => m === month) + 11) % 12
```
mxosman, line 358: Q: any risk here of `findIndex` returning `-1` and giving an incorrect month in the arithmetic?

lilidworkin:
Hey @terryttsai ! Ugh, that isn't good, thanks for catching this! I think we should spend a little more time investigating since it's kind of a bad bug ... here are a few questions/things to look into:

* Has the version of recharts been updated since your initial PR went in? If we revert back to the old version, does it fix the issue? Can we pinpoint the version where the bug appeared?
* If we try the other possible values for the `interval` option of XAxis, does that change anything?
* What if we format the label in another (simpler) way?
* If all else fails, can you come up with a minimal example _outside_ of our codebase? If so, we could post an issue in the recharts GH. 

LMK how this goes! I think it's worth spending a few hours on -- if we can't figure it out after that we can discuss what to do instead.

lilidworkin:
@terryttsai thanks for all of this! Super helpful. A few more questions:

> What if we format the label in another (simpler) way?

By this I meant, just as a test, not to actually ship, what if we just make the labels read something like "1, 2, 3, 4". Obviously we aren't going to ship that but that would help us isolate the issue, like if the problem is that our labels are too long, or something.

>  but the containers in which we put the chart in has been restyled.

Interesting, okay, that's a good lead. As a sanity check, is it possible/feasible to "revert" the styles back, and then see if the bug still appears? That will tell us for sure if that was the issue.

If it is the issue, can we do a "binary search" on the changes in an attempt to isolate exactly _which_ style change broke it?

If reverting the styles back _doesn't_ fix it, then I'm a bit stumped and will have to give it more thought ...

common/components/DataViz/utils.ts
mxosman: Oh nice! I was mostly wondering if there was a risk of the splitUtcString `month` returning something like `undefined` -- I like the solution of an error string but agree with you that it's highly unlikely you'll run into that case.

mxosman:
> OK so I finally figured out the issue! As soon as I got off this PR the issue was fixed, so the issue was in my PR the whole time, not sure why at first I thought I was still seeing the issue on an older branch.
> 
> The issue was that I was relying on the `index` param in the tickFormatter function, that index is the index of ticks to display, not the index of which data is being accessed.
> 
> The `value` param in the tickFormatter function is correct, it's the `start_date` of the datapoint because I was setting `start_date` as the key to be accessed in `dataKey`, but I need not just the `start_date` but also the `frequency` to know how to format the label.
> 
> Luckily, recharts allows you to pass in a function for the `dataKey`, it looks like the function has to return a string or number though, so I passed into `dataKey` a function that returns a stringified Datapoint so the tickFormatter can parse it as an object, then access the `start_date` and `frequency` to format the label.
> 
> Then I realized, wait, I can just do the formatting in the `dataKey` function itself so that's the solution I ended up with in my latest commit.

Wow - nice, debugging and sleuthing, Terry!

mxosman (approved):

lilidworkin:
Woah great debugging @terryttsai ! Nice work!

PR #406, opened 2023-02-13
Guidance: API

### LINKED ISSUE #408: [Frontend][Publisher] Guidance: API

mxosman:
## Description of the change

Implements the frontend API calls to retrieve and save progress throughout the Guidance flow.

Demo:

https://user-images.githubusercontent.com/59492998/218617077-18f05808-ecb5-45d5-a4db-0a5e12c9c360.mov

Note: it looks like I clicked on Skip the first time - but I refreshed twice on each screen after the welcome screen - my mouse just showed up at the perfect spot.

## Related issues

Closes #408

### Development

### Code review

### DIFF
--- a/publisher/src/stores/GuidanceStore.ts
+++ b/publisher/src/stores/GuidanceStore.ts
@@ -15,15 +15,15 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { showToast } from "@justice-counts/common/components/Toast";
 import { MetricConfigurationSettingsOptions } from "@justice-counts/common/types";
-import { makeAutoObservable, when } from "mobx";
+import { makeAutoObservable, runInAction, when } from "mobx";
 
 import {
   metricConfigurationProgressSteps,
-  mockTopicsStatus,
   OnboardingTopicsMetadata,
   onboardingTopicsMetadata,
-  OnboardingTopicsStatus,
+  OnboardingTopicsStatuses,
   ProgressSteps,
   TopicID,
 } from "../components/Guidance";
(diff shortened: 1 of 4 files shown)


publisher/src/stores/GuidanceStore.ts
```diff
@@ -73,27 +73,72 @@ class GuidanceStore {
...
+  ): Promise<Response> => {
+    const response = (await this.api.request({
+      path: `/api/agencies/${agencyId}/guidance`,
+      body: { updated_topic: updatedTopic },
+      method: "PUT",
+    })) as Response;
+
+    if (response.status !== 200) {
+      showToast(`Failed to update topic status.`, true, "red", 4000);
+      throw new Error("There was an issue updating the topic status.");
```
mxosman, line 127: It would be jarring to see this error - but incase it does error, what should be displayed to the user? Does "Failed to update topic status" make sense or is there a cleaner way to communicate that their progress didn't save?

Will need to think about this some more because right now, the store updates regardless, so they will proceed to the next step in the FE - but when it doesn't save, that will be a problem.

lilidworkin (approved):
Also looks good to me, I wouldn't worry too much about the failure case, do whatever is easiest.

publisher/src/stores/GuidanceStore.ts
mxosman: Yeah - it's definitely not consistent with how we handle it elsewhere where we update regardless and show a failure toast - but I feel like for this particular feature - it might not be best to show them step 3 when step 2 didn't fully clear the BE, for example.

As far as the awkwardness - I meant if a user is buried in the Metric Config, for example, and they check off the last thing needed to complete the entire topic, they won't see the toast failure until they hit the Get Started tab (which they will do anyway because we don't have a notion of redirecting back to that tab when Setting actions are completed). It feels like it's a bit strange - but maybe I'm overthinking it.

mxosman: Updated here: [65a95e7](https://github.com/Recidiviz/justice-counts/pull/406/commits/65a95e7aad68eb3dcb6ff2cb02272fd64e2963af)

Issue #329, opened 2023-01-22
[Frontend][Publisher] Guidance: Follow Ups (Adjustments/Improvements/Enhancements)


mxosman:
**Feedback Follow Ups**

**Data Store (#261)**
* What if mockTopicStatus was a map of topic ID to status, then we could update the map directly, this.onboardingTopicsStatus[topicID] = status, making this a O(1) operation instead of O(n)? (Lili) **(will be completed in #418)**

**Scaffolding (#262)**
* Add page not found if `isAgencyIdInUserAgencies` is false **(will be completed in #417)**
* Try to merge the GuidanceHeader with the regular Header instead of having separate Headers **(will be completed in #417)**

**Add Data (#266)**
* Merge the routing logic so there is no unnecessary duplication (currently there is routing specifically for Guidance and a separate but similar routing for non-guidance) **(will be completed in #417)**
* Have backend be responsible for determining guidance topic and eliminate the need to call `getReportOverviews` and add unnecessary round trips between FE/BE

**General UI**
* Add Logout, Learn More and Agencies dropdown to header **(will be completed in #417)**

**Playtesting Followups**
* Update copy in records page from “No records to display” to “No records to display. Create a new report?” or take the user to Create Report screen  **(will be completed in #419)**

PR #417, opened 2023-02-17
Guidance (Follow Ups): Routing & Header

mxosman:
## Description of the change

Addresses the following feedback from earlier PRs:

* Merge the routing logic so there is no unnecessary duplication (currently there is routing specifically for Guidance and a separate but similar routing for non-guidance)
* Add page not found if isAgencyIdInUserAgencies is false 
* Try to merge the GuidanceHeader with the regular Header instead of having separate Headers
* Add Logout, Learn More and Agencies dropdown to header

With Agencies menu:
Screenshot 2023-02-17 at 11 50 51 AM

With name and agency in menu:
Screenshot 2023-02-17 at 3 56 04 PM
Screenshot 2023-02-17 at 3 55 51 PM

## Related issues

Contributes to #329 

### Development

### Code review


publisher/src/components/Guidance/GuidanceMenu.tsx
```diff
@@ -0,0 +1,219 @@
...
+                onClick={() => isAddDataOrPublishDataStep && navigate("upload")}
+              >
+                Upload Data
+              </UploadDataButton>
+            </MenuItem>
+          </MenuContainer>
+        )}
+      </>
+    );
+  });
```
mxosman, line 219: I tried merging this with the `Menu` component, but I feel like they should be separate as to not pollute the `Menu` component with unnecessary guidance specific variables and effects.

publisher/src/components/Header/Header.tsx
```diff
@@ -15,40 +15,71 @@
...
+      return Promise.reject(
+        new Error(
+          "Something went wrong with clearing auth session or authStore is not initialized."
+        )
+      );
+    } catch (error) {
+      if (error instanceof Error) return error.message;
+      return String(error);
+    }
+  };
```
mxosman, line 58: Didn't touch this aside from moving it to a shared component as both Menu/GuidanceMenu use this.

publisher/src/components/Guidance/GuidanceMenu.tsx
```diff
@@ -0,0 +1,237 @@
...
+                href="https://justicecounts.csgjusticecenter.org/"
+                target="_blank"
+                rel="noreferrer"
+              >
+                Learn More
+              </a>
+            </MenuItem>
+
+            {/* Agencies Dropdown */}
+            {(userStore.permissions?.includes(Permission.RECIDIVIZ_ADMIN) ||
```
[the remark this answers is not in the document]
mxosman: Yeah - this entire feature branch is very out of sync with main. Once I rebase, it'll include everyone's changes. This was more/less to stabilize this branch and prevent it from crashing.

publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -231,7 +240,7 @@ export const Guidance = observer(() => {
                   kind="bordered"
                   onClick={() => navigate(`../${REPORTS_LOWERCASE}`)}
                 >
-                  Fill out ${REPORT_LOWERCASE}
+                  Fill out {REPORT_LOWERCASE}
```
lilidworkin, line 243: Was this an intentional change?

publisher/src/stores/GuidanceStore.ts
```diff
@@ -119,8 +119,7 @@ class GuidanceStore {
   };
 
   saveOnboardingTopicsStatuses = async (
-    updatedTopic: OnboardingTopicsStatuses,
-    agencyId: string
```
lilidworkin, line 123: Oh, ok, it just occurred to me, we're saving guidance progress per user, not per user-agency pair. So if a user is part of multiple agencies, and completes guidance for one of them, we will act as if they have completed guidance for all of them. That might be okay! But I'm not sure. Sorry that this didn't occur to me before! It's possible we'll have to move the guidance_progress column from the UserAccount table to the AgencyUserAccountAssocation table. Could you post in #justice-counts and ask @hobuobi  about this case? Thanks!

publisher/src/components/Guidance/GuidanceMenu.tsx
```diff
@@ -0,0 +1,239 @@
...
+              <a
+                href="https://justicecounts.csgjusticecenter.org/"
+                target="_blank"
+                rel="noreferrer"
+              >
+                Learn More
+              </a>
+            </MenuItem>
+
+            {/* Agencies Dropdown */}
```
lilidworkin, line 191: Is there any way to share _some_ of this logic with the main menu? I worry that now, if we want to make a change to the Agencies dropdown menu, we have to make it in two places. We don't have to share the whole Menu, but maybe we could share some of the MenuItems?

publisher/src/router/Router.tsx
```diff
@@ -42,95 +42,85 @@ export const Router = () => {
...
-            />
-            <Route
-              path={`/${REPORTS_LOWERCASE}/:id`}
-              element={<ReportDataEntry />}
-            />
-            <Route
-              path={`/${REPORTS_LOWERCASE}/:id/review`}
-              element={<ReviewReportDataEntry />}
-            />
+  return (
```
lilidworkin, line 49: Nice, I like how this is shared now! I know that meant pushing some of the guidance logic inline which might feel a little ugly but I have a feeling it'll be easier to debug and reason about when it's all in one place

publisher/src/stores/GuidanceStore.ts
lilidworkin: (Obviously can/should be done in separate PRs if we do decide to change this)

lilidworkin (commented):
Looks great! Couple questions inline -- I think the main one is whether we can share certain MenuItems -- but if you don't think that makes sense, lmk and I'll approve!

publisher/src/components/Guidance/Guidance.tsx
mxosman: Yes! It showed up as `Fill out $record` and I realized I'm not in a template string here in the render. 🤦

publisher/src/stores/GuidanceStore.ts
mxosman: Ah - that's such a great point! I didn't think of that either. I think it makes sense to me that they only need to go through the onboarding once. I can see the benefit of going through it for each agency - pushing the user to go through all of the configuration and adding data steps - but I'm imagining that might be exhausting for a user with 2+ agencies to have to keep going through the whole process when they'll know where everything is? I'm not sure now either - I'll ask in the channel!

publisher/src/components/Guidance/GuidanceMenu.tsx
mxosman: Great call! Absolutely - I'll think about this some more and find more opportunities!

publisher/src/router/Router.tsx
mxosman: I definitely prefer this now - it reads a lot better than the other one for me - thank you for suggesting this!!

publisher/src/components/Guidance/GuidanceMenu.tsx
lilidworkin: Great! I wouldn't spend too too much time on it, but see if there are some easy wins

mxosman: There weren't too many easy wins aside from extracting a few menu items into a component - which created an issue with the order of menu items and might be more confusing because you'd still have to make updates in two places (a portion in one component and the other in the shared parent component. So, I opted to try to merge both of them - here's a stable version of what that would look like: [ee69115](https://github.com/Recidiviz/justice-counts/pull/417/commits/ee691155d5bf799ab2a43c022f197a679604dd3f)

Would love to hear your thoughts - whether or not it's too confusing to consolidate or if this works. It does create some unnecessary calls to the `guidanceStore` every time the `Menu` re-renders, but I can short-circuit those variables and functions and keep them undefined if `hasCompletedOnboarding` is `true`.

publisher/src/stores/GuidanceStore.ts
mxosman: I created the following tasks to accomplish this: 
* [[Frontend][Publisher] Update Guidance API route to include agency ID](https://github.com/Recidiviz/justice-counts/issues/431)
* [[Backend][Justice Counts] Move `guidance_progress` to `agency_user_account_association` table to store progress based on agency id](https://github.com/Recidiviz/recidiviz-data/issues/19003)

lilidworkin: Perfect, thanks so much!

publisher/src/components/Guidance/Guidance.tsx
lilidworkin: Oh haha nice!

publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -242,10 +242,10 @@ export const Guidance = observer(() => {
                   onClick={() => {
                     if (currentTopicID) {
                       if (pathToTask) navigate(pathToTask);
-                      saveOnboardingTopicsStatuses(
-                        { topicID: currentTopicID, topicCompleted: true },
-                        agencyId
```
lilidworkin, line 247: We'll have to undo this change later, since we'll need to pass down agencyID to the API once we make the other changes, so maybe might a well keep it how it is how?

publisher/src/components/Menu/Menu.tsx
```diff
@@ -62,89 +113,182 @@ const Menu = () => {
...
+  useEffect(() => {
+    const initOnboardingTopicStatuses = async () => {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      await guidanceStore.getOnboardingTopicsStatuses();
+    };
+
+    initOnboardingTopicStatuses();
+  }, [guidanceStore, agencyId]);
+
+  if (hasCompletedOnboarding === false && !currentTopicID) return null;
```
lilidworkin, line 152: Is this a valid state to be in? Or should this never happen? Might be nice to add a comment

publisher/src/components/Menu/Menu.tsx
```diff
@@ -62,89 +113,182 @@ const Menu = () => {
...
-      </MenuItem>
-
-      <MenuItem buttonPadding>
-        <Button type="blue" onClick={() => navigate("upload")}>
-          Upload Data
-        </Button>
-      </MenuItem>
+          {/* Reports */}
+          {(hasCompletedOnboarding ||
+            hasCompletedOnboarding === null ||
```
lilidworkin, line 199: When would hasCompletedOnboarding be null?

publisher/src/components/Menu/Menu.tsx
```diff
@@ -62,89 +113,182 @@ const Menu = () => {
...
+          >
+            Settings
+          </MenuItem>
+
+          <MenuItem onClick={logout} highlight>
+            Log Out
+          </MenuItem>
+
+          <MenuItem buttonPadding>
+            {hasCompletedOnboarding ? (
```
lilidworkin, line 276: So the logic here is that no matter where you are in the guidance flow, the upload data button appears, but it is only activated if you're at the adddataorpublish step? What does activated mean?

lilidworkin (approved):
This looks really good @mxosman ! I actually do like how it looks combined, what do you think? I see what you mean about being a little "polluted" with the guidance logic but I really like how you can see, for each menu item, in one place, exactly when it will appear and when it won't. I have a feeling this will be easier to reason about and debug if we do run into weird behavior. I could be wrong, though -- let's see!

Left a few questions inline but nothing merge-blocking. Thanks for all your work on this, I know it was a lot!

publisher/src/components/Menu/Menu.tsx
mxosman: Ah - this is no longer necessary. I had it there because there was a brief milliseconds moment at the Welcome screen where menu items would show up while currentTopicID was being initialized before disappearing again.

publisher/src/components/Menu/Menu.tsx
mxosman: Great question - I changed `hasCompletedOnboarding` to return `null` when `Object.values(this.onboardingTopicsStatuses).length === 0` to differentiate this state a bit more because I was running into some issues when the guidance store wasn't being initialized (the component was unmounted because you're in Metric Config for example) - but this is no longer necessary after moving the initialization call to the Menu component! Thank you for catching this!

publisher/src/components/Menu/Menu.tsx
mxosman: Good question! So, there are two buttons - the standard generic button that we use throughout and one specific to guidance which controls the enabled ("activated")/disabled state of the Upload Button (when we're in Add Data/Publish Data step). Probably makes more sense to merge the logic here with the generic button. Always a battle within me of separating concerns vs. merging things together. Thanks for calling this out!

mxosman: Went with `enabledDuringOnboarding`

publisher/src/components/Menu/Menu.tsx
mxosman: Removed here: [23ad7d6](https://github.com/Recidiviz/justice-counts/pull/417/commits/23ad7d6c67422dffe3e3945c0caba25e540ffd01)

publisher/src/components/Menu/Menu.tsx
mxosman: Removed here: [23ad7d6](https://github.com/Recidiviz/justice-counts/pull/417/commits/23ad7d6c67422dffe3e3945c0caba25e540ffd01)

publisher/src/components/Guidance/Guidance.tsx
mxosman: Re-added here: [fe76941](https://github.com/Recidiviz/justice-counts/pull/417/commits/fe769419adeaeee00ab283687a103c84879635d3)

mxosman:
> This looks really good @mxosman ! I actually do like how it looks combined, what do you think? I see what you mean about being a little "polluted" with the guidance logic but I really like how you can see, for each menu item, in one place, exactly when it will appear and when it won't. I have a feeling this will be easier to reason about and debug if we do run into weird behavior. I could be wrong, though -- let's see!

Thank you, Lili! It'll take a minute for my eyes to adjust to it 😆 -- but I think this is **_definitely_** better than having to update items in multiples places - and agree that it'll be much easier to reason about and debug! Thank you for steering me in this direction!

publisher/src/components/Menu/Menu.tsx
lilidworkin: Makes sense to me! Thanks for explaining!

PR #418, opened 2023-02-17
Guidance (Follow Ups): Update `onboardingTopicsStatuses` to be a mapped object

mxosman:
## Description of the change

Updates `onboardingTopicsStatuses` in the store to be a mapped object vs. an array of objects.

## Related issues

Contributes to #329 

### Development

### Code review

### DIFF
--- a/publisher/src/stores/GuidanceStore.ts
+++ b/publisher/src/stores/GuidanceStore.ts
@@ -46,7 +46,7 @@ class GuidanceStore {
 
   onboardingTopicsMetadata: OnboardingTopicsMetadata;
 
-  onboardingTopicsStatuses: OnboardingTopicsStatuses[];
+  onboardingTopicsStatuses: { [topicID: string]: boolean };
 
   constructor(
     userStore: UserStore,
@@ -60,7 +60,7 @@ class GuidanceStore {
     this.metricConfigStore = metricConfigStore;
     this.isInitialized = false;
     this.onboardingTopicsMetadata = onboardingTopicsMetadata;
-    this.onboardingTopicsStatuses = [];
+    this.onboardingTopicsStatuses = {};
 
     when(
       () => metricConfigStore.isInitialized,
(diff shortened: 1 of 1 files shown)


publisher/src/stores/GuidanceStore.ts
```diff
@@ -69,20 +69,24 @@ class GuidanceStore {
...
     );
-    return indexOfTopicNotCompleted < 0;
   }
 
   get currentTopicID() {
-    if (this.onboardingTopicsStatuses.length === 0) return;
-    const topicIndex = this.onboardingTopicsStatuses.findIndex(
-      (topic) => !topic.topicCompleted
+    if (Object.values(this.onboardingTopicsStatuses).length === 0) return;
+    const topicEntry = Object.entries(this.onboardingTopicsStatuses).find(
```
lilidworkin, line 82: I think this may not always work as intended -- maps are unordered data structures, so I don't think this is guaranteed to return "first" uncompleted topic. Right?

lilidworkin (commented):
Looks good, just one concern inline!

publisher/src/stores/GuidanceStore.ts
mxosman: Great call! It should be in the order the original array was in - it goes through a reducer that creates the mapped object based on the order of the array. It shouldn't randomly change the order - but it is finiky in that if the order of the array sent from the backend is different, then it will create the mapped object based on that order.

Would you recommend adding an additional sorting step here?

lilidworkin: Oh ok, as long as you're sure the reducer preserves order, then it should be fine!

lilidworkin (approved):

publisher/src/stores/GuidanceStore.ts
mxosman: Here's a little snippet I used to test this (which can be pasted into a browser console):

```
let testArray = [{topicID: '1. A'}, {topicID: '2. B'}, {topicID: '3. C'},{topicID: '4. D'}]

let testArrayDiffOrder = [{topicID: '1. A'}, {topicID: '4. D'}, {topicID: '3. C'}, {topicID: '2. B'}]

let mappedObject = testArray.reduce((acc, topic) => {
    acc[topic.topicID] = topic;
    return acc;
}, {})

let mappedObjectDiffOrder = testArrayDiffOrder.reduce((acc, topic) => {
    acc[topic.topicID] = topic;
    return acc;
}, {})

console.log("testArray")
console.log(mappedObject)
console.log(Object.keys(mappedObject))
console.log('---')
console.log("testArrayDiffOrder")
console.log(mappedObject)
console.log(Object.keys(mappedObject))
```

lilidworkin: Well, just because it works in once instance does not mean it's guaranteed to always work that way on every browser/computer/etc. I do think it's probably fine, it's just something to be aware of. This [SO post](https://stackoverflow.com/questions/56230379/order-in-which-properties-are-added-to-array-reduce-return-object) is a good illustration of what I was concerned about.

mxosman: That's a great point. I think in the SO post - the trouble they were getting into was using "numbers" (they're strings but get treated as numbers if you only have numbers) for the keys, which JS will reorder:

Example: `{ "3": "2019-01-27", "2": "2019-01-20", "1": "2019-01-13" }` will end up being stored as `{1: "2019-01-13", 2: "2019-01-20", 3: "2019-01-27",}`

But, if they had non-numerical keys like: `{ "3a": "2019-01-27", "2b": "2019-01-20", "1c": "2019-01-13" }` it will be stored with the preserved order.

Regardless, I'm more than happy to adjust this logic to something that will explicitly preserve the order. I think as long as we're not using only numbers for the keys, we should be safe from the JS reordering.

lilidworkin: Ooh interesting! No that makes a lot of sense. I think this is fine!! Thanks for bearing with me.

> (they're strings but get treated as numbers if you only have numbers)

lol oh JS

mxosman: Sounds great! Absolutely - this is **_totally_** worth calling out because JS can be super sus' with its rules.

> > (they're strings but get treated as numbers if you only have numbers)
> 
> lol oh JS

😆  Exactly!

PR #428, opened 2023-02-21
[Fix] Disaggregation and dimensions out of sync with the backend

### LINKED ISSUE #420: [Frontend/Backend][Bug] Disaggregations appear enabled by default without any dimensions enabled
Example: https://justice-counts-web-qqec6jbn6a-uc.a.run.app/agency/161/settings/metric-config?system=jails&metric=jails_total_staff

mxosman:
## Description of the change

The disaggregation toggles appear to always default to on (checked) and the dimensions off (unchecked) for every breakdown and saving doesn't appear to persist on the FE. The issue was in the initialization of the metric config store dimensions - there was no `enabled` property included in the store's dimensions object for the FE to read from. The fix is adding back the storage of `.enabled` property in the `initializeDimension` function.

Demo of out-of-sync behavior:

https://user-images.githubusercontent.com/59492998/220479226-1e9ab8de-ba7a-4308-84bd-e8a6d201cbcc.mov

--

Demo of correction:

https://user-images.githubusercontent.com/59492998/220479215-5173d92e-a011-445a-8b9f-da6764f82ef8.mov

## Related issues

Closes #420 

### Development

### Code review

### DIFF
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -418,6 +418,8 @@ class MetricConfigStore {
       this.dimensions[systemMetricKey][disaggregationKey][dimensionKey] = {};
     }
 
+    this.dimensions[systemMetricKey][disaggregationKey][dimensionKey].enabled =
+      dimensionData.enabled;
     this.dimensions[systemMetricKey][disaggregationKey][dimensionKey].label =
       dimensionData.label;
     this.dimensions[systemMetricKey][disaggregationKey][dimensionKey].key =
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -20,6 +20,7 @@ import {
   AgencySystems,
   FormError,
   Metric,
+  MetricConfigurationSettings,
   MetricConfigurationSettingsOptions,
   MetricContext,
   MetricDisaggregationDimensions,
@@ -56,11 +57,7 @@ class MetricConfigStore {
 
   metricDefinitionSettings: {
     [systemMetricKey: string]: {
-      [settingKey: string]: {
-        included?: MetricConfigurationSettingsOptions;
-        default?: MetricConfigurationSettingsOptions;
-        label?: string;
-      };
+      [settingKey: string]: Partial<MetricConfigurationSettings>;
     };
   };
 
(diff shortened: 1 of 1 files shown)


lilidworkin (approved):
Oh wow, crazy that this was only a one-line fix! Just out of curiosity, this did used to work, right? I wonder what happened? 

And you also verified that the right metrics show as enabled/disabled on the records page?

Thanks for the quick fix!!

mxosman:
> Oh wow, crazy that this was only a one-line fix! Just out of curiosity, this did used to work, right? I wonder what happened?

I know - right? They _always_ take such a disproportionate amount of time to debug. Yes - it definitely worked normally before, but I think during the store refactor (moving the initialization logic to their own function) - this line was overlooked. I'm going to comb through the store again to see if there's any other metadata that is missing.

> And you also verified that the right metrics show as enabled/disabled on the records page?

Yes - I tested several metrics and they show up enabled/disabled based on their configuration.

lilidworkin:
> I'm going to comb through the store again to see if there's any other metadata that is missing.

Cool, thanks! Don't spend too much time on it, we'll catch other issues during testing.

> Yes - I tested several metrics and they show up enabled/disabled based on their configuration.

Perfect!

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -358,7 +354,7 @@ class MetricConfigStore {
     system: AgencySystems,
     metricKey: string,
     settingKey: string,
-    metricDefinitionSettings: { [key: string]: string }
+    metricDefinitionSettings: Partial<MetricConfigurationSettings>
```
mxosman, line 357: Made this more specific than `{ [key: string]: string }`

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -434,6 +432,42 @@ class MetricConfigStore {
...
+      this.dimensionDefinitionSettings[systemMetricKey][disaggregationKey][
+        dimensionKey
+      ] = {};
+    }
+
+    this.dimensionDefinitionSettings[systemMetricKey][disaggregationKey][
+      dimensionKey
+    ][settingKey] = dimensionDefinitionSettings;
+  };
+
```
mxosman, line 470: This block was sandwiched between the updater functions - moved it closer to the other initialization functions. Only thing I updated was the type of the `dimensionDefinitionSettings` param.

PR #463, opened 2023-03-03
Remove empty scrollbars from all repos

mxosman:
## Description of the change

Follow up on Ivan's work - extending the replacement of `overflow: scroll` to `auto` across all instances in the different apps.

## Related issues

Contributes to #441

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/DashboardModals/DashboardModal.styles.tsx
+++ b/agency-dashboard/src/DashboardModals/DashboardModal.styles.tsx
@@ -32,7 +32,7 @@ export const ModalContainer = styled.div`
   display: flex;
   align-items: left;
   justify-content: center;
-  overflow-y: scroll;
+  overflow-y: auto;
   z-index: 3;
 `;
 
--- a/agency-dashboard/src/Header/AboutModal.styles.tsx
+++ b/agency-dashboard/src/Header/AboutModal.styles.tsx
@@ -32,7 +32,7 @@ export const AboutModalContainer = styled.div`
   display: flex;
   align-items: left;
   justify-content: center;
-  overflow-y: scroll;
+  overflow-y: auto;
 `;
 
 export const AboutModalInnerContainer = styled.div`
@@ -115,7 +115,7 @@ export const AboutModalInfoButton = styled.div`
   font-weight: 400;
   border: 1px solid ${palette.highlight.white1};
   border-radius: 99px;
-  padding: 8px; 16px;
+  padding: 8px 16px;
   float: left;
   margin-right: 2px;
   margin-top: 2px;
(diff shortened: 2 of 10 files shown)


morden35:
Hey Mahmoud! I ran this locally to check it out. Did it change the guidance page by chance (or did it always look like this?). In particular, it looks like the Guidance part of the page has been 'pushed up' closer to the headers. What do you think?

Screenshot 2023-03-03 at 3 45 19 PM

morden35:
But it does look like the scroll bars we were seeing before are gone so that's great!

Screenshot 2023-03-03 at 3 49 10 PM

morden35:
However I do have a larger (Google Chrome) scroll bar on this entire page, even though there is no where to scroll.
Screenshot 2023-03-03 at 3 50 33 PM

mxosman:
> Hey Mahmoud! I ran this locally to check it out. Did it change the guidance page by chance (or did it always look like this?). In particular, it looks like the Guidance part of the page has been 'pushed up' closer to the headers. What do you think?
> 
> Screenshot 2023-03-03 at 3 45 19 PM

I noticed that everything was pushed up too when I was working off of main today... I'll look into that!

If you pull the latest main on the frontend - those scrollbars in this screenshot should be gone entirely as I removed the container itself.

mxosman:
> However I do have a larger (Google Chrome) scroll bar on this entire page, even though there is no where to scroll. Screenshot 2023-03-03 at 3 50 33 PM

I think this one makes sense to me because the page underneath of the modal is scrollable and I believe it's referencing that. When the modal is open, all we do is prevent scrolling on the body of the page so you don't have a situation where the modal is open and you can still scroll on everything underneath it - if that makes sense.

morden35:
> I think this one makes sense to me because the page underneath of the modal is scrollable and I believe it's referencing that. When the modal is open, all we do is prevent scrolling on the body of the page so you don't have a situation where the modal is open and you can still scroll on everything underneath it - if that makes sense.

Got it, makes sense!

mxosman:
> > Hey Mahmoud! I ran this locally to check it out. Did it change the guidance page by chance (or did it always look like this?). In particular, it looks like the Guidance part of the page has been 'pushed up' closer to the headers. What do you think?
> 
> I noticed that everything was pushed up too when I was working off of main today... I'll look into that!
> 

OK! I think I figured out what was going on - there was a `height` property clash in the mergers this week - fixed it here with this commit https://github.com/Recidiviz/justice-counts/pull/465/commits/26078121d99efd93e998c4309e7053e6bd132347

lilidworkin (approved):
Awesome! Thanks so much @mxosman for going out of your way to catch some of these missing ones! And thanks to @morden35 for helping to test!

PR #465, opened 2023-03-03
Guidance: Update Metric Configuration overview list for multi-system agencies & agencies with subsystems

### LINKED ISSUE #466: [Frontend][Publisher] Guidance: Handle multi-system and subsystem agency metrics in Configure Metrics step

mxosman:
## Description of the change

In the Configure Metrics step of the guidance flow, the agencies that have multiple systems will now see the overview list split by system. For Supervision agencies with subsystems - this removes those metrics from the view (those metrics are marked as unavailable so removing them from view should be OK as they will not count towards progress/block progressing from this step).

Demo:

https://user-images.githubusercontent.com/59492998/222842285-cbb79f85-19e1-409e-9f7c-94ecfafdcae3.mov

## Related issues

Closes #466 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Guidance/Guidance.tsx
+++ b/publisher/src/components/Guidance/Guidance.tsx
@@ -122,6 +122,7 @@ export const Guidance = observer(() => {
   );
   const totalMetrics = Object.values(metricsEntriesBySystem).reduce(
     (acc, metrics) => {
+      // eslint-disable-next-line no-param-reassign
       acc += metrics.length;
       return acc;
     },
--- a/publisher/src/components/Guidance/Guidance.tsx
+++ b/publisher/src/components/Guidance/Guidance.tsx
@@ -120,16 +120,13 @@ export const Guidance = observer(() => {
     },
     {} as { [system: string]: [string, MetricInfo][] }
   );
-
-  console.log(JSON.stringify(metricsEntriesBySystem, null, 2));
-  // console.log(
-  //   JSON.stringify(
-  //     metricConfigStore.getMetricsBySystem(currentAgency?.systems[0]),
-  //     null,
-  //     2
-  //   )
-  // );
-  const totalMetrics = metricsEntries.length;
+  const totalMetrics = Object.values(metricsEntriesBySystem).reduce(
+    (acc, metrics) => {
+      acc += metrics.length;
+      return acc;
+    },
+    0
+  );
 
   const renderProgressSteps = () => {
     if (currentTopicID === "WELCOME") return;
@@ -158,9 +155,9 @@ export const Guidance = observer(() => {
     );
   };
 
-  const numberOfMetricsCompleted = metricsEntries.filter(
-    ([key]) => getMetricCompletionValue(key) === 4
-  ).length;
+  const numberOfMetricsCompleted = Object.values(metricsEntriesBySystem)
+    .flatMap((entries) => entries)
+    .filter(([key]) => getMetricCompletionValue(key) === 4).length;
 
   useEffect(
     () => metricConfigStore.resetStore(),
--- a/publisher/src/components/Guidance/Guidance.tsx
+++ b/publisher/src/components/Guidance/Guidance.tsx
@@ -110,6 +110,7 @@ export const Guidance = observer(() => {
       const { system } =
         MetricConfigStore.splitSystemMetricKey(systemMetricKey);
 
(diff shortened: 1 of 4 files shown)


lilidworkin (commented):
@mxosman , so cool!

One question though, and you'll probably hate me for asking this -- what happens if a supervision agency chooses to disaggregate one of their metrics by subsystem? Then this comment will no longer be true:

> those metrics are marked as unavailable so removing them from view should be OK as they will not count towards progress/block progressing from this step).

mxosman:
> @mxosman , so cool! _Very_ clever solution to a problem that we didn't have designs for!
> 
> One question though, and you'll probably hate me for asking this -- what happens if a supervision agency chooses to disaggregate one of their metrics by subsystem? Then this comment will no longer be true:
> 
> > those metrics are marked as unavailable so removing them from view should be OK as they will not count towards progress/block progressing from this step).

Haha - that's such a darn good question and call out... For some reason I mistakenly assumed that they wouldn't do that initially - but that's not true at all! Hmm... I'll need to think about this some more.

On a new agency's initial view of the Metric Config step - everything is off and nothing is disaggregated. If they decide during this step to disaggregate it... I'm wondering what they should see when they go back to the overview list. Do we rework the logic to hide the original metric and only show the disaggregated one? This is trickier than I thought! Let me play around with the current behavior and see if there is a relatively easy win that makes sense.

mxosman:
@lilidworkin - what do you think about something like this?

https://user-images.githubusercontent.com/59492998/223219653-e0f7ea68-f1dd-44b6-8dfe-5bf983189d6e.mov

mxosman:
> Wait @mxosman this is amazing / perfect!! What wizardry did you do?? (Will actually review the PR later today)

All I did was allow a subsystem to be a part of the list if it is enabled (which happens when you disaggregate it). Will need to do a more thorough test.

I'm realizing there's one situation that might be a bit wonky... If you, for example, disaggregate the Funding metric (e.g. now it's in Parole and Probation) and then disable the metric in both Parole & Probation -- the Parole and Probation metrics disappear from the overview list and you're left with Funding (disabled) under Supervision still marked as disaggregated. I guess you can just hit 'Combined' to unblock access to 'Not Available', 'Monthly', 'Annual' buttons - but does that feel strange or does it make sense?

lilidworkin:
I think that makes sense and is totally 100% completely fine!! haha

publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -308,63 +346,89 @@ export const Guidance = observer(() => {
...
+                                    navigate(
+                                      `../settings/metric-config?system=${system}&metric=${metricKey}`
+                                    )
+                                  }
+                                >
+                                  <MetricName>{metric.label}</MetricName>
+                                  <MetricStatus
+                                    greyText={metric.enabled === false}
+                                  >
+                                    {metricCompletionProgressValue === 4 &&
```
lilidworkin, line 383: This is a nit, but we use this "magic number" 4 in a variety of places, I might suggest making it a constant somewhere

publisher/src/components/Guidance/Guidance.tsx
```diff
@@ -93,8 +101,38 @@ export const Guidance = observer(() => {
...
+  const currentAgency = userStore.getAgency(agencyId);
   const metricsEntries = Object.entries(metricConfigStore.metrics);
-  const totalMetrics = metricsEntries.length;
+  const metricsEntriesBySystem = metricsEntries.reduce(
+    (acc, [systemMetricKey, metric]) => {
+      const { system } =
+        MetricConfigStore.splitSystemMetricKey(systemMetricKey);
+
+      if (
+        !SupervisionSubsystems.includes(system as AgencySystems) ||
```
lilidworkin, line 113: Nice! Let's add a comment explaining this logic so we don't forget why we did it

lilidworkin (approved):
Awesome!

publisher/src/components/Guidance/Guidance.tsx
mxosman: Good call - will do!

publisher/src/components/Guidance/Guidance.tsx
mxosman: Good call!

mxosman:
> I think that makes sense and is totally 100% completely fine!! haha

Oh good - thank you for the sanity check!

Issue #296, opened 2023-01-12
"Annual" --> "Annually"


lilidworkin:
Matt from CSG requested the following tiny copy change of "Annual" to "Annually":
![Screenshot 2023-01-11 at 8.09.22 PM.png](https://images.zenhubusercontent.com/620e7d8c0c5fad82d897844d/6b0948db-e5b8-41ae-97ed-a5f615d6fdaf)

lilidworkin:
cc @hobuobi who wanted to see the process for making a copy change like this! @mxosman can you cc him in the PR you put up?

mxosman:
Sounds great! Will do! Quick question - we just want the UI label to change, right? We use the value "ANNUAL" throughout the FE and BE when doing comparisons (e.g. frequency === "ANNUAL"). I think we can just change the label and keep everything the same?

hobuobi:
Oh yeah, nothing should need to change on the BE! 
I also don't think we should change the rectangular "Annual" labels that appear in other parts of the app; the only reason why I agreed to change it here was because it makes grammatical sense as a response to the question ("Yes, I can share this metric **annually**") vs describing a metric ("**monthly** admissions," "**annual** budget").

mxosman:
Awesome! Thank you, Humphrey! That makes perfect sense! 🙏

PR #485, opened 2023-03-08
Metric Configuration Overview: show "Annually" instead of "Annual

mxosman:
## Description of the change

Updates Metric Config overview metric boxes to show "Annually" instead of "Annual".

Before:
image

After
Screenshot 2023-03-08 at 10 52 20 AM

## Related issues

Contributes to #296 

### Development

### Code review


mxosman:
This is simple enough to do on the frontend. I'm not sure what your thoughts are on updating the backend for this - but, I decided against it because it would be more work than its worth if we're only displaying it in this one spot (was going to say two - but we're removing the badges from inside the metric config).

lilidworkin (approved):

lilidworkin:
Yep, I think a frontend fix is fine/correct here!

mxosman:
Thank you, Lili!

PR #497, opened 2023-03-13
Guidance: Gate feature for non-JC Admins

### LINKED ISSUE #495: "Unlaunch" Guidance except for JC Admins
Ignore DB for non-JC Admins -- just assume Guidance is completed. For JC Admins, use whatever DB says.

Two aspects of Guidance experience: 1) the progression steps at the beginning, and 2) the redesigned Metric Config page which uses null, etc.

Note that "unlaunching" as above will only affect #1, not #2.

*Open Questions (waiting on Humphrey/CSG):*
  * For our existing users/agencies, what is the current behavior? What is the right behavior? For #1, we know the current behavior -- we manually set all of these folks to be completed, so they will skip all steps. (We may want to change this, depending on what Humphrey/CSG says, but wait on it for now). For #2, not yet entirely clear how existing stuff in the DB interacts with new design. _Possibly something to investigate._
   * When we are onboarding new agencies, do we want to be able to show them Records / Bulk Upload before they've finished configuring metrics? If so, how? Maybe we just always show those menu items to JC Admins so we can take over screen sharing at that point?
   * What should the criteria be for progression? Do we want to adjust this / make it more flexible?

mxosman:
## Description of the change

Gates the guidance feature for all users who are **not** Justice Counts Admins. This is accomplished by setting all non-JC Admin's guidance progress to completed for all topics. JC Admins will have their guidance progress read from the DB as usual.

## Related issues

Contributes to #495

### Development

### Code review

### DIFF
--- a/publisher/src/stores/GuidanceStore.ts
+++ b/publisher/src/stores/GuidanceStore.ts
@@ -117,7 +117,7 @@ class GuidanceStore {
            * If you are not a JC Admin, your guidance progress is marked as completed for all topics.
            * If you are a JC Admin, your guidance progress is based on your actual guidance progress from the DB.
            *
-           * TODO(#) ungate guidance flow for all users
+           * TODO(#496) ungate guidance flow for all users
            */
           acc[topic.topicID] = !this.userStore.isJusticeCountsAdmin(agencyId)
             ? true
--- a/publisher/src/stores/GuidanceStore.ts
+++ b/publisher/src/stores/GuidanceStore.ts
@@ -111,7 +111,17 @@ class GuidanceStore {
     runInAction(() => {
       this.onboardingTopicsStatuses =
         onboardingTopicsStatuses.guidance_progress.reduce((acc, topic) => {
-          acc[topic.topicID] = topic.topicCompleted;
+          /**
+           * NOTE:
+           * This gates the guidance flow for users who are not JC Admins.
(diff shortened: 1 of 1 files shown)


lilidworkin:
Ahh @mxosman yes, this is actually intentional even though it's super confusing. I wanted everyone to always be JC admin during local development so I made this change: https://github.com/Recidiviz/recidiviz-data/blob/d8876f19547ebb473170907586a725be51d87df0/recidiviz/justice_counts/control_panel/routes/api.py#L406

That's the reason for the weird behavior!! Sorry about that!

lilidworkin (approved):
Looks right to me!! I don't think we need to worry about testing locally, we can test when this gets pushed up to staging tomorrow as part of @nichelle-hall 's oncall deploy :D

mxosman:
> Ahh @mxosman yes, this is actually intentional even though it's super confusing. I wanted everyone to always be JC admin during local development so I made this change: https://github.com/Recidiviz/recidiviz-data/blob/d8876f19547ebb473170907586a725be51d87df0/recidiviz/justice_counts/control_panel/routes/api.py#L406
> 
> That's the reason for the weird behavior!! Sorry about that!

Oh that's right! I didn't think it would keep overriding - no worries at all! I was able to technically test via inversing the conditional - I think this will work just fine!

Thank you, Lili!

PR #501, opened 2023-03-14
Bulk Publishing: API

### LINKED ISSUE #499: [Frontend][Publisher] Bulk Publishing: Create FE API call functionality for Publishing and Unpublishing

mxosman:
## Description of the change

Implements the API calls to the new endpoint to update multiple reports' status to "PUBLISHED" or "DRAFT". 

Demo:

https://user-images.githubusercontent.com/59492998/225148976-f68c74a8-7fa6-4ffb-9768-e04fbf967012.mov

## Related issues

Closes #499

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/Reports.styles.tsx
+++ b/publisher/src/components/Reports/Reports.styles.tsx
@@ -129,8 +129,8 @@ export const ReportActions = styled.div`
 
 export const ReportActionsButton = styled.div<{
   disabled?: boolean;
-  textColor?: "blue";
-  buttonColor?: "red";
+  textColor?: "blue" | "white";
+  buttonColor?: "red" | "orange" | "green";
 }>`
   position: relative;
   border: ${({ buttonColor }) =>
@@ -147,11 +147,14 @@ export const ReportActionsButton = styled.div<{
 
     if (textColor === "blue") return palette.solid.blue;
 
-    if (buttonColor === "red") return palette.solid.white;
+    if (buttonColor && ["red", "orange", "green"].includes(buttonColor))
+      return palette.solid.white;
     return palette.solid.darkgrey;
   }};
   background-color: ${({ buttonColor }) => {
     if (buttonColor === "red") return palette.solid.red;
+    if (buttonColor === "green") return palette.solid.green;
+    if (buttonColor === "orange") return palette.solid.orange;
     return palette.solid.white;
   }};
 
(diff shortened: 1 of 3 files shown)


publisher/src/stores/ReportStore.ts
```diff
@@ -199,6 +199,41 @@ class ReportStore {
...
+    currentAgencyId: string,
+    status: ReportStatus
+  ): Promise<Response | Error | undefined> {
+    try {
+      const response = (await this.api.request({
+        path: `/api/reports`,
+        body: {
+          status,
+          // time_loaded:
+          //   this.reportOverviews[reportID].last_modified_at_timestamp,
```
mxosman, line 213: I'm not 100% sure what to do with this here or if it is necessary. It would originally take in the last modified timestamp for an individual report - now that we have multiple reports - I'm not sure which one to reference or if it should get a timestamp of now, since we're modifying the status?

publisher/src/components/Reports/Reports.styles.tsx
```diff
@@ -147,11 +147,14 @@ export const ReportActionsButton = styled.div<{
 
     if (textColor === "blue") return palette.solid.blue;
 
-    if (buttonColor === "red") return palette.solid.white;
+    if (buttonColor && ["red", "orange", "green"].includes(buttonColor))
+      return palette.solid.white;
```
ivan-kishko, line 151: I guess if button not disabled and it has backgroundColor we can write smth like that:
<code>if (disabled) return palette.highlight.grey8; return buttonColor ? white : grey</code>

publisher/src/pages/Reports.tsx
```diff
@@ -159,6 +159,18 @@ const Reports: React.FC = () => {
     }
   };
 
+  const publishOrUnpublishMultipleReports = async (
```
ivan-kishko, line 162: Just question: I will move that function to review multiple reports page later right? buttons on records page should just navigate you to a review page?

mxosman: You're exactly right! Right now, this is mostly to test the backend point - but eventually, this will not be called on the Records page - it'll be called from the multiple reports review page as you mentioned as well as the the review page after a user uploads a spreadsheet.

publisher/src/components/Reports/Reports.styles.tsx
mxosman: Nice! Great call & refinement!

lilidworkin (approved):
Looks great!

publisher/src/stores/ReportStore.ts
mxosman: Thank you for confirming! Done & done!

PR #506, opened 2023-03-15
Bulk Publishing: Create API call to retrieve multiple reports

### LINKED ISSUE #505: [Frontend][Publisher] Create API call to retrieve multiple reports

mxosman:
## Description of the change

Implements `getMultipleReports` to call a new endpoint (provided with a list of report `id`s) to retrieve a list of reports. This will be used in the Bulk Publishing feature to allow us to access and render multiple reports by report `id`s in the Publish review pages.

## Related issues

Closes #505 

### Development

### Code review

### DIFF
--- a/publisher/src/stores/ReportStore.ts
+++ b/publisher/src/stores/ReportStore.ts
@@ -153,6 +153,48 @@ class ReportStore {
     }
   }
 
+  async getMultipleReports(
+    reportIDs: number[],
+    currentAgencyId: string
+  ): Promise<void | Error> {
+    try {
+      const response = (await this.api.request({
+        path: `/api/reports/multiple`,
+        body: { report_ids: reportIDs, agency_id: parseInt(currentAgencyId) },
+        method: "POST",
+      })) as Response;
+
+      if (response.status !== 200) {
+        throw new Error(
+          `There was an issue retrieving these ${REPORTS_LOWERCASE}.`
+        );
+      }
+
+      const reports = (await response.json()) as Report[];
+
+      runInAction(() => {
+        reports.forEach((report) => {
+          const { metrics, ...overview } = report;
+          this.reportOverviews[report.id] = overview;
+          const metricsBySystem = groupBy(
+            metrics,
+            (metric) => metric.system.key
+          );
+          this.reportMetricsBySystem[report.id] = metricsBySystem;
+          // ensure that the order of the metrics in reportMetricsBySystem
+          // matches the order of the metrics in reportMetrics
+          this.reportMetrics[report.id] = Object.values(metricsBySystem).flat();
+        });
+      });
+    } catch (error) {
+      if (error instanceof Error) return new Error(error.message);
+    } finally {
+      runInAction(() => {
+        this.loadingReportData = false;
+      });
+    }
+  }
+
   async createReport(
     body: Record<string, unknown>,
     agencyId: number


publisher/src/stores/ReportStore.ts
```diff
@@ -153,6 +153,48 @@ class ReportStore {
...
 
+  async getMultipleReports(
+    reportIDs: number[],
+    currentAgencyId: string
+  ): Promise<void | Error> {
+    try {
+      const response = (await this.api.request({
+        path: `/api/reports/multiple`,
+        body: { report_ids: reportIDs, agency_id: parseInt(currentAgencyId) },
+        method: "POST",
```
mxosman, line 164: So, I'm a bit stumped with this one... It's technically a GET operation because we are getting a list of reports - but, we have to send a list of report ids that we wish to retrieve. It's not recommended to add a body to a GET request, so I'm left with POST, PUT, and PATCH - and none of those feel right because we're not updating anything...

One option (which is not pretty) is to make it a GET request and send all of the report IDs in the path and parse it in the backend? So the path would be something like `/api/${agencyId}/reports?ids=${id1}&ids=${id2}&ids=${id3}`

Need to research this a bit more but open to thoughts!

publisher/src/stores/ReportStore.ts
```diff
@@ -153,6 +153,48 @@ class ReportStore {
     }
   }
 
+  async getMultipleReports(
+    reportIDs: number[],
+    currentAgencyId: string
+  ): Promise<void | Error> {
+    try {
+      const response = (await this.api.request({
+        path: `/api/reports/multiple`,
```
mxosman, line 162: Still on the fence about this one - based on the feedback re: the HTTP method - I'll adjust this path!

nichelle-hall: My suggestion is that we update the existing reports method to get expect a list of `report_ids` in the request body instead of one `report_id`. When we are fetching one report, we just send a list with one report_id. That said, the response will be a list of reports instead of one singular report so you will have to update the frontend code to get the first element in the list when fetching a single report.

nichelle-hall: Wait, so is this getting all reports by report_id or just receive a subset of reports of an agency?

morden35: Can you use the existing get_reports_by_agency_id() endpoint?

mxosman: Damn - both are really great ideas!! I don't know why I didn't think to use the `get_reports_by_agency_id`!

In that case - we already have a call to that route in `getReportOverviews`. Let me see if it makes sense to marry the two functionalities (getReportOverviews + getMultipleReports into one) or if it makes more sense to create a separate function to call the same route and just provide you with the reports you requested (leaning on the latter for better separation of concerns). Love this! That means the backend API for this is no longer necessary.

mxosman: Thank you both!

mxosman: > Wait, so is this getting all reports by report_id or just receive a subset of reports of an agency?

Just a subset of reports requested. FE would request the following reports by giving a list of IDs (e.g. `[1,2,3]`) and the BE will return a list of all 3 reports (e.g. `[{id: 1, metrics: [...], etc: ...}, {id: 2, metrics: [...], etc: ...}, {id: 3, metrics: [...], etc: ...}]`)

mxosman: Hmm... the `get_reports_by_agency_id` only returns the report overviews without the metrics - and it would be nice to have both in one object so the values for each report's metrics are accessible in one place. I could try to update that method (`get_reports_by_agency_id`) to include the metrics with each report if that makes sense to do?

lilidworkin: This is a great conversation! Let me try to summarize:

* get_reports_by_agency_id gets 1) all reports for an agency, and 2) only returns report metadata, not report metrics
* get_report_by_id gets 1) a single report, and 2) returns report metadata AND report metrics

We want a method that 1) gets multiple reports by ID, and 2) returns report metadata AND report metrics

We could modify either of the above endpoints to do that, or create a new one. 

If we go the modify route, I think it makes more sense to modify the second one (@nichelle-hall 's original idea, I think). E.g. change

`api_blueprint.route("/reports/<report_id>", methods=["GET"])`

to 

`api_blueprint.route("/reports", methods=["GET"])`

but have it accept a parameter `report_ids` that can either be a single ID, or a list of IDs.

With regards to how to have a GET method that accepts a list of IDS -- this is a great question -- maybe check out [this SO post](https://stackoverflow.com/questions/4541338/how-to-construct-a-rest-api-that-takes-an-array-of-ids-for-the-resources) or [this one](https://stackoverflow.com/questions/9371195/rest-api-requesting-multiple-resources-in-a-single-get)? Looks like a comma separated list of IDs is the accepted standard. I imagine we JSON encode the list on the FE and then decode it on the BE?

publisher/src/stores/ReportStore.ts
lilidworkin: Great question -- answered above!!

publisher/src/stores/ReportStore.ts
```diff
@@ -153,6 +153,48 @@ class ReportStore {
...
+        throw new Error(
+          `There was an issue retrieving these ${REPORTS_LOWERCASE}.`
+        );
+      }
+
+      const reports = (await response.json()) as Report[];
+
+      runInAction(() => {
+        reports.forEach((report) => {
+          const { metrics, ...overview } = report;
```
lilidworkin, line 177: Possible to re-use this logic? I think the block below is repeated from getting a single report?

lilidworkin (commented):
This is all looking great! Amazing work on BE and FE @mxosman !

publisher/src/stores/ReportStore.ts
mxosman: Great summary! Spot on! I keep going back and forth with adding a new route or modifying an existing one. Now, I'd like to try going down the new endpoint and if it feels way too redundant, I can pivot into updating the existing route.

I ran across those SO posts too - nice find! I'll do the comma separated list and see how that plays.

Thank you all for the helpful feedback and discussion here!

publisher/src/stores/ReportStore.ts
mxosman: Absolutely! I was eyeing this loop for reference because we'll need to do the grouping by system as well and all the other good stuff.

mxosman: Oh sorry, you mean consolidate this logic with the one from getting a single report? If so - let me know what you think of the update to the path and how I'm sending over the `agency_id` and `report_ids` in the URL ([d5cc754](https://github.com/Recidiviz/justice-counts/pull/506/commits/d5cc75453ea6615725b13a9c58195df671e70df0)). If that's OK - I can merge both functions and update the components that get a single report. 

In which case - it makes more sense to consolidate the backend endpoint as Nichelle suggested (re: `/reports/<report_id>` into `/reports`), right? Want to make sure this way of making the GET request with `agency_id` and `report_ids` is OK before starting to consolidate the backend endpoint. 

And... if we do consolidate the endpoints - I think this would mean that I need to re-add the `editor_id_to_json` which will make multiple DB calls... Hmm. What do you think is the best way to go about this? Is it OK to keep them as separate endpoints or does that not feel right?

lilidworkin: Oh sorry for not being clear! I literally just meant this block of code:

```
this.reportOverviews[reportID] = overview;
const metricsBySystem = groupBy(metrics, (metric) => metric.system.key);
this.reportMetricsBySystem[reportID] = metricsBySystem;
// ensure that the order of the metrics in reportMetricsBySystem
// matches the order of the metrics in reportMetrics
this.reportMetrics[reportID] = Object.values(metricsBySystem).flat();
```

We repeat this verbatim (I think) in `getReport` and `getMultipleReports` (in the latter we call it in a loop). Can we put it in a helper function to re-use the code?

lilidworkin (commented):
Looks perfect! I think your solution / approach with the endpoint is exactly the right choise. I clarified what I meant about the code re-use inline, lmk if that helps clarify and if what I suggested is even feasible!

publisher/src/stores/ReportStore.ts
mxosman: Ah! Thank you for clarifying - pardon my tangent! That makes sense! It can definitely be put into a helper function in the store! Great call!

publisher/src/stores/ReportStore.ts
```diff
@@ -135,14 +150,38 @@ class ReportStore {
 
       const report = (await response.json()) as Report;
       const { metrics, ...overview } = report;
-
+      this.storeMetricDetails(reportID, metrics, overview);
```
lilidworkin, line 153: Yay thanks for moving to this helper function!

Out of curiosity is it a problem that this is now moved up into the body of the `try` statement, rather than in the `finally` clause where it was before? My instinct is that it shouldn't actually matter but just wanted to check.

Issue #24, opened 2022-09-20
[Frontend] Add Publish button to publish multiple reports at once


terryttsai:
After https://github.com/Recidiviz/recidiviz-data/issues/15443 is done, implement publishing multiple reports on the front end

terryttsai:
Wouldn't publishing those metrics also publish whatever is on those reports?

terryttsai:
A user wouldn't confirm that's what they want to upload in the current review metrics page in data upload

PR #507, opened 2023-03-16
Review page refresh

### LINKED ISSUE #481: Review page refresh
The Review page is shown whenever a user is about to publish data. This page currently appears in the following places:
* After choosing “Review” in the Data Entry View [[Figma](https://www.figma.com/file/brkAAatBCcvSnLJOIxZJIM/recidiviz-x-allison?node-id=592-22217&t=i91mPuajXacrzM2Q-0)]
* After uploading a spreadsheet [[Figma](https://www.figma.com/file/brkAAatBCcvSnLJOIxZJIM/recidiviz-x-allison?node-id=566-48811&t=Nya5X5zZWOkf3Bmy-0)]

We have redesigned the Review page to meet a few goals (Figma):
* Establish a more consistent structure for the Review page across all use cases.
* Make the important information easier to find, such as the overwrites, total number of records affected, etc.
* Celebrate the action of publishing (as it is the most important action).

Accomplishing this consists of a few components:
* New Structure: New page establishes a left side panel that contains the title, general description, and summary of records/overwrites/; the larger right panel contains all of the metrics in a scrollable pane.
* Confirmation modal: A new modal appears after pressing Publish that celebrates the actions and directs users to the Data Page or other previous page.
    * [Design for modal](https://www.figma.com/file/brkAAatBCcvSnLJOIxZJIM/recidiviz-x-allison?node-id=592-25857&t=Nya5X5zZWOkf3Bmy-0) when the Review page is appearing after Data Entry
   * [Design for modal](https://www.figma.com/file/brkAAatBCcvSnLJOIxZJIM/recidiviz-x-allison?node-id=552-35474&t=5Yg6bVVaTDprwe0G-0) when the Review Page is appearing after uploading a spreadsheet

Note: ignore any changes in the designs that are made to the action Buttons in top right corner. Keep the buttons the same for now.

ivan-kishko:
## Description of the change

> New design for review page for:
1. Data entry
2. Bulk publish/unpublish
3. Spreadsheet upload

## Type of change

Type: Non-breaking refactor

## Related issues

Closes #481 

### Development

### Code review


hobuobi (changes requested):
Leaving a few comments with suggested changes! 
- General: 
    - Definitely need a responsive version as a follow-up. Not blocking the merge though. 
- When you hit “Publish”: 
    - I don’t think we need both the toast and the modal. Just the modal is fine.
    - It would be great if hitting “Go to Data” took you to one of the metrics you were just looking at (for example, the first metric on the Review page you were just looking at). 
- Top bar: 
    - I think the current version has some option other than Publish in the top bar? Seems strange that we only have Publisher up there; would expect a "Save and Exit" for now or something. 
    - This wasn’t technically within scope, but I don’t know why we have the “Continue” and “New Upload” buttons in the body of the page instead of the navigation bar (next to “Cancel”). Seems inconsistent and a little unintuitive right now. 
- Summary: 
    - When we’re publishing less than maybe 10 records, it would be great to keep the list expanded. If more than ten, default to collapsed view. 
- Data: 
    - Need a better treatment for metrics that have long names – at the very least, the full name should be visible on hover (via tooltip? by hiding the date labels? etc) 
    - If there’s no data for a metric, it would be better to have a collapsed empty state for the section, rather than take up more space on the page with empty data. 
        - Same thing with breakdowns with no data; I would like to collapse these if there’s nothing relevant. I feel this pain especially with the Race/Ethnicity numbers. 
    - Expected to have the date label in the data table change colors or something when you hover over a column. Would help to know which date I’m highlighting / focusing on! (Maybe blue label with white text?) 
- Basic Functionality
    - I keep running into an issue where the data takes a long time to show up in the page! The Review page will show up with no data, and then it will populate a few seconds later. Not sure what’s up with this. 
    - “No Metrics to Review” page still has a Publish button? Shouldn’t be there – probably a Cancel or Return button.

ivan-kishko:
> When you hit “Publish”:
> * I don’t think we need both the toast and the modal. Just the modal is fine.

[3b7ebc9](https://github.com/Recidiviz/justice-counts/pull/507/commits/3b7ebc9b2368e056d94ae32d0d8b22c7abf608fb)

ivan-kishko:
> * It would be great if hitting “Go to Data” took you to one of the metrics you were just looking at (for example, the first metric on the Review page you were just looking at).

[28f6951](https://github.com/Recidiviz/justice-counts/pull/507/commits/28f6951b1720fb27b2a0cddfbd6f64565e3de43b)

ivan-kishko:
> * I keep running into an issue where the data takes a long time to show up in the page! The Review page will show up with no data, and then it will populate a few seconds later. Not sure what’s up with this.

[3f4909a](https://github.com/Recidiviz/justice-counts/pull/507/commits/3f4909a075ea1633fa6a96a06a863ab527aea20e)

ivan-kishko:
> Summary:
> * When we’re publishing less than maybe 10 records, it would be great to keep the list expanded. If more than ten, default to collapsed view.

[0d32cce](https://github.com/Recidiviz/justice-counts/pull/507/commits/0d32cce99d3432650f98a3db0024b90e0bac034d)

ivan-kishko:
> Top bar:
> * I think the current version has some option other than Publish in the top bar? Seems strange that we only have Publisher up there; would expect a "Save and Exit" for now or something.
> * This wasn’t technically within scope, but I don’t know why we have the “Continue” and “New Upload” buttons in the body of the page instead of the navigation bar (next to “Cancel”). Seems inconsistent and a little unintuitive right now.



> * “No Metrics to Review” page still has a Publish button? Shouldn’t be there – probably a Cancel or Return button.

Hello @hobuobi. Can you please specify which review page you referring to in this bullet points. I am not sure I understand. Thanks

> * Same thing with breakdowns with no data; I would like to collapse these if there’s nothing relevant. I feel this pain especially with the Race/Ethnicity numbers.

And I dont think I understand the idea behind this one

ivan-kishko:
> * Need a better treatment for metrics that have long names – at the very least, the full name should be visible on hover (via tooltip? by hiding the date labels? etc)

Because of how table works on HTML it is kinda hard to do both. For now I added title tag so full name is seen via small tooltip after a second or two.  If you want something specific I think I need a design so I am not wasting time on something that can be disapproved

 [bfdae37](https://github.com/Recidiviz/justice-counts/pull/507/commits/bfdae37a6be940d91347daf1dad76266491f8969)

ivan-kishko:
> * If there’s no data for a metric, it would be better to have a collapsed empty state for the section, rather than take up more space on the page with empty data.

How should collapsed state look? I mean that metric name and dates that are on a same row are both part of a DataTableViewComponent. Should it be like only metric name with plus sign? Or what are you thinking about?

publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const { reportStore, datapointsStore } = useStore();
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [loadingError, setLoadingError] = useState<string | undefined>(
+    undefined
+  );
+
+  useEffect(() => {
+    const initialize = async () => {
+      // probably need better way to get reports metrics
+      await datapointsStore.getDatapoints(agencyId);
```
lilidworkin, line 46: Ah, yes -- I actually wasn't expecting you to do this part, @ivan-kishko ! @mxosman is going to work on a proper solution to this that involves changing our backend. I think this is probably fine for now and @mxosman can just change it later?

publisher/src/pages/Reports.tsx
```diff
@@ -480,6 +480,37 @@ const Reports: React.FC = () => {
                           Delete
                         </ReportActionsButton>
                       )}
+                      {bulkAction === "publish" && selectedRecords.length > 0 && (
```
lilidworkin, line 483: Sorry for the confusion @ivan-kishko -- I actually wasn't expecting you to do this part. @mxosman is working on the backend piece for when the user hits Publish or Unpublish here. I think it's fine to keep this, and @mxosman can change it later? Up to you guys though. In this PR, I was mostly expecting @ivan-kishko just to focus on redesigning the Review Page where it _currently_ appears (after publishing a single Record, and after Bulk Upload), not adding it to appear in this third place (after Bulk Publishing). Sorry for not being more clear about this!

lilidworkin (commented):
Hi @ivan-kishko (cc @hobuobi , @mxosman ):

This is all looking great! I have a bunch of comments/ideas though.

## Code Organization

My main comment about the code is this: currently you have three different components for the Review Page, one for each place where it appears:

* `PublishConfirmation.tsx` appears after clicking Review from a single record
* `BulkActionReviewConfirmation.tsx` appears after clicking Bulk Publish from Records page
* `ReviewMetrics.tsx` appears after Bulk Upload

**Can we consolidate these three files into one?** Or at least, can we have a shared component that all three files rely on? The look and feel of the page should be almost identical in these three places. That way we can re-use more code and prevent duplication, and we can ensure the behavior of the page stays consistent. 

Let me know if this is going to be very hard. I'm also curious to hear @mxosman 's thoughts on it.

## Humphrey's Comments

> Definitely need a responsive version as a follow-up. Not blocking the merge though.

I think there is a design for a responsive version [here](https://www.figma.com/file/brkAAatBCcvSnLJOIxZJIM/recidiviz-x-allison?node-id=883-3730&t=0a8nCmAPCGlH815N-0) in Figma. Ivan, is that enough for you to work from?

> “No Metrics to Review” page still has a Publish button? Shouldn’t be there – probably a Cancel or Return button.

Humphrey is talking about when the Review Page appears after Bulk Upload. If you upload a blank, empty template, the page looks like this. Since there’s no data, instead of Publish, it should say Close (that’s the same as the current behavior on main). And the modal shouldn’t appear. Thanks!

Screenshot 2023-03-18 at 10 00 48 AM

> When we’re publishing less than maybe 10 records, it would be great to keep the list expanded. If more than ten, default to collapsed view.

Ivan, your fix to this looks great! I think there is one edge case though — when you publish a single record directly from the Record page, it looks like this section is still collapsed by default. 

>  I think the current version has some option other than Publish in the top bar? Seems strange that we only have Publisher up there; would expect a "Save and Exit" for now or something.

In my task for Ivan, I told him not to worry too much about the top bar, because some of this depends on the backend changes that Mahmoud is going to make. But to clarify/summarize, let me break this down by the different types of review pages:
1. Single Record Page —> Review page 
    1. The buttons that appear are “Back to Data Entry”, “Exit Without Publishing”, and “Publish”. **Let’s keep these as they are!**
2. Records Page —> Bulk Publish —> Review Page
    1. The buttons that appear are “Cancel” and “Publish”. **Let’s change “Cancel” to “Exit Without Publishing”. The functionality should stay the same.**
3. Bulk Upload —> Review Page
    1. The buttons that currently appear are just “Publish”. **Let’s also add an “Exit Without Publishing” button that takes you to the Records page.**

> 	Need a better treatment for metrics that have long names – at the very least, the full name should be visible on hover (via tooltip? by hiding the date labels? etc)

I think we can address this in a follow-up PR — I’ll make a ticket to come back to it.

> 	If there’s no data for a metric, it would be better to have a collapsed empty state for the section, rather than take up more space on the page with empty data.Same thing with breakdowns with no data; I would like to collapse these if there’s nothing relevant. I feel this pain especially with the Race/Ethnicity numbers.

I think we can address these in a follow-up PR as well. I'll make a ticket.

## Other thoughts of my own:

1. Do we need to restrict the height of the “Review and Publish” section? Right now it forces you to vertically scroll, even when there is plenty of space left underneath it. I also see a horizontal scrollbar in the right panel, even when there doesn’t need to be.

Screenshot 2023-03-18 at 10 05 56 AM

2. Small thing -- the design between breakdowns seems different from what is in the Figma, I think? In other words, in the screenshot below, I think there should be more space above "Race/Ethnicity", instead of a line?

Screenshot 2023-03-18 at 10 34 29 AM

3. _Very_ small thing -- if you are publishing multiple records, can we change the text "Here’s a breakdown of data from the record." to "Here’s a breakdown of data from the records."

Thank you for all your hard work on this Ivan -- it looks really really good!

ivan-kishko:
> 1. Do we need to restrict the height of the “Review and Publish” section? Right now it forces you to vertically scroll, even when there is plenty of space left underneath it. I also see a horizontal scrollbar in the right panel, even when there doesn’t need to be.

Hello @lilidworkin. Yes I kinda have to do that because for desktop view this part of page is fixed position and if I won't restrict its height and there will be a lot of records/metrics it will overlap with itself left part of the footer. I make its height a bit bigger but max height will be 48px above footer (since we decided that it will be the min space between footer and content above)

ivan-kishko:
> I think there is a design for a responsive version [here](https://www.figma.com/file/brkAAatBCcvSnLJOIxZJIM/recidiviz-x-allison?node-id=883-3730&t=0a8nCmAPCGlH815N-0) in Figma. Ivan, is that enough for you to work from?
> 
> >

It is but I will make it go vertical from min desktop width (1200px i guess) instead of 700px like it is in figma since just summary is 500px wide therefore on 700 px only part of names in table will be seen

lilidworkin:
> About this - the main reason this is 3 different pages is that they receive different type of data and in a different way. But I think I can come up with something or maybe @mxosman can suggest something as well. One thing if it matters - these pages are using the same styled components (like i dont remember maybe there are few components that are changed depending on a page but generally they are all the same)

I think we should do our best to try and see how it goes! It's not the styles I'm worried about, it's the functionality. 

What if they all have their own component that converts the different types of data into one shared type of data, and then passes that into a shared component?

lilidworkin:
> Hello @lilidworkin. Yes I kinda have to do that because for desktop view this part of page is fixed position and if I won't restrict its height and there will be a lot of records/metrics it will overlap with itself left part of the footer.

Is there no way to have it know where the footer starts, and only scroll if it would overlap with the footer?

ivan-kishko:
> > Hello @lilidworkin. Yes I kinda have to do that because for desktop view this part of page is fixed position and if I won't restrict its height and there will be a lot of records/metrics it will overlap with itself left part of the footer.
> 
> Is there no way to have it know where the footer starts, and only scroll if it would overlap with the footer?

Hello!) Please check latest commit I think this is the best I can do. Well to be honest I can make bottom padding to be slightly above footer (less than footer margin) but it still be here regardless of footer seen or not. And no, this is not how fixed positioned elements work. They basically removed from DOM flow and do not know anything about other elements properties. 

https://developer.mozilla.org/en-US/docs/Web/CSS/position#:~:text=with%20other%20margins.-,fixed,element%20is%20placed%20in%20the%20same%20position%20on%20every%20page.,-sticky

ivan-kishko:
> > Hello @lilidworkin. Yes I kinda have to do that because for desktop view this part of page is fixed position and if I won't restrict its height and there will be a lot of records/metrics it will overlap with itself left part of the footer.
> 
> Is there no way to have it know where the footer starts, and only scroll if it would overlap with the footer?

[631c26a](https://github.com/Recidiviz/justice-counts/pull/507/commits/631c26a997fc7d8945c29b50d86ba33efe4802fe) Check out this one

publisher/src/components/Reports/BulkActionReview.tsx
mxosman: Yes - totally fine with me! That's what I assumed - I think it might've been helpful to receive some form of data (this or mocks) to keep iterating on the design.

publisher/src/pages/Reports.tsx
mxosman: Not a problem at all! Totally good with keeping it as is and me replacing!

mxosman:
> @lilidworkin Also please review latest commit, I think I fixed all that you pointed out except things that go on another task and responsive design
> 
> > My main comment about the code is this: currently you have three different components for the Review Page, one for each place where it appears:
> > 
> > * `PublishConfirmation.tsx` appears after clicking Review from a single record
> > * `BulkActionReviewConfirmation.tsx` appears after clicking Bulk Publish from Records page
> > * `ReviewMetrics.tsx` appears after Bulk Upload
> > 
> > **Can we consolidate these three files into one?** Or at least, can we have a shared component that all three files rely on? The look and feel of the page should be almost identical in these three places. That way we can re-use more code and prevent duplication, and we can ensure the behavior of the page stays consistent.
> > Let me know if this is going to be very hard. I'm also curious to hear @mxosman 's thoughts on it.
> 
> About this - the main reason this is 3 different pages is that they receive different type of data and in a different way. But I think I can come up with something or maybe @mxosman can suggest something as well. One thing if it matters - these pages are using the same styled components (like i dont remember maybe there are few components that are changed depending on a page but generally they are all the same)

---

> I think we should do our best to try and see how it goes! It's not the styles I'm worried about, it's the functionality.
> 
> What if they all have their own component that converts the different types of data into one shared type of data, and then passes that into a shared component?

---
I definitely think it's worth consolidating these as much as possible. The publish review/confirmation components are the ones I'm least familiar with - so, I don't have any good suggestions to offer at the moment. Since they all look pretty much the same, I would see if there's a central data structure they can all conform to - and if there's a way to transform their data structure, similar to Lili's suggestion - so they can render in the same component. As I dive deeper in the review, I might have more specific suggestions, but that's where my mind would go (finding a way to funnel the 3 different types of data into one that can be easily rendered in one component)

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -15,132 +15,185 @@
...
-  Subheading,
+  Summary,
+  SummarySection,
+  SummarySectionLine,
+  SummarySectionsContainer,
+  SummarySectionTitle,
 } from "./ReviewMetrics.styles";
+import { ReviewMetricsModal } from "./ReviewMetricsModal";
+
+type Overwrite = {
```
mxosman, line 50: tiny nit: I would change this to something like `MetricOverwrites` for clarity

mxosman (commented):
Wow Ivan - this was a lot of work executed beautifully as always! Left a few notes, thoughts, and questions - nothing glaring to me that hasn't been pointed out or that won't be done in a follow up! Loved your solutions & implementations.

ivan-kishko:
@lilidworkin Hello) Check out this one [654f7d3](https://github.com/Recidiviz/justice-counts/pull/507/commits/654f7d35c7ed76e76006adb41ce5ef3b3d708b1d)

publisher/src/components/Reports/PublishConfirmation.tsx
```diff
@@ -15,246 +15,236 @@
...
+//
+//         {/* Contexts */}
+//         {metric.contexts.length > 0 &&
+//           metric.contexts.map((context) => {
+//             return <Context key={context.key} context={context} />;
+//           })}
+//       </MetricDetailWrapper>
+//     </Metric>
+//   );
+// };
```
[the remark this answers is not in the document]
ivan-kishko: Yeah) Its just, there was so much work done before this redesign and I thought if you want to go back to prev design it is going to be a pain) But I guess i am overthinking and too careful) I will delete all this)

publisher/src/components/ReviewMetrics/ReviewMetrics.styles.tsx
```diff
@@ -33,20 +34,73 @@ export const Container = styled(DataUploadContainer)`
 `;
 
 export const MainPanel = styled.div`
-  width: 100%;
-  max-width: ${MAIN_PANEL_MAX_WIDTH}px;
-  margin-top: ${56 + HEADER_BAR_HEIGHT}px;
-  margin-bottom: 128px;
+  max-width: 100%;
+  padding: ${40 + HEADER_BAR_HEIGHT}px 0 128px 612px;
```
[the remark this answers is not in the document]
ivan-kishko: [46c8187](https://github.com/Recidiviz/justice-counts/pull/507/commits/46c818719266bb4282a20a4925c85e0953ac1b20)

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <ReviewMetricsModalHint>
+          You can view the published data in the Data tab.
+        </ReviewMetricsModalHint>
+        <ReviewMetricsModalButtonsContainer>
+          <ReviewMetricsModalButton
+            buttonColor="blue"
+            onClick={() => navigate(`/agency/${agencyId}/data`)}
+          >
+            View Data
+          </ReviewMetricsModalButton>
```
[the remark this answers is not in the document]
ivan-kishko: well it is what design says so up to you

publisher/src/components/Reports/ReviewPublishModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+              )
+            }
+          >
+            Go to Data
+          </ReviewPublishModalButton>
+        </ReviewPublishModalButtonsContainer>
+      </ReviewPublishModalContainer>
+    </ReviewPublishModalWrapper>
+  );
+};
```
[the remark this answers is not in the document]
ivan-kishko: probably, but will do that after pages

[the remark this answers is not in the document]
ivan-kishko: probably, but will do that after pages

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
ivan-kishko: [20dbcc5](https://github.com/Recidiviz/justice-counts/pull/507/commits/20dbcc5ca5b188f822127be284eec44b7b3b76db)

publisher/src/components/Reports/ReviewReportDataEntry.tsx
```diff
@@ -16,62 +16,50 @@
...
     reportStore.reportOverviews[reportID].agency_id !== Number(params.agencyId)
   ) {
     return <NotFound />;
   }
 
   return (
-    <ReportDataEntryWrapper>
-      <PublishConfirmationSummaryPanel reportID={reportID} />
+    <ReviewWrapper>
+      {/* <PublishConfirmationSummaryPanel reportID={reportID} /> */}
```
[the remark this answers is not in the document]
ivan-kishko: [60cfcb0](https://github.com/Recidiviz/justice-counts/pull/507/commits/60cfcb0de4717ca916f4697ccfc645bde88f61e3)

publisher/src/components/Reports/PublishConfirmation.tsx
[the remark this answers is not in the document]
ivan-kishko: [60cfcb0](https://github.com/Recidiviz/justice-counts/pull/507/commits/60cfcb0de4717ca916f4697ccfc645bde88f61e3)

publisher/src/components/Reports/PublishConfirmation.tsx
```diff
@@ -338,32 +345,102 @@ const PublishConfirmation: React.FC<{ reportID: number }> = ({ reportID }) => {
...
+      {/*              metric={metric} */}
+      {/*              metricHasError={checkMetricForErrors(metric.key)} */}
+      {/*              index={i} */}
+      {/*            /> */}
+      {/*          ) */}
+      {/*        ); */}
+      {/*      })} */}
+      {/*    </div> */}
+      {/*  )} */}
+      {/* </ConfirmationDialogueWrapper> */}
```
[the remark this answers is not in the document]
ivan-kishko: [60cfcb0](https://github.com/Recidiviz/justice-counts/pull/507/commits/60cfcb0de4717ca916f4697ccfc645bde88f61e3)

publisher/src/components/Reports/PublishConfirmation.tsx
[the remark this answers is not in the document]
mxosman: Ah - thank you for clarifying! Definitely appreciate you looking out! I think it's fine to delete because we can refer to older commits if we want to revive this.

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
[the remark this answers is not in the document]
mxosman: No action needed. I'll bring this up during one of our testing sessions - still back and forth. Thanks, Ivan.

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -15,132 +15,185 @@
...
+  const { uploadedMetrics, fileName } = state as {
+    uploadedMetrics: UploadedMetric[] | null;
+    fileName: string;
+  };
   const navigate = useNavigate();
+  const [isSuccessModalOpen, setIsSuccessModalOpen] = useState(false);
 
   useEffect(() => {
-    if (!(location.state as UploadedMetric[] | null)) {
+    if (!uploadedMetrics) {
```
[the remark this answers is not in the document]
ivan-kishko: well there is no need for that I guess since if you have not selected any records button for publish/unpublish do not appear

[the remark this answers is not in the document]
mxosman: Ah yes! Great point. Thank you for clarifying.

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -15,135 +15,233 @@
...
+              onClick={onClick}
+              disabled={disabled}
+            >
+              {name}
+            </Button>
+          ))}
+        </ReviewMetricsButtonsContainer>
         {
-          // TODO(#24): Add Publish button to publish multiple reports at once
+          // TODO(#24): Add Publish button to publish multiple reports at once (todo from upload metrics)
```
mxosman, line 141: We can delete this whole TODO - I don't think we need this anymore. cc'ing @lilidworkin just to confirm (I can close out task #24 as well if you agree).

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+  text-align: center;
+  max-width: 264px;
+`;
+export const ReviewPublishModalButtonsContainer = styled(
+  RemoveRecordsModalButtonsContainer
+)``;
+export const ReviewPublishModalButton = styled(ReportActionsButton)`
+  margin-left: unset;
+  ${typography.sizeCSS.normal};
+`;
```
mxosman, line 61: nit: could move these to the .styles.tsx file

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -15,135 +15,233 @@
...
-      new Set(metric.datapoints.map((dp) => dp.start_date))
-    ).sort((a, b) => new Date(a).getTime() - new Date(b).getTime());
+type ReviewMetricsProps = {
+  title: string;
+  description: string | React.ReactNode;
+  buttons: ReviewHeaderActionButton[];
+  metrics: ReviewMetric[];
+  metricOverwrites?: ReviewMetricOverwrites[];
+  records?: ReportOverview[];
+};
```
mxosman, line 88: nit: no strong feelings - but might be worth moving to a `types.ts` file within this folder

mxosman (commented):
WOW, Ivan! This was a big undertaking -- I love the way you thought through this and how you consolidated the review and modal components and transformed the data in the other components to seamlessly feed into it. Thank you so much for all of your work on this! Most of my feedback are small nits.

There is one behavior that I wanted to note going from Data Entry to Publish Review -- there's a blip in between when the components are mounting and unmounting that we see a bit of the home page and header. Let me know if you're able to see this as well:

https://user-images.githubusercontent.com/59492998/227659472-b0f7b5b3-af26-4702-b4f9-b4a0e62667c4.mov

Any ideas on what's causing that? It looks like the component doesn't mount until we get a response from the `datapoints` endpoint. Maybe we can let it mount but add a loader (until the `datapoints` endpoint returns a response)? Let me know if I'm not making sense.

Aside from the feedback above - I'll be happy to approve this! I also want to give @lilidworkin a chance to do another pass if she wants to.

common/hooks/useIsFooterVisible.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (element) {
+    const footerRectangle = element.getBoundingClientRect();
+    const viewHeight = Math.max(
+      document.documentElement.clientHeight,
+      window.innerHeight
+    );
+    return !(
+      footerRectangle.bottom < 0 || footerRectangle.top - viewHeight >= 0
+    );
+  }
```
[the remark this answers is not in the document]
lilidworkin: Yea @ivan-kishko thank you so much for this fix! It looks a lot better now that the left side panel isn't fixed height!

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+  recordsCount,
+  fileName,
+  action,
+}) => {
+  const { agencyId } = useParams();
+  const navigate = useNavigate();
+
+  const goToDataPage = () => {
+    if (systemSearchParam && metricSearchParam) {
+      navigate(
```
lilidworkin, line 81: Question (maybe for @mxosman ) -- if only the system param OR the metric param is present, do we want to include them in the URL? Or do we only want to include them if both are present (the current behavior)?

publisher/src/pages/Reports.tsx
```diff
@@ -502,15 +499,43 @@ const Reports: React.FC = () => {
...
+                      {/*      buttonColor="orange" */}
+                      {/*      onClick={() => { */}
+                      {/*        clearAllSelectedRecords(); */}
+                      {/*        clearBulkAction(); */}
+                      {/*        publishOrUnpublishMultipleReports("DRAFT"); */}
+                      {/*      }} */}
+                      {/*    > */}
+                      {/*      Unpublish */}
+                      {/*    </ReportActionsButton> */}
+                      {/*  )} */}
```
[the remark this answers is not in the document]
lilidworkin: +1

publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -0,0 +1,190 @@
...
+            key: metric.key,
+          };
+          return [...acc, reviewMetric];
+        }, [] as ReviewMetric[])
+      : [];
+  const records = recordsIds.map(
+    (recordID) => reportStore.reportOverviews[recordID]
+  );
+  const publishActionTitle = "Review & Publish";
+  const unpublishActionTitle = "Review Unpublish";
```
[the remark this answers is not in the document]
lilidworkin: +1

publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -0,0 +1,190 @@
...
+  );
+  const unpublishActionDescription = `Here’s a breakdown of data you’ve selected to unpublish. All data in ${
+    records.length > 1
+      ? `these ${REPORTS_LOWERCASE}`
+      : `this ${REPORT_LOWERCASE}`
+  } will be saved, and you can re-publish at any time.`;
+
+  const buttons: ReviewHeaderActionButton[] = [
+    {
+      name: "Exit without Publishing",
```
lilidworkin, line 145: nit -- if you're currently Unpublishing, this should say "Exit without Unpublishing"

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -15,135 +15,233 @@
...
 
-    // keep track of count of overwritten values
-    let overwrittenValuesCount = 0;
-
-    metric.datapoints.forEach((dp) => {
-      if (dp.old_value !== null) {
-        overwrittenValuesCount += 1;
-      }
-    });
+export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
```
lilidworkin, line 90: @ivan-kishko you did _such_ a good job making this a re-usable component between all 3 review screens! I love the properties you chose to pass into it. Really excellent job thinking it through!

publisher/src/components/Reports/DataEntryForm.tsx
```diff
@@ -73,9 +73,13 @@ const TopBarButtonsContainer = styled.div<{ showDataEntryHelpPage: boolean }>`
   opacity: ${({ showDataEntryHelpPage }) => (showDataEntryHelpPage ? 0 : 1)};
 `;
 
-const DataEntryTopBarButton = styled(Button)`
+const DataEntryTopBarButton = styled(Button)<{ isSaveInProgress?: boolean }>`
```
lilidworkin, line 76: Is the goal of these changes to prevent the user from pressing "Review" when data is currently being saved?

publisher/src/components/DataUpload/UploadReview.tsx
```diff
@@ -0,0 +1,115 @@
...
+  };
+  const navigate = useNavigate();
+  const [isSuccessModalOpen, setIsSuccessModalOpen] = useState(false);
+
+  useEffect(() => {
+    if (!uploadedMetrics) {
+      // no metrics in passed in navigation state, redirect to home page
+      navigate(`/agency/${agencyId}/${REPORTS_LOWERCASE}`, { replace: true });
+    }
+  });
```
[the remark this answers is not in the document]
lilidworkin: Maybe related to @mxosman 's question -- how is the `if (!uploadedMetrics)` check on line 42 different from the one on line 48?

publisher/src/components/DataUpload/UploadReview.tsx
```diff
@@ -0,0 +1,115 @@
...
+      }
+    });
+  });
+  const title =
+    filteredMetrics.length > 0 ? "Review Upload" : "No Metrics to Review";
+  const description =
+    filteredMetrics.length > 0
+      ? "Here’s a breakdown of data from the file you uploaded. You can publish these changes now, or save as a draft for later."
+      : "Uploaded file contains no metrics to review.";
+  const buttons: ReviewHeaderActionButton[] =
```
lilidworkin, line 79: The logic for these buttons looks perfect, thank you for your patience on this!

publisher/src/components/Reports/DataEntryReview.tsx
```diff
@@ -0,0 +1,197 @@
...
+        });
+        setIsPublishable(true);
+      }
+    }
+  };
+
+  useEffect(() => {
+    const initialize = async () => {
+      formStore.validatePreviouslySavedInputs(reportID);
+      await datapointsStore.getDatapoints(agencyId);
```
lilidworkin, line 92: So this is fine for now and I totally see why you had to add this call here @ivan-kishko , but @mxosman you're going to want/need to refactor this later for performance reasons. We should ideally remove the `datapointsStore.getDatapoints(agencyId)` call from both BulkActionReview and DataEntryReview. This endpoint is really expensive because it pulls _all_ datapoints for an agency. BulkActionReview only needs datapoints for some reports (the ones we are publishing), and DataEntryReview only needs datapoints for one report. So it's overkill to call this method. 

@mxosman I was going to say we could call your new `get_reports_by_id` method instead, and either pass in a list of reports or just a single report, but I actually think that might not quite work. That method returns "metrics" objects, whereas this `getDatapoints` method returns `datapoints` objects, and the review page requires the latter. I think it should be doable to refactor `get_reports_by_id` to return datapoints instead of metrics. Anyway, let me know if you're following my train of thought here and we can discuss further!

But @ivan-kishko you don't have to worry about this, you made the absolute right decision here -- this is the only backend endpoint currently available that does what you need!

publisher/src/components/Reports/ReviewReportDataEntry.tsx
```diff
@@ -1,78 +0,0 @@
...
-
-  if (
-    reportStore.reportOverviews[reportID].agency_id !== Number(params.agencyId)
-  ) {
-    return <NotFound />;
-  }
-
-  return (
-    <ReportDataEntryWrapper>
-      <PublishConfirmationSummaryPanel reportID={reportID} />
```
lilidworkin, line 72: nit, I think we can delete the PublishConfirmationSummaryPanel page as well?

publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -0,0 +1,190 @@
...
+  // review component props
+  const metrics =
+    enabledMetrics.length > 0
+      ? enabledMetrics.reduce((acc, metric) => {
+          const reviewMetric = {
+            datapoints: datapointsStore.rawDatapointsByMetric[
+              metric.key
+            ].filter((dp) => dp.report_id && recordsIds.includes(dp.report_id)),
+            display_name: metric.display_name,
+            key: metric.key,
```
lilidworkin, line 115: Because we don't pass in `metricHasValidInput` here, the checkmarks in the left sidepanel will always be blank circles on the Bulk Action Review page (and same for the Bulk Upload Review page). The Data Entry Review page is the only one that can ever have blue checkmarks.

see for instance:
Screenshot 2023-03-25 at 12 52 00 PM

I think we should fix this, but I'll make another ticket for it and we can do it in a follow-up PR!

I think the idea is that the checkmark should be filled if we have data for the top-level metric, is that right @mxosman ? Basically I'm referring to this logic:

```
metricHasValidInput={Boolean(
  formStore.metricsValues?.[reportID]?.[metric.key]?.value
)}
```

So we'd need to figure out an analog of this logic for when you're bulk publishing/uploading a metric over multiple time periods. I guess we'd want to check if, for each time period for which you are uploading the metric, do we have data for the top level metric?

We could also punt on this and just use the blue checkmark in all cases. I'm not opposed to that :D @mxosman why don't you give it some thought and then suggest a path forward to @ivan-kishko ?

Again, not blocking this PR though!

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
lilidworkin: That's right!

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -15,135 +15,233 @@
...
+                    metricHasValidInput,
+                  }) => (
+                    <SummarySectionLine key={key}>
+                      {!metricHasError && metricHasValidInput && (
+                        <MetricStatusIcon src={checkIcon} alt="" />
+                      )}
+                      {metricHasError && (
+                        <MetricStatusIcon src={errorIcon} alt="" />
+                      )}
+                      {!metricHasError && !metricHasValidInput && <EmptyIcon />}
```
lilidworkin, line 180: As discussed in my comment, above, we will always be in the EmptyIcon state unless we are coming to this page from Data Entry. There isn't an obvious fix for this, so let's address in a follow up PR after giving it more thought.

lilidworkin (approved):
@ivan-kishko , this looks fantastic! I left a few small comments, but after you address those, you can go ahead and merge. I'll file tickets for some of the follow-up work. Thank you so much for all your hard work on this!

> There is one behavior that I wanted to note going from Data Entry to Publish Review -- there's a blip in between when the components are mounting and unmounting that we see a bit of the home page and header. Let me know if you're able to see this as well

I actually wasn't able to repro this! This is what I see:

https://user-images.githubusercontent.com/99899591/227751018-6270573c-63eb-435a-a799-43cdd557b739.mov

In any case @mxosman you're going to have to change around some of this logic, so maybe it's okay? But if @ivan-kishko is able to repro then feel free to fix!


One small thing I did see @ivan-kishko is if you go here, there's some text overlap: http://localhost:3000/agency/161/data?system=jails&metric=jails_funding. Is it possible to have the table column width expand based on the text length? If it's hard, don't worry about it for now.

Screenshot 2023-03-25 at 8 57 29 PM

publisher/src/components/Reports/DataEntryReview.tsx
lilidworkin: Filed ticket for @mxosman [here](https://github.com/Recidiviz/justice-counts/issues/529).

publisher/src/components/Reports/BulkActionReview.tsx
lilidworkin: Filed this ticket for both @mxosman / @ivan-kishko [here](https://github.com/Recidiviz/justice-counts/issues/530). I think @mxosman  can help figure out the right behavior (fine if you want to discuss with Humphrey, but I personally would just pick whatever we think is easiest and sensible), and then @ivan-kishko  can implement!

ivan-kishko:
> WOW, Ivan! This was a big undertaking -- I love the way you thought through this and how you consolidated the review and modal components and transformed the data in the other components to seamlessly feed into it. Thank you so much for all of your work on this! Most of my feedback are small nits.
> 
> There is one behavior that I wanted to note going from Data Entry to Publish Review -- there's a blip in between when the components are mounting and unmounting that we see a bit of the home page and header. Let me know if you're able to see this as well:
> 
> Any ideas on what's causing that? It looks like the component doesn't mount until we get a response from the `datapoints` endpoint. Maybe we can let it mount but add a loader (until the `datapoints` endpoint returns a response)? Let me know if I'm not making sense.
> 
> Aside from the feedback above - I'll be happy to approve this! I also want to give @lilidworkin a chance to do another pass if she wants to.

Hello @mxosman I know why is that. It is because data entry and review headers are allocated above main header. What you are suggesting is fine but as Humphrey mentioned if you change something in data entry there is like few seconds you see old data in review and then it updates. So it is either load and see main header or no loader and see data changing inside tables. If all pages would have their own header it would not be a problem but our main header is located apart from pages that is why we see it if something like this going on

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
ivan-kishko: well for review pages it is kinda both or none since we review metrics, they will always have a system, if (like in upload) no metrics to review system does not make sense cause you are not able to publish anyway and see whit option in modal

publisher/src/components/Reports/ReviewReportDataEntry.tsx
ivan-kishko: it is removed already, probably you look at old commit

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
ivan-kishko: [018b930](https://github.com/Recidiviz/justice-counts/pull/507/commits/018b930a0363c9f93b6bbcd0042a8ab02ef21175)

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
ivan-kishko: [018b930](https://github.com/Recidiviz/justice-counts/pull/507/commits/018b930a0363c9f93b6bbcd0042a8ab02ef21175)

ivan-kishko:
[f643153](https://github.com/Recidiviz/justice-counts/pull/507/commits/f643153e4687cac02b80c7044adb2ca869f73b8d) fixed table names here @lilidworkin

publisher/src/components/Reports/BulkActionReview.tsx
ivan-kishko: [438efae](https://github.com/Recidiviz/justice-counts/pull/507/commits/438efae8659cc6db3aa3fe552f43af64df18fc1a)

ivan-kishko:
@hobuobi Hello Humphrey!) I think I need an approve from you to merge this and go on

publisher/src/components/ReviewMetrics/ReviewMetricsModal.tsx
mxosman: At least the way the data page is set up now, not having both will redirect you to the homepage. I wouldn't try to hit this endpoint without having both.

ivan-kishko: I think it is not redirect but first system first metric. this function is wrote just to not see undefined in address bar for system and metric

ivan-kishko: and to not catch an error <code>TypeError: Cannot read properties of undefined (reading 'toLowerCase')</code>

mxosman:
> > There is one behavior that I wanted to note going from Data Entry to Publish Review -- there's a blip in between when the components are mounting and unmounting that we see a bit of the home page and header. Let me know if you're able to see this as well:
> > Any ideas on what's causing that? It looks like the component doesn't mount until we get a response from the `datapoints` endpoint. Maybe we can let it mount but add a loader (until the `datapoints` endpoint returns a response)? Let me know if I'm not making sense.
> 
> Hello @mxosman I know why is that. It is because data entry and review headers are allocated above main header. What you are suggesting is fine but as Humphrey mentioned if you change something in data entry there is like few seconds you see old data in review and then it updates. So it is either load and see main header or no loader and see data changing inside tables. If all pages would have their own header it would not be a problem but our main header is located apart from pages that is why we see it if something like this going on

Thanks for clarifying, Ivan. If we added the loader to the review page (we just see the review page header and the content of the review page is just a loader), that should take care of seeing the old data in the review page coming from data entry, right? Because the user will just see a loader w/ the review page header for a few moments until datapoints endpoint returns (at which point they'll see the rest of the page). I'll play around with this after I make endpoint adjustments and move some logic around! No need to worry about it in this PR!

publisher/src/components/Reports/DataEntryReview.tsx
mxosman: Ivan brought up a really good point about the `metrics` data structure having a `datapoints` property that's almost always set to `null`. I can't remember if that was an intentional decision to deprecate this property in favor of calling the `/datapoints` endpoint or because it was too expensive or not - but, I wonder if it would be helpful now to include datapoints for each report?

Screenshot 2023-03-27 at 2 20 40 PM

publisher/src/components/Reports/ReviewReportDataEntry.tsx
lilidworkin: Oops, thanks!

publisher/src/components/Reports/DataEntryReview.tsx
lilidworkin: Interesting, thanks so much for calling this out, @ivan-kishko and @mxosman ! A few thoughts (really just for @mxosman  -- @ivan-kishko can ignore this for now):

* I don't remember the story behind this property -- can you investigate a bit on the backend and see what you can figure out? The way I'd do that is do a search for `"datapoints"` -- _include_ the quotation marks, because we're looking for backend code that looks something like 'return {"metrics": [{"datapoints": <some json> ...}, ...]`. Then, see if you can trace upwards to see when this property is populated and when it isn't. I believe this is something Terry did, so we're going to have to reverse engineer it because we can't ask him :(
* It's definitely not a bad idea to start using this property now! However, there are a few things to consider: 1) how expensive it is, and 2) when do we _actually_ need the datapoints. We only need the datapoints when we load the Review page, which only happens when the user is publishing a Record. Most of the time the user loads the Records page or an individual Record, they are not going to be publishing. So unless it's very cheap to add the datapoints to the response in all cases, we probably don't want to. Thus, my intuition is that we shouldn't add them in all cases, but we _should_ add them to your `get_reports_by_id` endpoint, which we call when the user hits publish. LMK if that makes sense!
* It probably won't be straightforward how to populate the `datapoints` property. When you get to this point in the project, it would be good to write up a few sentences explaining what you need to do, and then start a thread with me and @nichelle-hall to discuss the best approach!

Generally this is a little confusing -- we have two different data models -- Metrics (each metric has multiple datapoints) and Datapoints. Most of our frontend pages work with the Metrics data model, but a few pages deal with Datapoints -- I think just the Data page and the Review page. This isn't ideal and I'm not 100% sure why this is the case, but it is probably out of scope for now to refactor.

publisher/src/components/Reports/BulkActionReview.tsx
lilidworkin: I agree @mxosman ! And I'm actually pretty sure that the metric will only show up in this side panel if there is _some_ data provided. Is that right @ivan-kishko ? If so, then for now, @ivan-kishko can you actually just change the logic as follows:

* Review page after Data Entry Page: keep logic as currently implemented for empty check mark vs. red error vs. empty checkbox 
* Review page after Bulk Publish: always show blue checkmark
* Review page after Bulk Upload: always show blue checkmark

Thank you!

lilidworkin (approved):
Thank you @ivan-kishko ! Here are my thoughts on the remaining discussion items:

* Loading page: You can leave this as-is for now -- @mxosman can explore alternatives when he works on this later!
* Datapoints response: @mxosman will take it from here!
* Empty checkmarks: This is the only remaining TODO for you -- just change the logic to be always a blue checkmark (unless you're coming from Data Entry, in which case keep the logic as is).

After you address the last bullet point, you can go ahead and merge! I cleared out Humprey's old review so you should be unblocked.

publisher/src/components/Reports/DataEntryReview.tsx
mxosman: * Will do! Thank you for the strategy!
* That makes perfect sense! I agree - definitely not needed in all cases. I'll add the datapoints just for the reports returned by `get_reports_by_id` endpoint.
* Sounds great! Will do!

Agreed on the two different data models being confusing. You're right that it's only the data viz and review page that deal with datapoints - I'm wondering if it's just an ready-to-go transformed data structure that allows the FE to easily render them & filter them in a visual/table format? I feel like it might've been inspired by what Recharts needed -- but, I can't recall. Happy to revisit this if we ever want to explore consolidation/refactor -- it might be a matter of having the FE handle all of the data transformation - but would want to dig deeper into what inspired the separate endpoint and data structure.

PR #516, opened 2023-03-21
[Bulk Upload] Adjust title / definition of a successfully saved metric 

morden35:
## Description of the change

Previously, we considered metrics as being saved successfully via bulk upload if no errors were associated with those metrics. For example, the title below is shown when an empty template is uploaded (no data is saved).
image

We want to change this logic so that we consider metrics as being saved successfully via bulk upload if they are associated with datapoints in the database (if we actually persisted some data) and no errors were associated with those metrics.

Now, when an empty template is uploaded, the user sees the following:
Screenshot 2023-03-23 at 10 50 06 AM
When 1 metric was uploaded:
Screenshot 2023-03-22 at 11 51 09 AM
When 2 (or more) metrics were uploaded:
Screenshot 2023-03-22 at 11 51 24 AM

## Type of change

## Related issues

Related to [#19584](https://github.com/Recidiviz/recidiviz-data/issues/19584)

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataUpload/DataUpload.tsx
+++ b/publisher/src/components/DataUpload/DataUpload.tsx
@@ -161,8 +161,7 @@ export const DataUpload: React.FC = observer(() => {
   ): ErrorsWarningsMetrics => {
     const errorsWarningsAndSuccessfulMetrics = data.metrics.reduce(
       (acc, metric) => {
-        const isSuccessfulMetric = metric.datapoints.length > 0;
-        // console.log(isSuccessfulMetric);
+        const isSuccessfulMetric = metric.successful_ingest;
 
         /**
          * If there are no errors and only warnings, we still want to show the
--- a/publisher/src/components/DataUpload/DataUpload.tsx
+++ b/publisher/src/components/DataUpload/DataUpload.tsx
@@ -161,7 +161,13 @@ export const DataUpload: React.FC = observer(() => {
   ): ErrorsWarningsMetrics => {
     const errorsWarningsAndSuccessfulMetrics = data.metrics.reduce(
       (acc, metric) => {
-        const isSuccessfulMetric = metric.successful_ingest;
+        const noSheetErrorsFound =
+          metric.metric_errors.filter(
+            (sheet) =>
+              sheet.messages.filter((msg) => msg.type === "ERROR")?.length > 0
+          ).length === 0;
+        const isSuccessfulMetric =
+          noSheetErrorsFound && metric.datapoints.length > 0;
 
         /**
          * If there are no errors and only warnings, we still want to show the
--- a/publisher/src/components/DataUpload/DataUpload.tsx
+++ b/publisher/src/components/DataUpload/DataUpload.tsx
@@ -161,13 +161,6 @@ export const DataUpload: React.FC = observer(() => {
   ): ErrorsWarningsMetrics => {
     const errorsWarningsAndSuccessfulMetrics = data.metrics.reduce(
       (acc, metric) => {
-        // const noSheetErrorsFound =
-        //   metric.metric_errors.filter(
-        //     (sheet) =>
-        //       sheet.messages.filter((msg) => msg.type === "ERROR")?.length > 0
-        //   ).length === 0;
-        // const isSuccessfulMetric =
-        //   metric.metric_errors.length === 0 || noSheetErrorsFound;
         const isSuccessfulMetric = metric.datapoints.length > 0;
         // console.log(isSuccessfulMetric);
 
--- a/publisher/src/components/DataUpload/DataUpload.tsx
+++ b/publisher/src/components/DataUpload/DataUpload.tsx
@@ -25,11 +25,6 @@ import { observer } from "mobx-react-lite";
 import React, { useEffect, useMemo, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 
-import { useStore } from "../../stores";
-import logoImg from "../assets/jc-logo-vector-new.svg";
-import { REPORTS_LOWERCASE } from "../Global/constants";
-import { Logo, LogoContainer } from "../Header";
-import { Loader } from "../Loading";
 import {
   Button,
   DataUploadContainer,
@@ -40,6 +35,11 @@ import {
   SystemSelection,
   UploadFile,
 } from ".";
+import { useStore } from "../../stores";
+import logoImg from "../assets/jc-logo-vector-new.svg";
+import { REPORTS_LOWERCASE } from "../Global/constants";
+import { Logo, LogoContainer } from "../Header";
+import { Loader } from "../Loading";
 import {
   DataUploadResponseBody,
   ErrorsWarningsMetrics,
@@ -161,13 +161,15 @@ export const DataUpload: React.FC = observer(() => {
   ): ErrorsWarningsMetrics => {
     const errorsWarningsAndSuccessfulMetrics = data.metrics.reduce(
       (acc, metric) => {
-        const noSheetErrorsFound =
-          metric.metric_errors.filter(
-            (sheet) =>
-              sheet.messages.filter((msg) => msg.type === "ERROR")?.length > 0
-          ).length === 0;
-        const isSuccessfulMetric =
-          metric.metric_errors.length === 0 || noSheetErrorsFound;
+        // const noSheetErrorsFound =
+        //   metric.metric_errors.filter(
+        //     (sheet) =>
+        //       sheet.messages.filter((msg) => msg.type === "ERROR")?.length > 0
+        //   ).length === 0;
+        // const isSuccessfulMetric =
+        //   metric.metric_errors.length === 0 || noSheetErrorsFound;
+        const isSuccessfulMetric = metric.datapoints.length > 0;
+        // console.log(isSuccessfulMetric);
 
         /**
          * If there are no errors and only warnings, we still want to show the
--- a/publisher/src/components/DataUpload/DataUpload.tsx
+++ b/publisher/src/components/DataUpload/DataUpload.tsx
@@ -25,6 +25,11 @@ import { observer } from "mobx-react-lite";
 import React, { useEffect, useMemo, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 
+import { useStore } from "../../stores";
+import logoImg from "../assets/jc-logo-vector-new.svg";
+import { REPORTS_LOWERCASE } from "../Global/constants";
+import { Logo, LogoContainer } from "../Header";
+import { Loader } from "../Loading";
 import {
   Button,
   DataUploadContainer,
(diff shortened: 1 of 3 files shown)


publisher/src/components/DataUpload/DataUpload.tsx
```diff
@@ -25,11 +25,6 @@ import { observer } from "mobx-react-lite";
 import React, { useEffect, useMemo, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 
-import { useStore } from "../../stores";
```
morden35, line 28: My VSCode is automatically rearranging the order of imports, which is causing a warning when the app is run

morden35:
This is my first pass at frontend/react in a long time! I know this PR will need a ton of work. Please don't hold back on any feedback or suggestions! Thank you!

publisher/src/components/DataUpload/DataUpload.tsx
```diff
@@ -161,13 +161,15 @@ export const DataUpload: React.FC = observer(() => {
...
-        const isSuccessfulMetric =
-          metric.metric_errors.length === 0 || noSheetErrorsFound;
+        // const noSheetErrorsFound =
+        //   metric.metric_errors.filter(
+        //     (sheet) =>
+        //       sheet.messages.filter((msg) => msg.type === "ERROR")?.length > 0
+        //   ).length === 0;
+        // const isSuccessfulMetric =
+        //   metric.metric_errors.length === 0 || noSheetErrorsFound;
+        const isSuccessfulMetric = metric.datapoints.length > 0;
```
mxosman, line 171: Just to clarify - we are using the presence of `datapoints` to determine whether or not a metric is successful (if there are any, it is successful, if there are none - unsuccessful) and no longer the `metric_errors` property, right? If so, this looks great to me! You can delete all the commented out code too.

publisher/src/components/DataUpload/DataUpload.tsx
mxosman: Ah - yes. There is a line in the settings.json of VS code that was in the backend setup that organized imports in a different way on the FE after saving the file - took me a while to figure out what was going on! LMK if you want to disable it unless it is working fine for you in the BE!

One thing you can do after you've finished your edits in the file and save it is run `yarn lint --fix` and it will reorganize the import. Then, I would commit and push. LMK if that doesn't work!

Running `yarn lint` will tell you any other linting errors/warnings, but I think the imports is the only thing here!

publisher/src/components/DataUpload/DataUpload.tsx
lilidworkin: I think we probably _also_ want to make sure that there are no errors, in addition to there being datapoints, right @morden35 ?

lilidworkin (commented):
Cool, nice work diving into the frontend!

As per our convo yesterday though, I think we decided we didn't want to outright error if there are missing metrics, right? So instead of "We encountered 12 errors" can it be like "0 metrics were saved successfully"?

publisher/src/components/DataUpload/DataUpload.tsx
morden35: oh fantastic! Thank you! this did the trick

morden35:
> Cool, nice work diving into the frontend!
> 
> As per our convo yesterday though, I think we decided we didn't want to outright error if there are missing metrics, right? So instead of "We encountered 12 errors" can it be like "0 metrics were saved successfully"?

Good point! I wasn't focusing on the error message at this point (I figured that would be a separate pr). Although I see that this logic is all interconnected. I will try to take a more holistic approach at this today

publisher/src/components/DataUpload/DataUpload.tsx
morden35: hmm I need to think about this more. metric.metric_errors will contain both errors and warnings. in the presence of warnings, data may or may not be saved. in the presence of errors, data will not be saved (i think)

publisher/src/components/DataUpload/DataUpload.tsx
```diff
@@ -161,13 +161,7 @@ export const DataUpload: React.FC = observer(() => {
...
     const errorsWarningsAndSuccessfulMetrics = data.metrics.reduce(
       (acc, metric) => {
-        const noSheetErrorsFound =
-          metric.metric_errors.filter(
-            (sheet) =>
-              sheet.messages.filter((msg) => msg.type === "ERROR")?.length > 0
-          ).length === 0;
-        const isSuccessfulMetric =
-          metric.metric_errors.length === 0 || noSheetErrorsFound;
+        const isSuccessfulMetric = metric.successful_ingest;
```
morden35, line 164: this will be true if len(metric_key_to_successfully_ingested_sheet_names[metric_definition.key]) > 0 see https://github.com/Recidiviz/recidiviz-data/pull/19684/files#diff-d3e180104898a32d7cc8f2a707469df2f23b7c2350b5e6d1946e72c173194ee4R342

publisher/src/components/DataUpload/UploadErrorsWarnings.tsx
```diff
@@ -211,41 +211,33 @@ export const UploadErrorsWarnings: React.FC<UploadErrorsWarningsProps> = ({
...
-            .
+            <BlueText>{successfulMetricsCount}</BlueText> metrics were saved
+            successfully.
+            <br />
           </>
         )}
 
-        {/* Case 2: Has Errors Only */}
-        {errorCount > 0 && successfulMetricsCount === 0 && (
+        {/* Case 2: More than 1 metric was saved */}
```
mxosman, line 224: tiniest nit: instead of "More than 1" - something like "1 or more" since it'll show for just one metric saved.

mxosman (approved):
FE looks great to me, Michelle! Happy to approve but not sure if @lilidworkin wants to give another pass!

publisher/src/components/DataUpload/DataUpload.tsx
lilidworkin: I guess I'm curious if we need this -- why can't the logic be:

`const isSuccessfulMetric = (metric.metric_errors.length === 0 || noSheetErrorsFound) && metric.datapoints.length > 0`

If there are problems with that, just lmk!

lilidworkin: Basically, this is the same as the logic we currently have, but _also_ requires the metric to have datapoints present (i.e. not be missing)

publisher/src/components/DataUpload/UploadErrorsWarnings.tsx
```diff
@@ -211,41 +211,33 @@ export const UploadErrorsWarnings: React.FC<UploadErrorsWarningsProps> = ({
...
           <>
-            <BlueText>{successfulMetricsCount}</BlueText> metric
-            {successfulMetricsCount === 0 || successfulMetricsCount > 1
-              ? "s"
-              : ""}{" "}
-            were saved successfully
-            {errorsWarningsAndSuccessfulMetrics.hasWarnings &&
-              " (with warnings)"}
-            .
+            <BlueText>{successfulMetricsCount}</BlueText> metrics were saved
```
lilidworkin, line 218: Since we know successfulMetricsCount === 0 here, maybe we just say "No metrics were saved successfully."?

publisher/src/components/DataUpload/UploadErrorsWarnings.tsx
```diff
@@ -211,41 +211,33 @@ export const UploadErrorsWarnings: React.FC<UploadErrorsWarningsProps> = ({
   const renderErrorWarningTitle = () => {
     return (
       <>
-        {/* Case 1: Only Warnings (which means all metrics were successfully ingested) */}
-        {errorCount === 0 && successfulMetricsCount > 0 && (
+        {/* Section 1: Show number of metrics successfully ingested */}
```
lilidworkin, line 214: I like how you broke this into "sections", great idea!

publisher/src/components/DataUpload/UploadErrorsWarnings.tsx
```diff
@@ -211,41 +211,33 @@ export const UploadErrorsWarnings: React.FC<UploadErrorsWarningsProps> = ({
...
+            .<br />
           </>
         )}
 
-        {/* Case 3: Has Errors and Successes */}
-        {errorCount > 0 && successfulMetricsCount > 0 && (
+        {/* Section 2: Show number of metrics with errors */}
+        {errorCount > 0 && (
           <>
             We encountered <RedText>{errorCount}</RedText> error
```
lilidworkin, line 239: So this is still counting/showing Missing Metrics as errors. As per our convo yesterday, I think we decided not to do that? Or are you planning to do that in a separate PR? Totally fine if so!

lilidworkin (changes requested):
Left some comments/questions that we should resolve before merging, but really excited to see you so productive on the frontend already @morden35 ! Great work!

publisher/src/components/DataUpload/DataUpload.tsx
morden35: so perhaps one solution is to do
`const isSuccessfulMetric = (noSheetErrorsFound && metric.datapoints.length > 0)`

so we're just concerned with errors (not warnings) and if data has been saved

lilidworkin: Ah I see, interesting. Is that why we have the `|| no sheetErrorsFound` logic, though? Basically it seems the logic is "either there are no errors at all, OR there are but there are none with type ERROR". Curious if @mxosman remembers the reasoning here.

lilidworkin: Ah yea I think we just posted at the same time -- I agree! -- I think the `metric.metric_errors.length === 0` part is redundant, and will always be "trumped" by the second part of that or statement. I think `const isSuccessfulMetric = (noSheetErrorsFound && metric.datapoints.length > 0)` is right.

mxosman: > Ah I see, interesting. Is that why we have the `|| no sheetErrorsFound` logic, though? Basically it seems the logic is "either there are no errors at all, OR there are but there are none with type ERROR". Curious if @mxosman remembers the reasoning here.

Blast from the past! Hard to fully remember - but I think when I wrote this originally, it was checking to see if the `metric.metric_errors` array was empty or if not empty, check in each sheet to see if there were any errors or just warnings (because I believe at the time the `metric.metric_errors` array held both errors and warnings within the sheets and there was a possibility of the `metric.metric_errors` array's length being > 1 and only have all warnings within the sheets).

lilidworkin: Thanks @mxosman ! Just to clarify, I think our confusion (but @morden35 correct me if I'm wrong) is that this statement is doing the following:
1. Checking that the length of metric.metric_errors is 0, OR
2. Checking that metric.metric_errors doesn't contain any true errors

Is it true that the logic/behavior would be the same if we only did check 2, and not check 1? Because there are no cases where 1 is true but 2 is false, right? So the only possibility is that 1 is false and 2 is true, or both are true/false, in which cases it should be sufficient just to look at 2?

Sorry if this is a convoluted/confusing question, haha! I think we're all in agreement in any case.

publisher/src/components/DataUpload/UploadErrorsWarnings.tsx
morden35: planning to do this in a separate pr!

lilidworkin (approved):
Looks perfect @morden35 , thanks for making these changes and sorry for the back-and-forth! Thanks @mxosman  for sanity checking us!

mxosman:
🥳 🥳 🥳 Beautifully done, Michelle!

PR #541, opened 2023-03-29
Bulk Publishing: Replace calls to `datapointsStore.getDatapoints(agencyId)` w/ the updated `get_reports_by_id` endpoint

### LINKED ISSUE #529: [Bulk Publishing] Remove calls to `datapointsStore.getDatapoints(agencyId)`
See [this comment](https://github.com/Recidiviz/justice-counts/pull/507#discussion_r1148459416) for context.

mxosman:
## Description of the change

Calling the `/datapoints` endpoint is too expensive for what we need to power the review pages. The updated `get_reports_by_id` should return a list of datapoints (instead of metrics) for each report provided to it. This way, we only get the relevant datapoints we need for the reports that have been selected by the user and not the entire set of datapoints.

## Related issues

Closes #529

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/BulkActionReview.tsx
+++ b/publisher/src/components/Reports/BulkActionReview.tsx
@@ -15,6 +15,8 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import DatapointsStore from "@justice-counts/common/stores/BaseDatapointsStore";
+import { RawDatapointsByMetric, Report } from "@justice-counts/common/types";
 import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
 import { useLocation, useNavigate, useParams } from "react-router-dom";
@@ -46,7 +48,9 @@ const BulkActionReview = () => {
     undefined
   );
   const [isSuccessModalOpen, setIsSuccessModalOpen] = useState(false);
-  const { reportStore, datapointsStore } = useStore();
+  const { reportStore } = useStore();
+
+  const [datapoints, setDatapoints] = useState<RawDatapointsByMetric>({});
 
   const publishMultipleRecords = async () => {
     await reportStore.updateMultipleReportStatuses(
@@ -69,7 +73,7 @@ const BulkActionReview = () => {
   useEffect(() => {
     const initialize = async () => {
       // probably need better way to get reports metrics
-      await datapointsStore.getDatapoints(agencyId);
+      // await datapointsStore.getDatapoints(agencyId);
       const result = await reportStore.initializeReportSettings(
         agencyId.toString()
       );
@@ -77,6 +81,19 @@ const BulkActionReview = () => {
         setIsLoading(false);
         return setLoadingError(result.message);
       }
+      const reportsWithDatapoints =
+        (await reportStore.getMultipleReportsWithDatapoints(
+          recordsIds,
+          String(agencyId)
+        )) as Report[];
+      const combinedDatapointsFromAllReports = reportsWithDatapoints
+        .map((report) => report.datapoints)
+        .flat();
+      setDatapoints(
+        DatapointsStore.keyRawDatapointsByMetric(
+          combinedDatapointsFromAllReports
+        )
+      );
       setIsLoading(false);
     };
 
@@ -88,7 +105,7 @@ const BulkActionReview = () => {
     document.body.style.overflow = isSuccessModalOpen ? "hidden" : "unset";
   }, [isSuccessModalOpen]);
 
-  if (isLoading || datapointsStore.loading)
+  if (isLoading)
     return (
       <PageWrapper>
         <Loading />
@@ -108,9 +125,7 @@ const BulkActionReview = () => {
     enabledMetrics.length > 0
       ? enabledMetrics.reduce((acc, metric) => {
           const reviewMetric = {
-            datapoints: datapointsStore.rawDatapointsByMetric[
-              metric.key
-            ].filter((dp) => dp.report_id && recordsIds.includes(dp.report_id)),
+            datapoints: datapoints[metric.key],
             display_name: metric.display_name,
             key: metric.key,
             metricHasError: false,
--- a/publisher/src/stores/ReportStore.ts
+++ b/publisher/src/stores/ReportStore.ts
@@ -160,10 +160,10 @@ class ReportStore {
     }
   }
 
-  async getMultipleReports(
+  async getMultipleReportsWithDatapoints(
     reportIDs: number[],
     currentAgencyId: string
-  ): Promise<void | Error> {
+  ): Promise<Report[] | Error | void> {
     try {
       const response = (await this.api.request({
         path: `/api/reports?agency_id=${parseInt(
@@ -180,9 +180,7 @@ class ReportStore {
 
       const reports = (await response.json()) as Report[];
 
-      reports.forEach((report) => {
-        this.storeMetricDetails(report.id, report.metrics);
-      });
+      return reports;
     } catch (error) {
       if (error instanceof Error) return new Error(error.message);
     } finally {
--- a/common/stores/BaseDatapointsStore.ts
+++ b/common/stores/BaseDatapointsStore.ts
@@ -114,7 +114,11 @@ abstract class DatapointsStore {
   }
 
   get rawDatapointsByMetric(): RawDatapointsByMetric {
-    return this.rawDatapoints.reduce((res: RawDatapointsByMetric, dp) => {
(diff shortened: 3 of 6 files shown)


publisher/src/stores/ReportStore.ts
```diff
@@ -180,9 +180,7 @@ class ReportStore {
 
       const reports = (await response.json()) as Report[];
 
-      reports.forEach((report) => {
-        this.storeMetricDetails(report.id, report.metrics);
-      });
```
mxosman, line 185: We don't need to store anything in the store regarding these reports - they should already exist within the store.

lilidworkin (approved):
Looks totally correct to me! Weird about the bug you're running into -- not sure if it's a FE or BE thing -- after looking about both PRs I don't see an obvious candidate. Spend a bit more time on it and then we can revisit tomorrow.

publisher/src/stores/ReportStore.ts
lilidworkin: Totally makes sense!

mxosman:
> Looks totally correct to me! Weird about the bug you're running into -- not sure if it's a FE or BE thing -- after looking about both PRs I don't see an obvious candidate. Spend a bit more time on it and then we can revisit tomorrow.

I _think_ I know what's going on... So, it looks like we're using `enabledMetrics` as a source of truth for all of the metrics via:

```
  const enabledMetrics = reportStore.agencyMetrics.filter(
    (metric) => metric.enabled
  );
```

The trouble with this is -- say we have an Annual report with 1 metric enabled and a month report with 2 metrics enabled. When you select a single report (annual or monthly) -- the left hand side list of metrics shows all of the metrics, and oddly renders the page with blank metrics.

Example Annual Report with Funding as its only enabled metric:
Screenshot 2023-03-30 at 2 13 50 PM
The other two metrics are enabled for monthly reports.

I'm going to first work on the backend of this and incorporate your suggestions and then come back and think of a better way to get a list of `enabledMetrics` relevant to the set of reports a user has selected.

lilidworkin:
@mxosman interesting, I think I understand! Actually, is this why Ivan was doing this:

```
reports.forEach((report) => {
    this.storeMetricDetails(report.id, report.metrics);
});
```

Now that I think about it, we _do_ actually need to do this, because while the store already knows about all reports, it only knows the report overview details -- it doesn't know about the report metrics. Right?

And in fact, it's more complicated than that, because we just changed the backend so that `report.metrics` is no longer populated, only `report.datapoints`. But it looks like we might actually need both to be populated. I think if you 1) change the backend so that `report.metrics` and `report.datapoints` are both populated, and 2) put back this `storeMetricDetails` logic, then it should work -- can you test that?

mxosman:
> @mxosman interesting, I think I understand! Actually, is this why Ivan was doing this:
> 
> ```
> reports.forEach((report) => {
>     this.storeMetricDetails(report.id, report.metrics);
> });
> ```
> 
> Now that I think about it, we _do_ actually need to do this, because while the store already knows about all reports, it only knows the report overview details -- it doesn't know about the report metrics. Right?
> 
> And in fact, it's more complicated than that, because we just changed the backend so that `report.metrics` is no longer populated, only `report.datapoints`. But it looks like we might actually need both to be populated. I think if you 1) change the backend so that `report.metrics` and `report.datapoints` are both populated, and 2) put back this `storeMetricDetails` logic, then it should work -- can you test that?

Yeah, I think you're right... bummer! I actually wrote and removed that snippet above. The interesting thing is... we are getting a list of the metrics via the `/metrics` endpoint and storing them in the ReportStore with this function `initializeReportSettings`. So we might be OK not needing to update the endpoint to send the list of metrics. One thing I might try is to filter the list of enabled reports further to exclude metrics w/ frequencies that don't match the list of reports the user selected? Might not be pretty, but will at least solve the left hand-side list of metrics showing all enabled metrics regardless of whether or not you selected only monthly reports/only selected annual reports. If you select a monthly + annual report to publish, then it should show you all the enabled metrics. Does that make sense?

Let me try a few things first before readjusting the backend to also send metrics!

lilidworkin:
Hmm @mxosman that sort of makes sense but I also feel like I'm getting confused here. In order to determine what metrics show up on the left side, don't we want to just check what metrics appear in the list of datapoints that the BE has sent over? For instance, if `get_reports_by_ids` returns something that looks like this:

```
[
     {
          "report 1",
          "datapoints": [ {"funding": 100} ]
     },
     {
          "report 2",
          "datapoints": [ {"arrests": 5} ]
     },
     {},   
]
```

Then we should show funding and arrests in the left side bar. So I guess I'm confused about why we need to reference anything else that already exists in the store. But I think I'm probably missing something.

mxosman:
> Hmm @mxosman that sort of makes sense but I also feel like I'm getting confused here. In order to determine what metrics show up on the left side, don't we want to just check what metrics appear in the list of datapoints that the BE has sent over? For instance, if `get_reports_by_ids` returns something that looks like this:
> 
> ```
> [
>      {
>           "report 1",
>           "datapoints": [ {"funding": 100} ]
>      },
>      {
>           "report 2",
>           "datapoints": [ {"arrests": 5} ]
>      },
>      {},   
> ]
> ```
> 
> Then we should show funding and arrests in the left side bar. So I guess I'm confused about why we need to reference anything else that already exists in the store. But I think I'm probably missing something.

Sorry, I didn't do a great job of explaining -

```
  const enabledMetrics = reportStore.agencyMetrics.filter(
    (metric) => metric.enabled
  );
```

This snippet is currently what's being used to render that metrics list -- it's taking the entire list of metrics and filtering out the disabled ones. We need to replace this with something that will just show the list of enabled metrics for the reports that were selected because this will not help us unless we do further filtering that's a little wonky.

I just came to the same conclusion as you! I think it makes sense to loop through all the datapoints and create a set of the metrics provided by the datapoints.

mxosman:
This might be tricky because we're receiving the full list of datapoints including disabled metrics. I guess we can have the FE go through and give us a list of enabledMetricKeys and use that to filter the datapoints. But, now I'm wondering if it's just easier to update the backend to send a list of metrics -- really all we need from a list of metrics (now that we have the datapoints) are a list of the enabled ones - we get everything else from the datapoints.

mxosman:
> This might be tricky because we're receiving the full list of datapoints including disabled metrics. I guess we can have the FE go through and give us a list of enabledMetricKeys and use that to filter the datapoints. But, now I'm wondering if it's just easier to update the backend to send a list of metrics -- really all we need from a list of metrics (now that we have the datapoints) are a list of the enabled ones - we get everything else from the datapoints.

Annual + Monthly Report:
Screenshot 2023-03-31 at 10 41 57 AM

Datapoints for both reports:
Screenshot 2023-03-31 at 10 42 02 AM

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -202,11 +205,19 @@ export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
...
       </Summary>
-      <MetricsPanel>
-        {metrics.map((metric) => {
-          return renderSection(metric);
-        })}
-      </MetricsPanel>
+
+      {hasNoDatapoints ? (
+        <NoDatapointsMessage>
+          There are no metric data to review. Please go back and add metric data
```
lilidworkin, line 211: nit, how about "There is no data to review. Please go back and enter data in order to Review & Publish."

publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -99,26 +117,44 @@ const BulkActionReview = () => {
...
+  let currentSystemKey;
+  const enabledMetricKeys = reportStore.agencyMetrics.reduce((keys, metric) => {
+    if (metric.enabled) {
+      currentSystemKey = metric.system.key;
+      keys.push(metric.key);
+    }
+    return keys;
+  }, [] as string[]);
+
+  const metricsToDisplay = Object.entries(datapoints).reduce(
```
lilidworkin, line 129: Nice! So we're basically taking an intersection of enabled metrics and the metrics that appear in the list of datapoints. Seems perfect!

lilidworkin (approved):
Your solution looks great!

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: I love that so much! Thank you!

publisher/src/components/Reports/BulkActionReview.tsx
mxosman: Exactly!

PR #561, opened 2023-04-05
Bulk Publishing: Error handling & review page refresh fix

mxosman:
## Description of the change

Goal: add basic error handling when publishing is unsuccessful and handle user refresh in the BulkActionReview and DataEntryReview pages

* Adds basic error toast when the bulk publishing is not successful.
Screenshot 2023-04-05 at 4 51 34 PM

* Previously, a refresh in the BulkActionReview and DataEntryReview crashed the app. Now, the BulkActionReview can handle a refresh and still persist data. The DataEntryReview will redirect a user back to the data entry form on refresh. I'm thinking there should be a way to persist this too and spent time trying to figure out, but kept coming up empty. I'll keep digging - but for now, we can redirect users back to the data entry form and avoid the crash behavior.
  * Update on persisting the data entry review: 
    * There's no way to do this without making another call to the backend. When a user goes through the Records Overview -> ReportDataEntry > DataEntryForm there is a call to `getReports` in `ReportDataEntry` which gets a report + its metrics and stores it - so by the time the review page looks for report metrics, it's ready and available in the store. When you refresh in the Data Entry review page, the component that called `getReports` doesn't remount and thus nothing is available in the store.

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/BulkActionReview.tsx
+++ b/publisher/src/components/Reports/BulkActionReview.tsx
@@ -73,11 +73,21 @@ const BulkActionReview = () => {
   };
 
   const unpublishMultipleRecords = async () => {
-    await reportStore.updateMultipleReportStatuses(
+    const response = (await reportStore.updateMultipleReportStatuses(
       recordsIds,
       agencyIdString,
       "DRAFT"
-    );
+    )) as Response;
+
+    if (response.status !== 200) {
+      showToast({
+        message: `Failed to publish. Please try again.`,
+        color: "red",
+        timeout: 3000,
+      });
+      return;
+    }
+
     setIsSuccessModalOpen(true);
   };
 
--- a/publisher/src/components/Reports/BulkActionReview.tsx
+++ b/publisher/src/components/Reports/BulkActionReview.tsx
@@ -15,6 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { showToast } from "@justice-counts/common/components/Toast";
 import DatapointsStore from "@justice-counts/common/stores/BaseDatapointsStore";
 import { RawDatapointsByMetric, Report } from "@justice-counts/common/types";
 import { observer } from "mobx-react-lite";
@@ -54,11 +55,20 @@ const BulkActionReview = () => {
   const [datapoints, setDatapoints] = useState<RawDatapointsByMetric>({});
 
   const publishMultipleRecords = async () => {
-    await reportStore.updateMultipleReportStatuses(
+    const response = (await reportStore.updateMultipleReportStatuses(
       recordsIds,
       agencyIdString,
       "PUBLISHED"
-    );
+    )) as Response;
+
+    if (response.status !== 200) {
+      showToast({
+        message: `Failed to publish. Please try again.`,
+        color: "red",
+        timeout: 3000,
+      });
+      return;
+    }
     setIsSuccessModalOpen(true);
   };
 
--- a/publisher/src/components/Reports/BulkActionReview.tsx
+++ b/publisher/src/components/Reports/BulkActionReview.tsx
@@ -17,7 +17,11 @@
 
 import { showToast } from "@justice-counts/common/components/Toast";
 import DatapointsStore from "@justice-counts/common/stores/BaseDatapointsStore";
-import { RawDatapointsByMetric, Report } from "@justice-counts/common/types";
+import {
+  RawDatapointsByMetric,
+  Report,
+  ReportOverview,
+} from "@justice-counts/common/types";
 import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
 import { useLocation, useNavigate, useParams } from "react-router-dom";
@@ -53,6 +57,7 @@ const BulkActionReview = () => {
   const { reportStore } = useStore();
 
   const [datapoints, setDatapoints] = useState<RawDatapointsByMetric>({});
+  const [records, setRecords] = useState<ReportOverview[]>([]);
 
   const publishMultipleRecords = async () => {
     const response = (await reportStore.updateMultipleReportStatuses(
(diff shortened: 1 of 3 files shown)


publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -54,20 +55,39 @@ const BulkActionReview = () => {
...
+    )) as Response;
+
+    if (response.status !== 200) {
+      showToast({
+        message: `Failed to publish. Please try again.`,
+        color: "red",
+        timeout: 3000,
+      });
+      return;
+    }
```
mxosman, line 71: I recognize it's not ideal that we have `publishMultipleRecords` in two places, but centralizing them would require a slightly deeper refactor because of how we're passing these functions (and their state updates) via the `button` property's `onClick`. Happy to consolidate in a follow up PR!

publisher/src/components/Reports/DataEntryReview.tsx
```diff
@@ -189,10 +204,10 @@ const DataEntryReview = () => {
...
 
   // modal props
-  const systemSearchParam = metricsPreview
-    ? metricsPreview[0].system.key
-    : undefined;
-  const metricSearchParam = metricsPreview ? metricsPreview[0].key : undefined;
+  const systemSearchParam =
+    enabledMetricKeys.length > 0 ? currentSystemKey : undefined;
+  const metricSearchParam =
+    enabledMetricKeys.length > 0 ? enabledMetricKeys[0] : undefined;
```
mxosman, line 210: I didn't catch this before, but I realized that these review component props don't match the way we get, filter and reduce metrics in the bulk action flow which is causing the same issues we had before with the review page rendering blank spaces for metrics that aren't enabled.

publisher/src/components/Reports/BulkActionReview.tsx
lilidworkin: Cool, thanks for catching this / calling it out! I think it's probably ok since the function is relatively simple, but up to you if you want to tackle in a follow-up.

publisher/src/components/Reports/DataEntryReview.tsx
```diff
@@ -96,6 +91,7 @@ const DataEntryReview = () => {
   useEffect(() => {
     const initialize = async () => {
       formStore.validatePreviouslySavedInputs(reportID);
+      await reportStore.initializeReportSettings(String(agencyId));
```
lilidworkin, line 94: Hold on, so this is making another request to the backend, right? Let's be really careful whenever we add another request given performance issues. Why do we need this?

publisher/src/components/Reports/DataEntryReview.tsx
lilidworkin: > these review component props don't match the way we get, filter and reduce metrics in the bulk action flow

Right so this makes me think there might be an opportunity for consolidation of some sort, so that we share this logic between all three flows?

mxosman: This particular way only needs to be shared between Bulk Action Review and Data Entry page. I thought the way we got metrics in the data entry page wouldn't produce the weird rendering behavior in the review page, but as I was going through different agencies testing the Data Entry -> Publish flow, I noticed the empty rendering of metrics pop up in some records. Matching the logic of the Bulk Action flow fixed that weird behavior throughout all single records -> review I tested.

I can work on a way to consolidate this - it might be a bit tricky, but I'll give it a shot.

lilidworkin: > I thought the way we got metrics in the data entry page wouldn't produce the weird rendering behavior in the review page, but as I was going through different agencies testing the Data Entry -> Publish flow, I noticed the empty rendering of metrics pop up in some records. Matching the logic of the Bulk Action flow fixed that weird behavior throughout all single records -> review I tested.

This makes me feel like we don't quite understand what's going on yet, which makes me nervous. Can you take some time to look into this and figure out why this is the case? It's totally worth spending some time investigating. I'd much rather you have a solid, deep understanding of all the different data formats than we put a bandaid on top of something! And it's totally fine if it takes awhile, there's really no rush!

publisher/src/components/Reports/DataEntryReview.tsx
mxosman: If we're going to use the same method of getting metrics as the Bulk Action review (using `reportStore.agencyMetrics`), then we'll need to call this function to populate the `metricsBySystem` property in the reportStore. I can work on another way to do this.

```
    const { metrics, isPublishable: publishable } =
      formStore.validateAndGetAllMetricFormValues(reportID);

    const enabledMetrics = metrics.filter((metric) => metric.enabled);
``` 

Prior to this PR, we obtained the enabled metrics from the above function and it doesn't always work consistently (this way of getting and filtering metrics used to be in the Bulk Action review page - which caused the inconsistent renderings - and was fixed by using `initializeReportSettings` + `reportStore.agencyMetrics` to get the metrics).

I'll look deeper into pinpointing why `metrics.filter((metric) => metric.enabled)` causes inconsistencies and see what other solutions I can come up.

lilidworkin: And meanwhile if you can also think through, for each of these backend calls, what information it is returning, and whether we need all of that information, or just a part of the information, and whether that information could possibly be gotten some other way?

mxosman:
Adjusting the scope of this PR now to just include the error handling, and handling of refresh in the review pages (bulk publish review and data entry review).

The bulk publish review page can persist the information after refreshing. The data entry was the trickiest for some reason - will need to keep digging into that one to get it to persist, but right now the solution is to navigate back to the data entry form when a user refreshes (previous behavior navigated a user back to the home page). 

I still need to clean up and undo some of the changes in the data entry to stay within the new scope of this PR, so I'll move this to draft for now and pick up where I left off later.

PR #568, opened 2023-04-10
[Bug Bash] Data Page: Show the first enabled metric for an agency on load and handle no enabled metrics case

### LINKED ISSUE #523: [Frontend] Data: the first set of metrics for an agency that have enabled metrics should appear uncollapsed on load

mxosman:
## Description of the change

The Data page currently shows a chart for the first metric (regardless of whether or not it is enabled/disabled) when a user loads the page. It also does not currently handle a case where no metrics are enabled. This PR allows a user to see the first _enabled_ metric on load, and when no metrics are enabled, the user sees a message with a link to the Metric Configuration page. 

Currently, when there are no metrics enabled, the user just sees a loader that never resolves. Does this new message screen feel jarring or does it feel like it works within this context? Open to all thoughts!

Before:

https://user-images.githubusercontent.com/59492998/231017800-92ea5000-bcad-44ca-a51c-09e0bc247c62.mov

After:

https://user-images.githubusercontent.com/59492998/231018789-fbcb4410-919d-4b7c-9db7-b4031a078868.mov

## Related issues

Closes #523 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricConfiguration/MetricsView.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricsView.tsx
@@ -219,7 +219,7 @@ export const MetricsView: React.FC = observer(() => {
                       </SystemName>
                     </SystemNameContainer>
                   ) : (
-                    <SystemNameContainer isSystemActive={false}>
+                    <SystemNameContainer isSystemActive>
                       <SystemName>
                         {metrics[0].system.display_name} (No enabled metrics)
                       </SystemName>
--- a/publisher/src/components/MetricConfiguration/MetricsView.tsx
+++ b/publisher/src/components/MetricConfiguration/MetricsView.tsx
@@ -62,7 +62,6 @@ import {
   PanelRightTopButtonsContainer,
   SystemName,
   SystemNameContainer,
-  SystemNamePlusSign,
   SystemsContainer,
 } from ".";
 
@@ -92,12 +91,13 @@ export const MetricsView: React.FC = observer(() => {
       return setLoadingError(result.message);
     }
 
-    const defaultSystemSearchParam = Object.keys(result)[0]
+    const firstEnabledMetric = Object.values(result)
+      ?.flat()
+      .find((metric) => metric.enabled);
+    const defaultSystemSearchParam = firstEnabledMetric?.system.key
       .toLowerCase()
       .replace(" ", "_") as AgencySystems;
-    const defaultMetricSearchParam = Object.values(result)[0]?.find(
-      (metric) => metric.enabled
-    )?.key;
+    const defaultMetricSearchParam = firstEnabledMetric?.key;
 
     // same logic as in metric config page, the only difference is
     // there should always be metric search param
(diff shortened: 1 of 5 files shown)


publisher/src/components/MetricConfiguration/MetricsView.tsx
```diff
@@ -166,6 +169,22 @@ export const MetricsView: React.FC = observer(() => {
...
+      <NoEnabledMetricsMessage>
+        There are no enabled metrics to view. Please go to{" "}
+        <DisclaimerLink
+          onClick={() => {
+            navigate("../settings/metric-config");
+          }}
+        >
+          Metric Configuration
+        </DisclaimerLink>{" "}
+        to enable a metric.
```
mxosman, line 183: Always open to better copy suggestions!

lilidworkin: This looks great to me!

lilidworkin (commented):
@mxosman when an agency has multiple systems, and there are enabled metrics in each system, can we uncollapse any system with enabled metrics? Currently I see this
Screenshot 2023-04-11 at 2 57 17 PM
which makes it looks like Jails has no enabled metrics, but it actually does. 

Also, the plus sign to expand the section doesn't appear until you hover over it - maybe it should always appear?
Screenshot 2023-04-11 at 2 57 57 PM

mxosman:
> Do we anticipate running into the case where an agency uploads data for a metric, and then later on disables the metric? Probably unlikely, but I feel like in this event we might want to still show the data page for the disabled metric. If this is a concern, one solution could be showing metrics on the Data Page that actually have data, regardless of their enabled status.

Yes, I think that's definitely possible. If I'm understanding correctly - I think if an agency uploads data for a metric and that metric is enabled, they should see it in the Data page. If they later decide to disable the metric, it won't appear on the list (this was the regular behavior prior to this adjustment). I do like that solution if we want to change the behavior from only showing enabled metrics to only showing metrics with data! It does feel like a separate undertaking though that will take a bit more thought - but what you are saying totally makes sense!

mxosman:
> @mxosman when an agency has multiple systems, and there are enabled metrics in each system, can we uncollapse any system with enabled metrics?

This ended up being a little trickier than I thought. After uncollapsing all the systems with enabled metrics, I realized the chart stretched vertically when there were a lot of enabled metrics on the list (I understand now the past decision to have an accordion style list of systems & metrics [one system open at a time]) - I made an adjustment to have the chart's max height be the viewport height.

I also made an adjustment to the logic that was displaying the first system & first enabled metric. It doesn't work well when there are multiple systems and the first system has no enabled metrics (e.g. a user w/ Prisons & Supervision systems -- Prisons is the first system in the list and has no enabled metrics, Supervision is the second system in the list and has enabled metrics -- the user ends up seeing "there are no enabled metrics" screen since the prev. logic is only grabbing the first system and its first enabled metrics). Now, we loop through all of the systems & their metrics and find the first enabled metric and display that on load. LMK if I'm not making sense here - happy to clarify further!


https://user-images.githubusercontent.com/59492998/231511540-e4e586d4-cd66-4089-b1c7-0b6b549cd0e7.mov


---
> Also, the plus sign to expand the section doesn't appear until you hover over it - maybe it should always appear?

After digging into this - I realized we don't actually need the + and hover cursor on the system names if everything is going to be uncollapsed when there are enabled metrics (and when there are no enabled metrics, there's nothing to uncollapse) -- so I removed the notion of the "+" sign and hover effect.

PR #577, opened 2023-04-19
Bulk Publishing: Refactor for UX, code reusability and performance

### LINKED ISSUE #510: Review Page Refresh Follow-Ups
Not at all urgent, but some P1s to consider addressing at the end when everything else is done. These are based on Humphrey's comment [here](https://github.com/Recidiviz/justice-counts/pull/507#pullrequestreview-1344693763).

### Long Metric Names

> Need a better treatment for metrics that have long names – at the very least, the full name should be visible on hover (via tooltip? by hiding the date labels? etc)

This is easiest to repro on a Jails agency:

Ivan says this is hard to do because of how HTML tables work. At the least, can we just have a prettier tooltip? Feel free to ask Juan if you want design for this, but I would suggest just coming up with something yourself based on our other components!

### Hiding metrics/breakdowns with no data

> If there’s no data for a metric, it would be better to have a collapsed empty state for the section, rather than take up more space on the page with empty data.

> Same thing with breakdowns with no data; I would like to collapse these if there’s nothing relevant. I feel this pain especially with the Race/Ethnicity numbers.

Ivan needed more clarification before he was able to address these, but I think you can probably figure something out on your own. Feel free to ask Juan if you want design, but I would just do whatever is easiest. i don't think we need true expand/collapse functionality (which is what Ivan was getting hung up on, I think) -- if there's no data, we can just hide the rows.

### LINKED ISSUE #578: [Frontend] Bulk Publishing: Refactor for UX, code reusability and performance
Reference design doc: https://paper.dropbox.com/doc/Design-Doc-Bulk-Publishing--B2qEY02vPAJ8BF~f6rWvH4Y4Ag-QTpFOEugc7BteE7O14DJR

mxosman:
## Description of the change

Refactoring the `BulkActionReview` & `DataEntryReview` components to:
* show all metrics that have datapoints in the review page instead of only enabled metrics
* use the datapoints as the source of truth for the list of metrics and everything else needed by the review components
  * this helps efforts to improve performance by minimizing excess calls to the backend
* consolidate similar logic
* render a consistent review page across all review components
* handle no datapoints consistently

Still left to figure out:
* handling Supervision agency with subsystems that have datapoints and the overall Supervision system is combined (it shows the subsystems' data in the review page - maybe this is the right behavior!)

Demo:

https://user-images.githubusercontent.com/59492998/233157121-0e7f5e0f-02ae-443d-a81c-6e0e17275d77.mov

## Related issues

Closes #578
Closes #510

### Development

### Code review

### DIFF
--- a/common/components/DataViz/DatapointsTableView.tsx
+++ b/common/components/DataViz/DatapointsTableView.tsx
@@ -24,6 +24,7 @@ import {
 } from "../../types";
 import {
   DatapointsMetricNameCell,
+  DatapointsMetricNameCellTooltip,
   DatapointsTableBottomBorder,
   DatapointsTableContainer,
   DatapointsTableDetailsCell,
@@ -151,6 +152,9 @@ export const DatapointsTableView: React.FC<{
                 <DatapointsTableNamesRow>
                   <DatapointsMetricNameCell title={metricName}>
                     {metricName}
+                    <DatapointsMetricNameCellTooltip>
+                      {metricName}
+                    </DatapointsMetricNameCellTooltip>
                   </DatapointsMetricNameCell>
                 </DatapointsTableNamesRow>
               )}
--- a/publisher/src/components/Reports/BulkActionReview.tsx
+++ b/publisher/src/components/Reports/BulkActionReview.tsx
@@ -32,6 +32,7 @@ import { PageWrapper } from "../Forms";
 import { REPORT_LOWERCASE, REPORTS_LOWERCASE } from "../Global/constants";
 import { Loading } from "../Loading";
 import {
+  PublishReviewPropsFromDatapoints,
   ReviewHeaderActionButton,
   ReviewMetric,
   ReviewMetrics,
@@ -56,8 +57,20 @@ const BulkActionReview = () => {
   const [isSuccessModalOpen, setIsSuccessModalOpen] = useState(false);
   const { reportStore } = useStore();
 
-  const [datapoints, setDatapoints] = useState<RawDatapointsByMetric>({});
-  const [records, setRecords] = useState<ReportOverview[]>([]);
+  // const [datapoints, setDatapoints] = useState<RawDatapointsByMetric>({});
+  const [publishReviewProps, setPublishReviewProps] =
+    useState<PublishReviewPropsFromDatapoints>(
+      {} as PublishReviewPropsFromDatapoints
+    );
+  // const [records, setRecords] = useState<ReportOverview[]>([]);
+
+  const {
+    reports,
+    datapointsByMetric,
+    metricsToDisplay,
+    firstSystemKey,
+    firstMetricKey,
+  } = publishReviewProps;
 
   const publishMultipleRecords = async () => {
     const response = (await reportStore.updateMultipleReportStatuses(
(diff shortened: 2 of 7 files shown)


publisher/src/stores/ReportStore.ts
```diff
@@ -190,6 +192,54 @@ class ReportStore {
...
+        reports: reportsWithDatapoints,
+        datapointsByMetric,
+        metricsToDisplay,
+        firstSystemKey,
+        firstMetricKey,
+      };
+    } catch (error) {
+      if (error instanceof Error) return new Error(error.message);
+    }
+  }
```
mxosman, line 241: My consolidation of the shared logic was to create this method within the ReportStore to `getMultipleReportsWithDatapoints` and return all of the necessary props the DataEntryReview and BulkActionReview page need to render the Review page.

common/components/DataViz/DatapointsTableView.tsx
```diff
@@ -151,6 +152,9 @@ export const DatapointsTableView: React.FC<{
                 <DatapointsTableNamesRow>
                   <DatapointsMetricNameCell title={metricName}>
                     {metricName}
+                    <DatapointsMetricNameCellTooltip>
+                      {metricName}
+                    </DatapointsMetricNameCellTooltip>
```
mxosman, line 157: These two `DatapointsTableView` files are quick side missions - they add a tooltip to all of the metric name cells on hover so the long names can still be read. LMK if it should _only_ be for metrics with long names - it's a bit more complex as I'd have to store each element and compare their `offsetWidth` and `scrollWidth` to determine whether or not they are overflowing - but certainly doable.

I can also separate this out to a different PR if that's preferable!

https://user-images.githubusercontent.com/59492998/233169603-b2e40dd8-59e4-4fad-92b6-0589fc3197f7.mov

publisher/src/stores/ReportStore.ts
```diff
@@ -190,6 +192,54 @@ class ReportStore {
...
+      const datapointsEntries = Object.entries(datapointsByMetric);
+      const metricsToDisplay = datapointsEntries.map(
+        ([metricKey, metricDatapoints]) => {
+          return {
+            key: metricKey,
+            displayName: metricDatapoints[0].metric_display_name as string,
+          };
+        }
+      );
+      const firstSystemKey = datapointsEntries[0][0].split("_")[0]; // get first system key via splitting a datapoint's metric key
```
mxosman, line 228: This won't work for all systems... - I'll think of a better way of getting the first system key.

common/components/DataViz/DatapointsTableView.tsx
lilidworkin: Fantastic! Fine to slip this into this PR -- thanks for calling it out in a comment though. Also I think it's totally fine to show the tooltip in all cases (but thanks for asking about that)!

publisher/src/components/Reports/BulkActionReview.tsx
```diff
@@ -142,35 +132,12 @@ const BulkActionReview = () => {
...
-    [] as { key: string; displayName: string }[]
-  );
-
   // review component props
   const metrics =
     metricsToDisplay.length > 0
       ? metricsToDisplay.reduce((acc, metric) => {
           const reviewMetric = {
-            datapoints: datapoints[metric.key],
+            datapoints: datapointsByMetric[metric.key],
```
lilidworkin, line 140: Love this renaming

publisher/src/stores/ReportStore.ts
```diff
@@ -190,6 +192,57 @@ class ReportStore {
...
+      const datapointsEntries = Object.entries(datapointsByMetric);
+      const metricsToDisplay = datapointsEntries.map(
+        ([metricKey, metricDatapoints]) => {
+          return {
+            key: metricKey,
+            displayName: metricDatapoints[0].metric_display_name as string,
+          };
+        }
+      );
+      const firstSystemKey = this.userStore.userAgencies?.find(
```
lilidworkin, line 229: Can you remind me what these two variables are doing and why we need them? I'm sure it's fine, they just feel out of place compared to what else is going on in this method, so maybe a comment would be helpful

lilidworkin (approved):
@mxosman -- I am OBSESSED with this refactor. You handled this BEAUTIFULLY. I literally have no comments or suggestions! Thank you so much! How does it feel to you??

I'm not worried at all about the edge case with Supervision -- I think you're right that this behavior might even be desirable -- but in any case it doesn't seem worth worrying about right now, if it ever comes up in the future we can address!

publisher/src/stores/ReportStore.ts
mxosman: Absolutely - this is not clear at all - will definitely leave a comment! This is for the `Go to Data` button in the modal that appears after you successfully publish - that way when you click it, you're lead to the data viz page w/ the system and metric key params in the URL.

mxosman:
> @mxosman -- I am OBSESSED with this refactor. You handled this BEAUTIFULLY. I literally have no comments or suggestions! Thank you so much! How does it feel to you??


You're too kind! Thank you so much for suggesting this in the first place and for the very thoughtfully proposed solution! I absolutely love it - it was a **delightful** clean up and refactor (seeing everything consolidated is always satisfying) - and **such a great solution**! I truly appreciate all your guidance with this one! 😄 

> I'm not worried at all about the edge case with Supervision -- I think you're right that this behavior might even be desirable -- but in any case it doesn't seem worth worrying about right now, if it ever comes up in the future we can address!

Sounds great! Thank you for the confirmation!

publisher/src/stores/ReportStore.ts
lilidworkin: Oh great!

PR #580, opened 2023-04-20
Guidance: Gate entire backend call to non-JC admins

mxosman:
## Description of the change

Gates the entire backend call for non-JC admins as opposed to overwriting the completion boolean on each topic. 

* Non-JC-admins will not go through the guidance flow and the app will not make a call to the guidance endpoint
* JC-admins will go through the guidance flow based on their guidance status in the DB

### Development

### Code review

### DIFF
--- a/publisher/src/stores/GuidanceStore.ts
+++ b/publisher/src/stores/GuidanceStore.ts
@@ -93,7 +93,7 @@ class GuidanceStore {
   getOnboardingTopicsStatuses = async (
     agencyId: string
   ): Promise<OnboardingTopicsStatuses[]> => {
-    if (this.userStore.isJusticeCountsAdmin(agencyId)) {
+    if (!this.userStore.isJusticeCountsAdmin(agencyId)) {
       /**
        * NOTE:
        * This gates the guidance flow for users who are not JC Admins.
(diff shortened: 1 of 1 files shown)


publisher/src/stores/GuidanceStore.ts
```diff
@@ -93,6 +93,24 @@ class GuidanceStore {
...
+       * TODO(#496) ungate guidance flow for all users
+       */
+
+      runInAction(() => {
+        this.onboardingTopicsStatuses = {
+          WELCOME: true,
+        };
+      });
+      return [];
+    }
```
mxosman, line 112: This was the simplest way to accomplish this - we essentially short-circuit this function and add a simple 1-topic-completed object to `onboardingTopicsStatuses`. `hasCompletedOnboarding` will check this list and return `true` if all topics are `true`.

lilidworkin: Perfect, thanks for the explanation! This makes total sense to me.

I think hopefully based on the results of the conversation yesterday, we might be able to stop storing guidance state in the DB entirely, in which case we'll be able to remove this call for JC admins too. But let's see how that convo evolves. Curious if you have thoughts on it -- I know it might be more work in the short-term to refactor, but ultimately I think it will save us a lot of debugging time down the line.

lilidworkin (approved):
TY!

publisher/src/stores/GuidanceStore.ts
mxosman: Great convo yesterday and a lot of great ideas and points! I think it would be nice to not have to rely on storing this state in the DB and figuring out ways to have the FE infer everything - it's a lot of calls esp. after you're long done with Guidance... I honestly have no problem if we realize we have throw away most/all of this logic and start from scratch!

PR #601, opened 2023-05-01
[Bug] Clicking on Data link while in the Data page does not reload the page

mxosman:
## Description of the change

**Issue**: when you are on the `Data` page and you click on the `Data` link, you're redirected to the `Data` page with no query params, and will encounter the "There are no enabled metrics to view. Please go to Metric Configuration to enable a metric."

This is more/less a temporary fix - ~I need to dig deeper to figure out why things don't appear to be reinitializing~ (see below conversation). We're basically taking no action if a user is already on the page, and clicking on the link - which might be good to do anyway because doing a reinitialization seems excessive and not entirely necessary. Open to thoughts!

https://user-images.githubusercontent.com/59492998/235522899-898981cb-3773-434e-835a-4ead01e782ee.mov

### Development

### Code review

### DIFF
--- a/publisher/src/components/Menu/Menu.tsx
+++ b/publisher/src/components/Menu/Menu.tsx
@@ -276,7 +276,7 @@ const Menu: React.FC = () => {
         {/* Data (Visualizations) */}
         <MenuItem
           onClick={() => {
-            navigate("data");
+            if (pathWithoutAgency !== "data") navigate("data");
             handleCloseMobileMenu();
           }}
           active={pathWithoutAgency === "data"}


nichelle-hall (approved):
LGTM!

lilidworkin (approved):
Nice catch! This seems like a fine fix! It might still be good to investigate why the re-initialization wasn't happening anyway, just for our own understanding / knowledge. But I agree it would actually be a waste in this case!

mxosman:
> Nice catch! This seems like a fine fix! It might still be good to investigate why the re-initialization wasn't happening anyway, just for our own understanding / knowledge. But I agree it would actually be a waste in this case!

Thank you both!

For sure! This is happening in this `useEffect`:

```
  useEffect(() => {
    datapointsStore.resetState();
      setIsLoading(true);
      await datapointsStore.getDatapoints(Number(agencyId));
      await initDataPageMetrics();
    };

    initialize();
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [agencyId]);
```

So, this will fire off when the component mounts and only again when a user has switched agencies. Since you're already in the `/data` page, clicking on it again just strips the URL params - the component is still mounted, but there is no reinitialization. 

One way to solve for this is to import the `useLocation` hook and check to see if there are any search params. We can store this check in a variable and added it to the dependency array - that way if there are no params, this `useEffect` will be called again.

PR #603, opened 2023-05-02
Bulk Publish Review: Adjust validation behavior

### LINKED ISSUE #514: [Frontend] Relax validation for report publishing

mxosman:
## Description of the change

> Description here

## Related issues

Closes #514

### Development

### Code review

### DIFF
--- a/publisher/src/stores/ReportStore.ts
+++ b/publisher/src/stores/ReportStore.ts
@@ -254,7 +254,6 @@ class ReportStore {
           return {
             key: metricKey,
             displayName: metricDatapoints[0].metric_display_name as string,
-            metricErrors,
           };
         }
       );
@@ -263,6 +262,7 @@ class ReportStore {
         records: reportsWithDatapoints,
         datapointsByMetric,
         metricsToDisplay,
+        metricErrors,
       };
     } catch (error) {
       if (error instanceof Error) return new Error(error.message);
--- a/publisher/src/stores/ReportStore.ts
+++ b/publisher/src/stores/ReportStore.ts
@@ -31,7 +31,10 @@ import {
   REPORTS_LOWERCASE,
 } from "../components/Global/constants";
 import { MetricSettings } from "../components/MetricConfiguration";
-import { PublishReviewPropsFromDatapoints } from "../components/ReviewMetrics";
+import {
+  PublishReviewMetricErrors,
+  PublishReviewPropsFromDatapoints,
+} from "../components/ReviewMetrics";
 import { groupBy } from "../utils";
 import API from "./API";
 import DatapointsStore from "./DatapointsStore";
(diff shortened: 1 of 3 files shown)


publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -114,6 +114,12 @@ export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
                 buttonColor={buttonColor}
                 labelColor={labelColor}
                 borderColor={borderColor}
+                showTooltip={disabled && name === "Publish"}
```
lilidworkin, line 117: What else is it possible for this button to be named? This is a super nit, but I wonder if it is better practice to pass a flag indicating whether this is the Publish version of the button, rather than relying on the name. That way if we change the name down the line to, say "Publish your Data" or whatever, the logic won't be affected.

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -114,6 +114,12 @@ export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
...
+                tooltipMsg={
+                  disabled && name === "Publish"
+                    ? "There are errors in your data that need to be addressed before publishing. Please contact the Justice Counts team at support@justice-counts.com if you need help."
```
lilidworkin, line 120: Oops I misspoke before, I think the email address is actually `justice-counts-support@csg.org`! Otherwise this message looks great.

publisher/src/stores/ReportStore.ts
```diff
@@ -212,11 +215,50 @@ class ReportStore {
...
       const combinedFilteredDatapointsFromAllReports = reportsWithDatapoints
         ?.map((report) => report.datapoints)
-        .flat()
-        .filter((dp) => dp.value !== null);
-      const datapointsByMetric = DatapointsStore.keyRawDatapointsByMetric(
-        combinedFilteredDatapointsFromAllReports
-      );
+        .flat();
+
+      /** Find metric errors from datapoints (non-numeric characters and breakdowns with values but no value for the top-level metric)  */
```
lilidworkin, line 220: Great comments!

Since we're doing this validation in two places now (one on the Metric objects for publishing a single report, and one on the Datapoint objects for Bulk Publish) can we add a comment in both places that says something like "Important! If you change anything about this validation logic, also make sure to change it in <other place>."

publisher/src/stores/ReportStore.ts
```diff
@@ -212,11 +215,50 @@ class ReportStore {
...
+        /**  First, sort the datapoints so breakdowns come before top level metrics */
+        .sort((a, _) => (a.dimension_display_name ? -1 : 1))
+        .reduce((acc, val) => {
+          /** Add non-numeric characters */
+          if (Number.isNaN(Number(val.value))) {
+            acc[val.metric_definition_key] = true;
+            return acc;
+          }
+          /**
+           * If there is a value in the breakdowns and it's the first time we've come across this breakdown,
```
lilidworkin, line 231: How fun/clever! Thanks for the detailed comment.

publisher/src/stores/ReportStore.ts
```diff
@@ -212,11 +215,50 @@ class ReportStore {
...
+            acc[val.metric_definition_key] === undefined &&
+            val.value !== null
+          ) {
+            acc[val.metric_definition_key] = false;
+          }
+          /**
+           * After going through all of the breakdowns values, check to see if the top level metric has a null value.
+           * If so, we've encountered an error with breakdowns having a value and the top level metric having no value.
+           */
+          if (
```
lilidworkin, line 245: This is safe because the top level metric datapoint will always exist, right? It will just be null if it hasn't been filled out? I believe this is correct, anyway!

lilidworkin (approved):
This looks great @mxosman ! I think this was the right tradeoff to make, a small bit of hackiness here is better than doing something way more complicated. I think the right solution is to eventually working with Datapoint objects and replace them with the OG Metric object, but that's going to be a large refactor. Let's just keep an eye on how much pain the discrepancy is causing, and at the point that it feels worth tackling, we will.

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: Yeah - great call here! Let me see if there's something more stable to use!

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: Oh nice! Thank you, Lili!

publisher/src/stores/ReportStore.ts
mxosman: Great idea! Will do!

mxosman: I know, I cringed a little too having the same validation in two places.

publisher/src/stores/ReportStore.ts
mxosman: Yes - that is my assumption as far as the datapoints go. Before we filter all the `null` datapoints out, it should show a top level metric for each, but I'm not 100% sure how to best verify that the datapoints logic will not give us a case where there is no top-level metric.

lilidworkin: I'm fine with the risk for now ... #yolo

mxosman: 🤣

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: Added a flag just like you suggested - again - thank you for this!

[5c9ac3f](https://github.com/Recidiviz/justice-counts/pull/603/commits/5c9ac3fb3b6f097bc4fdc1a1c8e9f6787de8e95a)

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: Done & done: [5ed8ac7](https://github.com/Recidiviz/justice-counts/pull/603/commits/5ed8ac79430669ee9347872a18ee134928233ed7)

publisher/src/stores/ReportStore.ts
mxosman: Added here: [5ed8ac7](https://github.com/Recidiviz/justice-counts/pull/603/commits/5ed8ac79430669ee9347872a18ee134928233ed7)

PR #608, opened 2023-05-02
[Justice Counts] get reports for child agencies before publish

nichelle-hall:
## Description of the change

I was debugging why I wasn't able to publish the child agency reports from the super agency bulk upload metric review page and it turns out that they were not in the `ReportStore` because we only fetch the reports by the id of the current agency. I made a change here where we request the reports again at the data review page. This is another call to the BE, and another transaction with the DB. 

I tried to return them all at once when we populate the report store initially, but then the records overview page had 5x the original records because it included the records of the super agency  _and_ the child agencies.  

We could reduce the number of calls to the BE if the BE sends all records for the child agencies + super agencies whenever we went to fetch the records for a super agency. The response would key reports by the agency id / name. On the records overview page, the records could be sorted by super agency with some simple labels or we could only show the records created for the super agency. This refactor will probably break other parts of the FE as well. Alternatively, the FE can do this grouping.

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataUpload/UploadReview.tsx
+++ b/publisher/src/components/DataUpload/UploadReview.tsx
@@ -19,7 +19,7 @@ import { ButtonColor } from "@justice-counts/common/components/Button";
 import { showToast } from "@justice-counts/common/components/Toast";
 import { ReportOverview } from "@justice-counts/common/types";
 import { observer } from "mobx-react-lite";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import {
   Navigate,
   useLocation,
@@ -54,6 +54,19 @@ const UploadReview: React.FC = observer(() => {
     updatedReportIDs: number[];
     unchangedReportIDs: number[];
   };
+  useEffect(() => {
+    const getAllRelevantReports = async () => {
+      reportStore.resetState();
+      if (agencyId !== undefined) {
+        await reportStore.getReportOverviews(
+          /* agencyId */ agencyId?.toString(),
+          /* includeChildAgencies */ true
+        );
+      }
+    };
+    getAllRelevantReports();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
   const navigate = useNavigate();
   const [isSuccessModalOpen, setIsSuccessModalOpen] = useState(false);
   const [isExistingReportWarningModalOpen, setExistingReportWarningOpen] =
--- a/publisher/src/stores/ReportStore.ts
+++ b/publisher/src/stores/ReportStore.ts
@@ -110,10 +110,17 @@ class ReportStore {
     });
   }
 
-  async getReportOverviews(agencyId: string): Promise<void | Error> {
+  async getReportOverviews(
+    agencyId: string,
+    includeChildAgencies?: boolean
+  ): Promise<void | Error> {
+    let path = `/api/agencies/${agencyId}/reports`;
+    if (includeChildAgencies === true) {
+      path += "?includeChildAgencies=true";
+    }
     try {
       const response = (await this.api.request({
-        path: `/api/agencies/${agencyId}/reports`,
+        path,
         method: "GET",
       })) as Response;
       if (response.status === 200) {


lilidworkin (commented):
Oh wow, good debugging and great fix @nichelle-hall ! I want to give this a little bit more thought just to make sure there isn't a different way to do it, but I think you're probably right that this is the best way. I'll review more closely tomorrow.

mxosman:
> I was debugging why I wasn't able to publish the child agency reports from the super agency bulk upload metric review page and it turns out that they were not in the `ReportStore` because we only fetch the reports by the id of the current agency. I made a change here where we request the reports again at the data review page. This is another call to the BE, and another transaction with the DB.
> 
> I tried to return them all at once when we populate the report store initially, but then the records overview page had 5x the original records because it included the records of the super agency _and_ the child agencies.

Great call on this, Nichelle! Everything you wrote makes sense to me. I can't think of a better way to do this with our current set up now.

> We could reduce the number of calls to the BE if the BE sends all records for the child agencies + super agencies whenever we went to fetch the records for a super agency. The response would key reports by the agency id / name. On the records overview page, the records could be sorted by super agency with some simple labels or we could only show the records created for the super agency. This refactor will probably break other parts of the FE as well. Alternatively, the FE can do this grouping.

Yeah, this will take a bit more thought... It sounds like the right move to send over everything when we fetch records for a super agency. If we have an additional nesting/grouping to the way we store the reports in the store, we'll need to update every component that reaches for that information - but like you suggested, there might be a way to do this where it can exist in our current structure and the FE does the grouping for when they need to be displayed by group.

lilidworkin (commented):
@nichelle-hall , thanks for exploring this! A couple quick thoughts/questions:

* So to make sure I'm understanding correctly, the backend IS sending the right list of report IDs to the FE -- i.e. updatedReportIDs, unchangedReportIDs, those lists include the report IDs of child agencies, right?
* But then on line 80, when we look for the ReportOverview object corresponding to these IDs, we don't find the corresponding object, and that means that later, when we call `publishingMultipleRecords`, we send over an empty list?

If that's correct, I'm actually curious for @mxosman to chime in here -- is there a reason why we need to access the ReportOverview object? It looks like the only information we're grabbing from it is whether or not the record status is published, which we use to display the right warning. Maybe we can just skip that check in this case.

Basically what I'm trying to say is: if the BE is sending the right list of report IDs to the frontend (i.e. a list that includes child IDs) and all the FE has to do is send this list of report IDs _back_ to the BE when the user clicks publish, there doesn't really seem to be a reason that we need the full report object.

LMK know if this makes sense!

mxosman:
Great questions, Lili! If the issue was the FE uploading an empty list, then I wonder if the BE for the `updatedReportIDs`, etc. is agency specific (I didn't think it was, but might be).

I think the only need for a list of reports via the report overview is to display the list of Reports on the left side panel (we use the `month` `year` and `frequency` of the report overview object), and for the warning modal - which we can absolutely skip for this case -- and to check the publish status (to disable publishing for the rare case a user uploads unchanged published reports).

To your point, if the BE is indeed sending a list of all report IDs for all relevant agencies, the `publishingMultipleRecords` shouldn't have an issue accepting that list and publishing. I think the only thing the report overview obj is needed for is to render the list of reports - which we can decide to forgo.

LMK if there's an easy way for me to play around with this to see what's happening E2E - it's hard to wrap my mind around  the details looking at this at a high level.

lilidworkin:
Great discussion here, thanks @mxosman !

I think a few action items here:
* @nichelle-hall , can you help @mxosman repro the super agency setup locally so he can play with this?
* @nichelle-hall can you confirm whether or not the backend is sending the right list of report IDs to the FE -- i.e. updatedReportIDs, unchangedReportIDs -- do those lists include the report IDs of child agencies?

If so, then I think @mxosman can go ahead and forgo rendering the Report component on the sidebar and skip the warning modal, and just pass the report IDs directly into publishMultipleReports, which would be perfect!

nichelle-hall:
To repro, you will need to create create a super agency in your local environment. You can do by logging into the docker terminal locally that is attached to the `justice_counts_db` container. 

1) Query the `agency_id` of a Law Enforcement Agency locally. 
2) Then you can run this query to create 2 new child agencies!

`INSERT INTO source (name, type, systems, state_code, fips_county_code, super_agency_id) VALUES ('Law Enforcement Child Agency 1', 'agency', '{"LAW_ENFORCEMENT"}', 'US_NY', 'us_ny_new_york', <id of the law enforcement agency in your DB>);`

`INSERT INTO source (name, type, systems, state_code, fips_county_code, super_agency_id) VALUES ('Law Enforcement Child Agency 2', 'agency', '{"LAW_ENFORCEMENT"}', 'US_NY', 'us_ny_new_york', <id of the law enforcement agency in your DB>);`

3) Run Publisher Locally
4) Navigate to Bulk Upload, **select the account of the super agency (i.e the original agency you queried for** and upload this sheet --> [sample-super-agency-sheet.xlsx](https://github.com/Recidiviz/justice-counts/files/11392857/sample-super-agency-sheet.xlsx)


As for the other question, I was able to confirm that the ids `updatedReportIDs, unchangedReportIDs` are correct. They are ids of the child agencies.

lilidworkin:
Fantastic, thanks @nichelle-hall !

mxosman:
OK, so the instructions worked beautifully and I'm set up with Law Enforcement as the super agency w/ 2 child agencies. 

I was testing this earlier on `nichelle-hall/super-agency-bulk-upload` FE and `nichelle-hall/super-agencies-error-no-agency-col` BE just to see what the current behavior is when I upload the spreadsheet you attached.

The upload is successful - saw `1 metric was saved successfully (with warnings)` but the `/spreadsheet` endpoint is returning an empty list/set for everything (`unchanged_report_id`, `updated_report_ids`, `existing_report_ids`, `new_reports`). The only time I get something different is if I upload the same spreadsheet again - then the child agency report ids show up in `unchanged_report_id`. Otherwise - even when writing print statements in various places throughout the `upload_spreadsheet` endpoint and it's downstream functions, I'm getting empty lists/sets for `updated_report_ids` (even when you change a datapoint in a child agency), and `new_reports` (even when you upload new reports that don't exist in the child agencies). I'm suspecting there's something out of sync with those two pieces... Will dig in deeper but I can repro and see what you're running into with this.

nichelle-hall:
I'll try to reproduce this as well Mahmoud. I was using the same test sheet over and over so didn't see the behavior you described. I'll check it out!

mxosman:
But, it should still let you publish those unchanged reports! That's what was stumping me. But, I _totally_ see what's happening here. Without a reference to the reports, `hasAllPublishedRecordsNoOverwrites` unreliably becomes true and `existingReports` is empty because it's mapping for existing reports it's not finding in the store. Let me rethink this logic with no dependency on reports and see if we can make the same inferences without needing to fetch all child agency reports - at the _very_ least we should be able to successfully publish based on the report IDs alone. I have some ideas to play with - will follow up! Thanks for your patience!

mxosman:
Unfortunately I don't see a way for us to determine whether or not the user is uploading already published reports without getting those reports first to check their status. But, if we're OK not dealing with reports and not showing the "data has already been published" when a user tries to publish unchanged published reports (say that 10x fast) - we can just remove the checks and remappings entirely. Here's an example of all we'd need to do: https://github.com/Recidiviz/justice-counts/pull/615 (it should work hopefully in this PR's branch and allow you to publish those reports)

lilidworkin:
Thank you for investigating this so deeply @mxosman ! A few quick thoughts:

> But, if we're OK not dealing with reports and not showing the "data has already been published" when a user tries to publish unchanged published reports (say that 10x fast) - we can just remove the checks and remappings entirely. Here's an example of all we'd need to do: https://github.com/Recidiviz/justice-counts/pull/615

I think I'm fine with this. If we did want to keep this warning, then the right way to do it is probably for the backend to send this information. (This might have been the right way to do it anyway.) Instead of sending over multiple lists of report IDs, the backend could send over a list of dictionaries, each of which contains the report ID and other metadata about it -- i.e. whether it's newly created or updated, and if it's updated then whether its published or not, etc. 

We might find it helpful to move in this direction anyway at some point, but I'll leave this up to you guys.

As for the other stuff -- I'm not deep in the weeds to weigh in yet -- I think I'll leave this up to the two of you to troubleshoot together. Let me know if you need my help on anything!

nichelle-hall:
I was able to debug the `updated_report_ids` bug! With that change, what approach do you think is best @mxosman? Now it should be sending the right values for `updated_report_ids`, `new_reports` and `unchanged_report_ids`.

nichelle-hall:
Yes! It is fixed in the backend [companion PR](https://github.com/Recidiviz/recidiviz-data/pull/20618).

lilidworkin:
Ah thanks! I hadn't seen that one!

mxosman:
> I was able to debug the `updated_report_ids` bug! With that change, what approach do you think is best @mxosman? Now it should be sending the right values for `updated_report_ids`, `new_reports` and `unchanged_report_ids`.

Oh nice! I really like Lili's idea of sending a list of report metadata vs. a list of IDs for new, updated and unchanged reports. Since we surface `all_reports` anyway and a list of report IDs for each type (new, updated and unchanged), any harm in sending the FE the full reports? If that's too much bloat, we can extract only what is needed before sending. The only reason I'm leaning this way is in the interest of avoiding another BE call + it keeps all the current functionality (I think?). Curious to hear your thoughts as well! I think what you proposed here also works and makes sense!

nichelle-hall:
> companion PR

I think that this is doable, let me try it out and I'll let you know when those changes are in the backend PR.

mxosman:
> > companion PR
> 
> @mxosman I've pushed the change to the companion BE PR. Just a heads up, I didn't include editor metadata in the report response. I ran it with the FE and nothing is breaking! It was a pretty straight forward change.

Amazing!! I don't think we need editor metadata, so this works perfectly! Just tried it and I see the full report metadata! The only thing is the unchanged reports are not updating when I make an edit, but I think you solved that in another PR. 

Let me work on refactoring the FE to use the full reports instead of just the IDs + retrieving report details from the store - if this works, I think we can eliminate this PR's call to the backend for child agency reports.

nichelle-hall:
Oh! The unchanged reports behavior is a bug! I'll work on that as you update the FE. I'll make sure the changes are on the companion BE PR.

mxosman:
Thank you, Nichelle! Here's the FE portion - let me know if it works for you!

PR #614, opened 2023-05-04
[Fix] Datapoint validation

mxosman:
## Description of the change

**Issue**: when you are bulk publishing reports, if one report has metrics + breakdown values, and another report does not have any values, you will encounter a validation error mark in the review page (the error being that there was a disaggregation with a value and a top level metric without a value) and are thus blocked from publishing. 

**Fix**: we need to be more specific with our metric errors mapping and include a reference to the report ID so that it only checks to see if there's a disaggregation-with-value-top-level-metric-with-no-value within each individual report and not across the all reports.

https://user-images.githubusercontent.com/59492998/236291573-7c98440d-7e56-4ac8-b088-40efe81881af.mov

## Related issues

Closes #XXXX

### Development

### Code review


publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -109,9 +111,9 @@ export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
                 buttonColor={buttonColor}
                 labelColor={labelColor}
                 borderColor={borderColor}
-                showTooltip={disabled && isPublishButton}
+                showTooltip={hasMetricErrors && isPublishButton}
```
lilidworkin, line 114: Can you elaborate on this change? Where was `disabled` coming from before?

lilidworkin (approved):
Thanks for the fix @mxosman , this "double key" looks like a good solution to me!

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: Aha! In the DataEntryReview component, whenever you publish a report, isPublishable is set to false, which sets `disabled` to true...

```
  const publishReport = async () => {
    if (isPublishable) {
      setIsPublishable(false);
```

We can remove this and it will fix it, but I think I'm going to double down on doing this check in the Review page and remove it from the BulkActionReview & DataEntryReview - that way the logic is one place. We leave both of those components with the task of logging errors they detect going through each metric and leave the ReviewMetric component with the task of determining whether or not everything in the review page is publishable. Does that make sense?

lilidworkin: Yes, perfect! Thanks for explaining!

lilidworkin (approved):

PR #639, opened 2023-05-18
Shared radio button

### LINKED ISSUE #543: [UI Components] Radio Button
Create a reusable button component that allows a user to select only one option (button) out of many

ivan-kishko:
## Description of the change

> Shared radio button component

## Type of change
Type: Breaking Change

## Related issues

Closes #543 

### Development

### Code review


publisher/src/components/Forms/BinaryRadioButton.tsx
```diff
@@ -1,191 +0,0 @@
...
-      <RadioButtonLabel
-        disabled={disabled}
-        buttonSize={buttonSize}
-        htmlFor={props.id}
-      >
-        {label}
-      </RadioButtonLabel>
-    </RadioButtonWrapper>
-  );
-};
```
mxosman, line 191: 🧹 🧹 🧹 🥳

publisher/src/components/Reports/CreateReport.tsx
```diff
@@ -65,6 +64,14 @@ function createIntegerRange(start: number, end: number) {
...
   font-size: ${typography.sizeCSS.medium};
   margin-top: 24px;
+  margin-bottom: 12px;
+`;
+
+const RadioButtonsWrapper = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  gap: 8px;
```
mxosman, line 74: small nit: I want to avoid having multiple wrappers - I know this particular styling is only used here, but it might be useful in the future - can we consolidate this with the `RadioButtonWrapper` in `RadioButton.styled.tsx` and have a prop control which styling is used?

mxosman (approved):
Thank you so much, Ivan! Looks fantastic! One small nit to address and it's good to merge after that!

publisher/src/components/Reports/CreateReport.tsx
ivan-kishko: Hello @mxosman. Hope you had nice holidays. About this, I am not sure what you mean, to make a few different wrappers for radio buttons inside RadioButton.styled? If so then you have to somehow control spacing between this wrappers and other elements where this wrapper is used, and it will end up either with another wrapper around radio buttonwrapper or styled RadioButtonWrapper. THe idea for shared components is that they do not control space around them, only inside. That is why we have all this regular buttons wrappers throughout application. If I get this wrong please let me know.

mxosman: Hi Ivan! Thank you - hope the same for you! 

Sorry - to clarify, I just meant can we use `RadioButtonWrapper` that's in `RadioButton.styled.tsx` for this and refactor it so it can be used for this component as well?

Something like this:

```
const RadioButtonsWrapper = styled.div<{fullWidth?: boolean}>`
  display: flex;
  ${({fullWidth}) => fullWidth ? "width: 100%; flex-direction: row; gap: 8px;" : "flex: 1 1 0;"}
`
```

Let me know if it doesn't make sense to do that! What you have works perfectly fine - just wanted to see if it makes sense to consolidate the wrapper.

ivan-kishko: But in CreateReport Wrapper is for 2 buttons and in RadioButton component wrapper is only used for 1 button because we hide input to make this radio button custom looking

mxosman: Oh! Gotcha! Thank you for clarifying - that makes sense. I still think it would be good to revisit and include some kind of default wrapper for two radio buttons in the shared styles - especially since it's common for radio buttons to be wrapped in pairs. The regular buttons makes sense to not include a default wrapper since they have a huge variety of possible wrappers depending on where they are in the app. But, radio buttons almost always have some kind of wrapper that encapsulates two radio buttons - and for the most part the styling is relatively similar (e.g. `SpecifyEthnicityButtonsContainer`, `SettingOptions`, etc.)

ivan-kishko: [220d6a4](https://github.com/Recidiviz/justice-counts/pull/639/commits/220d6a4f7e9a4e235dc94ff4e866223ec51d40e2) Hello @mxosman!) how about this?

common/components/RadioButton/RadioButton.styled.tsx
```diff
@@ -18,7 +18,20 @@
...
+}>`
+  width: 100%;
+  display: flex;
+  flex-direction: row;
+  gap: 8px;
+  margin: ${({ spacing }) =>
+    `${spacing?.top || 0}px ${spacing?.right || 0}px ${
+      spacing?.bottom || 0
+    }px ${spacing?.left || 0}px`};
+`;
```
mxosman, line 34: Nice! This is exactly what I had in mind. I don't think it's entirely necessary to create a component for this - but just this styling is enough for us to import when using two radio buttons.

ivan-kishko: well I mean this will cause confusion if you import Styled component and React component from the same place, maybe I can put Wrapper inside RadioButton.tsx, but I think it is better to be consistent here. Idea is to import styled components only in styling files if you want to add some specific customization for shared ones. Does it make sense?

mxosman: I like that argument - I'm good with what you have! Makes sense to me. I always appreciate hearing your thoughts - thank you for the explanation! Beautifully executed.

mxosman (approved):
Good to merge! Thank you for entertaining my thoughts and for the discussion!

PR #649, opened 2023-05-22
Adjustment to UploadReview logic to allow a super agency to publish for their child agencies

mxosman:
## Description of the change

Now that the BE sends reports' metadata, this updates the `UploadReview` logic to utilize the metadata for rendering instead of the report IDs + querying the `reportStore`.

## Related issues

Closes #653 

### Development

**This box MUST be checked by the submitter prior to merging**:
- [x] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [ ] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #660, opened 2023-05-24
Refactor bulk upload review page to handle and properly display multiple agency uploads

### LINKED ISSUE #658: [Superagencies] Adjust Bulk Upload Review Page to support Superagencies
See task for design [here](https://app.zenhub.com/workspaces/justice-counts--622109ea3bb6e0001d2d06ad/issues/gh/recidiviz/justice-counts/657)

mxosman:
## Description of the change

Refactors the bulk upload review page to handle and properly display multiple agency uploads. The review page treatment remains the same for single agency uploads.

Screenshot of multiple agency upload treatment w/ overwrites:
Screenshot 2023-05-25 at 2 04 37 PM
Question: does this look right for uploading overwrites? It doesn't show the Staff metrics in the review page, which I thought were included in the spreadsheet (albeit there were no changes there - I only changed the arrest metrics to trigger the overwrite)

Video comparing single agency vs multiple agency upload:

https://github.com/Recidiviz/justice-counts/assets/59492998/fc2221a0-c227-4a8d-b311-d5f1477b713c

Spreadsheet I've been using: 
[sample-super-agency-sheet-two-metrics.xlsx](https://github.com/Recidiviz/justice-counts/files/11568161/sample-super-agency-sheet-two-metrics.xlsx)

NOTE: If running locally, you'll need to setup your account to have a super agency + child agencies. In the example above, my super agency is Law Enforcement w/ two Law Enforcement child agencies. Instructions can be found here: https://github.com/Recidiviz/justice-counts/pull/608#issuecomment-1534087462

## Related issues

Closes #658 

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataUpload/UploadReview.tsx
+++ b/publisher/src/components/DataUpload/UploadReview.tsx
@@ -125,7 +125,6 @@ const UploadReview: React.FC = observer(() => {
 
   const { isSuperAgencyUpload, datapointsByAgencyName } =
     uploadedMetricsToDatapointsByAgencyName(uploadedMetrics);
-
   const metrics = uploadedMetrics
     .map((metric) => {
       return {
--- a/publisher/src/components/DataUpload/UploadReview.tsx
+++ b/publisher/src/components/DataUpload/UploadReview.tsx
@@ -18,7 +18,7 @@
 import { palette } from "@justice-counts/common/components/GlobalStyles";
 import { Modal } from "@justice-counts/common/components/Modal";
 import { showToast } from "@justice-counts/common/components/Toast";
-import { ReportOverview } from "@justice-counts/common/types";
+import { RawDatapoint, ReportOverview } from "@justice-counts/common/types";
 import { printReportTitle } from "@justice-counts/common/utils";
 import { observer } from "mobx-react-lite";
 import React, { useState } from "react";
@@ -103,13 +103,38 @@ const UploadReview: React.FC = observer(() => {
     setIsSuccessModalOpen(true);
   };
 
+  const uploadedMetricsToDatapointsByAgencyName = (
+    metrics: UploadedMetric[]
+  ) => {
+    const allDatapoints = metrics
+      .flatMap((metric) => metric.datapoints)
+      .filter((dp) => dp.value);
+    const datapointsByAgencyName = allDatapoints.reduce((acc, dp) => {
+      if (!dp.agency_name) return acc;
+      if (!acc[dp.agency_name]) {
+        acc[dp.agency_name] = [];
+      }
+      acc[dp.agency_name].push(dp);
+      return acc;
+    }, {} as { [key: string]: RawDatapoint[] });
+    const isSuperAgencyUpload =
+      Object.values(datapointsByAgencyName).length > 1;
+
+    return { isSuperAgencyUpload, datapointsByAgencyName };
+  };
+
+  const { isSuperAgencyUpload, datapointsByAgencyName } =
+    uploadedMetricsToDatapointsByAgencyName(uploadedMetrics);
+
   const metrics = uploadedMetrics
-    .map((metric) => ({
-      ...metric,
-      datapoints: metric.datapoints.filter((dp) => dp.value !== null),
-      metricHasError: false,
-      metricHasValidInput: true,
-    }))
+    .map((metric) => {
+      return {
+        ...metric,
+        datapoints: metric.datapoints.filter((dp) => dp.value !== null),
+        metricHasError: false,
+        metricHasValidInput: true,
+      };
+    })
     .filter((metric) => metric.datapoints.length > 0);
   const overwrites: ReviewMetricOverwrites[] = [];
   metrics.forEach((metric) => {
@@ -221,6 +246,8 @@ const UploadReview: React.FC = observer(() => {
         metrics={metrics}
         metricOverwrites={overwrites}
         records={existingAndNewRecords}
+        isSuperAgencyUpload={isSuperAgencyUpload}
+        datapointsByAgencyName={datapointsByAgencyName}
       />
     </>
   );
--- a/publisher/src/components/DataUpload/UploadReview.tsx
+++ b/publisher/src/components/DataUpload/UploadReview.tsx
@@ -159,6 +159,7 @@ const UploadReview: React.FC = observer(() => {
           key: dp.id,
           metricName: dp.metric_display_name || "",
           dimensionName: dp.dimension_display_name || "",
+          agencyName: dp.agency_name,
           startDate: dp.start_date,
         };
         overwrites.push(overwriteData);
--- a/publisher/src/components/DataUpload/UploadReview.tsx
+++ b/publisher/src/components/DataUpload/UploadReview.tsx
@@ -110,11 +110,14 @@ const UploadReview: React.FC = observer(() => {
       .flatMap((metric) => metric.datapoints)
       .filter((dp) => dp.value);
     const datapointsByAgencyName = allDatapoints.reduce((acc, dp) => {
-      if (!dp.agency_name) return acc;
+      if (!dp.agency_name || !dp.metric_display_name) return acc;
       if (!acc[dp.agency_name]) {
-        acc[dp.agency_name] = [];
+        acc[dp.agency_name] = {};
       }
-      acc[dp.agency_name].push(dp);
+      if (!acc[dp.agency_name][dp.metric_display_name]) {
+        acc[dp.agency_name][dp.metric_display_name] = [];
+      }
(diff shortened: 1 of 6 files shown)


publisher/src/components/DataUpload/UploadReview.tsx
```diff
@@ -103,13 +103,37 @@ const UploadReview: React.FC = observer(() => {
...
+    const datapointsByAgencyName = allDatapoints.reduce((acc, dp) => {
+      if (!dp.agency_name) return acc;
+      if (!acc[dp.agency_name]) {
+        acc[dp.agency_name] = [];
+      }
+      acc[dp.agency_name].push(dp);
+      return acc;
+    }, {} as { [key: string]: RawDatapoint[] });
+    const isSuperAgencyUpload =
+      Object.values(datapointsByAgencyName).length > 1;
```
mxosman, line 121: TODO: come up with a better name for this - since a super agency can upload for one agency. Maybe `isMultiAgencyUpload`?

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -86,6 +96,66 @@ export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
...
+      (acc, overwrite) => {
+        if (!overwrite.agencyName) return acc;
+        if (!acc[overwrite.agencyName]) {
+          acc[overwrite.agencyName] = [];
+        }
+        acc[overwrite.agencyName].push(overwrite);
+        return acc;
+      },
+      {} as { [key: string]: ReviewMetricOverwrites[] }
+    );
```
mxosman, line 136: I debated whether or not to do this grouping here or in the `UploadReview` component. I ultimately chose here because it only needs to do this grouping if `isMultiAgencyUpload` is true and this function only runs when `isMultiAgencyUpload` is true. 

I could also do the grouping in the `UploadReview` component and not run the reducer if `isMultiAgencyUpload` is false, and then pass in an optional `groupedMetricOverwrites` object as a prop to this component, but that didn't feel entirely necessary. Happy to adjust if it makes more sense to keep this logic in the `UploadReview`.

mxosman:
Would love a sanity check if folks are able to test this locally! One thing I'm noticing is if a user is uploading new reports, the agency_name is `null` - which then affects the multiple agency logic & treatment since it heavily relies on the `agency_name`. (cc @nichelle-hall - incase you have any insight/ideas)

publisher/src/components/DataUpload/UploadReview.tsx
```diff
@@ -103,6 +104,45 @@ const UploadReview: React.FC = observer(() => {
...
+   * @returns Object { isMultiAgencyUpload, datapointsByMetricNameByAgencyName }
+   *          - isMultiAgencyUpload { boolean } - whether or not this is a multiple agency upload
+   *          - datapointsByMetricNameByAgencyName { object } - grouped
+   * @example
+   *  {
+   *    "Agency 1": { "Staff": [...datapoints], "Arrests": [...datapoints], ... }
+   *    "Agency 2": { "Staff": [...datapoints], ... },
+   *  }
+   *   */
+  const metricsToDatapointsByMetricNameByAgencyName = (
```
lilidworkin, line 118: Logic here looks great to me

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
lilidworkin: I think this is totally fine

lilidworkin:
> One thing I'm noticing is if a user is uploading new reports, the agency_name is null - which then affects the multiple agency logic & treatment since it heavily relies on the agency_name.

Also noticed this! Thanks for calling this out. Commented on @nichelle-hall 's PR [here](https://github.com/Recidiviz/recidiviz-data/pull/20618/files#r1206993709)

lilidworkin:
> Question: does this look right for uploading overwrites? It doesn't show the Staff metrics in the review page, which I thought were included in the spreadsheet (albeit there were no changes there - I only changed the arrest metrics to trigger the overwrite)

@mxosman I'm not able to repro this, but it does seem like a bug -- it should still show even if there are no changes. LMK if you see it again ...

mxosman:
> @mxosman this looks great. I agree from reading your PR description + comments that we need to investigate some backend behavior here, but from what I can tell, that seems independent from your changes. All of the FE logic looks straightforward/simple enough and I think the ball is in the BE court now to figure out if we're sending you the right info. Do you agree / does that make sense? And if I'm understanding correctly, this is also safe to merge now even without waiting for the BE changes, because we're not currently sending agency_name yet, in which case everything here should just default to working the way it currently does. Anyway, all this to say I feel comfortable accepting/merging this, and then I can work with @nichelle-hall to figure out the BE pieces. LMK if you agree or if I'm missing something. Thank you again for taking the detour to work on this!
> 
That makes sense to me! This is stacked on top of #649 - so should be safe to merge into #649 and we'll wait for to merge that when #20618 is merged. But regardless - I think you're right, this shouldn't break the current behavior at all if merged independently. Sounds good to me, Lili! Absolutely - thank you for your eyes and the sanity check!

> One side note -- I do think, NOT in this PR, but maybe as a followup that Ivan could work on, that we may also want to separate the "Metrics" and "Records" boxes in the left hand side by agency too. Do you see any reasons why that would be trickier than how you separated out the "Overwrites" section? Don't spend too much time thinking about this, just curious if anything pops to mind immediately.

Yeah - great call! I had this question too but completely forgot about it! Definitely think it's worth doing - looks very strange without. It shouldn't be trickier at all - we'll just have to group the reports and metrics list by agency name. In fact, I'm now recalling a lovely `groupBy` helper function a certain Lili Dworkin wrote to do this exact thing! 😄 It's not a big lift at all - happy to include in this PR! 

> > Question: does this look right for uploading overwrites? It doesn't show the Staff metrics in the review page, which I thought were included in the spreadsheet (albeit there were no changes there - I only changed the arrest metrics to trigger the overwrite)
> 
> @mxosman I'm not able to repro this, but it does seem like a bug -- it should still show even if there are no changes. LMK if you see it again ...

Let me look at this again and see if I can repro and note all the steps - will definitely keep an eye on it!

mxosman:
Records & Metrics list on the left side have been updated:

Screenshot 2023-05-29 at 12 40 57 PM

mxosman:
> Looks perfect. I'm fine with the choice to remove the metric icons in the multi agency view, I feel like that needs more thought anyway. Thanks so much for making these additional changes @mxosman !

Oh whoops! That was a total accident - I did not intend to do that. Should I keep them removed? Also - happy to add them back in!

lilidworkin:
@mxosman if it's easy to add them back I guess we could! I don't feel strongly honestly. I assumed it was intentional because it was hard to do

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
```diff
@@ -86,6 +97,130 @@ export const ReviewMetrics: React.FC<ReviewMetricsProps> = ({
...
+      <SummaryWrapper>
+        {datapointsByMetricNameByAgencyName &&
+          Object.entries(datapointsByMetricNameByAgencyName).map(
+            ([agencyName, agencyMetrics]) => {
+              return (
+                <Fragment key={agencyName}>
+                  <SummaryAgencyName>{agencyName}</SummaryAgencyName>
+                  {Object.keys(agencyMetrics).map((metric) => (
+                    <SummarySectionLine key={metric}>
+                      <MetricStatusIcon src={checkIcon} alt="" />
```
lilidworkin, line 181: > I think this was done because we have a separate errors/warnings page for the upload -> review flow - and by the time they get to the review page, all metrics will be in checkmark status, right? I added the checkmark icon back for now.

This makes sense to me -- let's add a comment here explaining though, because this code is in ReviewMetrics, not UploadReview, so someone reading it might be confused. I think we can say that currently the only way to upload data for multiple agencies is via BulkUpload, so if we're in this block of code that's where we must be coming from, and  in that case it's safe to assume that by the time we get here, none of the metrics have errors.

lilidworkin (approved):
Awesome, thanks for making this change! I agree with your reasoning. Added a comment inline about adding a comment in the code :) Thanks!

publisher/src/components/ReviewMetrics/ReviewMetrics.tsx
mxosman: Let me know if this feels clear: [f8454c2](https://github.com/Recidiviz/justice-counts/pull/660/commits/f8454c2c2cd6f8c8fb72096a8cddbe0f84227a4e)

mxosman: Will merge for now since I noticed the BE is merged but happy to adjust if necessary!

PR #682, opened 2023-05-29
New Homepage: UI Scaffolding & Task Card Interactions [1/n]

### LINKED ISSUE #669: [Frontend] Home Page: Scaffolding

### LINKED ISSUE #670: [Frontend] Home Page: Task cards

### LINKED ISSUE #717: [Frontend] New Homepage: fix linking from task card to report review page

### LINKED ISSUE #719: [Frontend] New Homepage: handle case when agency has no records

mxosman:
## Description of the change

Introduces UI scaffolding and basic task card functionality for the new home page - namely the `Home` and `TaskCard` components. 

Reference: 
  * [Figma](https://www.figma.com/file/dvX3yg7FMGJSWFaqHWrWgQ/Justice-Counts-Publisher?type=design&node-id=8312%3A20720&t=7rIYt3rsxpsw4f72-1)
  * API: https://paper.dropbox.com/doc/API-Home--B6TtO70zEp8JK6LOuK07RO4LAg-ybX2dpUPcuizO5E76WLgu

Backend branch if running locally: `mahmoud/home-be`
Demo: https://mahmoudtest---justice-counts-web-qqec6jbn6a-uc.a.run.app/

*Quick Note: linking to reports and review page is still funky right now - will tackle in a follow up PR (see below)

---

Remaining to dos for the homepage:
  * Support multi-system agencies (https://github.com/Recidiviz/justice-counts/issues/716)
  * Finish implementing backend (https://github.com/Recidiviz/recidiviz-data/issues/21609)
  * ~Handle case when agency has no records (https://github.com/Recidiviz/justice-counts/issues/719)~ [Resolved in this PR]
  * ~Reports linked from the task cards don't always load properly - will need to investigate and potentially refactor parts of the ReportStore/DataEntryForm/DataEntryReview components which will be done in a follow up PR (https://github.com/Recidiviz/justice-counts/issues/717)~ [Resolved in this PR]

Remaining to dos for project:
  * Add "Back to home" button to Publish success modal
  * Add "Back to home" button to data entry form
  * Scroll to metric in Data Entry Form after clicking on "Manual Entry" action link in task card
  * Agency Settings & Metric Config: Action Required treatment
  * Adjust toast size/positioning to avoid clashing with the new agency dropdown menu
  * Responsiveness

## Related issues

Closes #669, Closes #670, Closes #719, Closes #717

### Development

### Code review


publisher/src/components/DataUpload/DataUpload.tsx
```diff
@@ -333,6 +333,10 @@ export const DataUpload: React.FC = observer(() => {
     setSelectedSystem(userSystems.length === 1 ? userSystems[0] : undefined);
   }, [userSystems]);
 
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, []);
+
```
mxosman, line 339: Since the home page is fully scrollable, the scroll position is inherited in the `DataUpload` component (e.g. you scroll down the home page and click on `Data Upload` - you'll be scrolled down the same distance in the `DataUpload` instruction template). This ensures whenever you load the `DataUpload` screen, you always start at the top of the page.

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,333 @@
...
+          description: metric.description,
+          actionLinks: hasMetricValue
+            ? [taskCardLabelsActionLinks.publish]
+            : [
+                taskCardLabelsActionLinks.uploadData,
+                taskCardLabelsActionLinks.manualEntry,
+              ],
+          metricFrequency,
+          hasMetricValue,
+        };
```
mxosman, line 99: These objects are the `TaskCardMetadata` that power the `TaskCard `component. 

Here, we're looping through all of the agency's metrics, filtering to keep only the enabled metrics, and building an array of `TaskCardMetadata`. We do the same immediately below for the unconfigured metrics. We use both arrays to render the appropriate task cards.

publisher/src/stores/ReportStore.ts
```diff
@@ -165,6 +166,28 @@ class ReportStore {
...
+          `There was an issue getting this ${REPORT_LOWERCASE} .`
+        );
+      }
+
+      const latestReportsAndMetrics = await response.json();
+      return latestReportsAndMetrics;
+    } catch (error) {
+      if (error instanceof Error) return new Error(error.message);
+    }
+  }
```
mxosman, line 189: Calls the new endpoint to retrieve the agency's metrics, latest monthly and annual record(s).

publisher/src/components/Reports/DataEntryForm.tsx
```diff
@@ -250,6 +250,10 @@ const DataEntryForm: React.FC<{
     document.body.style.overflow = showDataEntryHelpPage ? "hidden" : "unset";
   }, [showDataEntryHelpPage]);
 
+  useEffect(() => {
+    window.scrollTo(0, 0);
+  }, []);
```
mxosman, line 255: Fixes the same issue found in the `DataUpload` component above.

publisher/src/components/Reports/DataEntryForm.tsx
```diff
@@ -268,7 +272,7 @@ const DataEntryForm: React.FC<{
   return (
     <>
       <HeaderBar
-        onLogoClick={() => navigate(`/agency/${agencyId}/${REPORTS_LOWERCASE}`)}
+        onLogoClick={() => navigate(`/agency/${agencyId}`)}
```
mxosman, line 275: Updating the logo path here to go back to the home page - before this went to the records page because that was the OG homepage.

publisher/src/components/Home/types.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export type AnnualRecordMetadata = {
+  [startingMonth: string]: LatestRecordMetadata;
+};
+
+export type LatestAnnualMonthlyRecordMetadata = {
+  monthly: LatestRecordMetadata;
+  annual: AnnualRecordMetadata;
+};
```
mxosman, line 61: Have I used the word "metadata" enough in this feature?

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,333 @@
...
+      }
+      return acc;
+    },
+    {
+      allMetricsWithValues: [],
+      allMetricsWithoutValuesOrNotConfigured: [],
+    } as {
+      [key: string]: TaskCardMetadata[];
+    }
+  );
```
mxosman, line 139: Metrics **_with_** values are considered ready to publish and receive different prep and rendering than metrics without values and unconfigured metrics.

Metrics without values or not yet configured are straightforwardly rendered. 

Since we don't have the functionality to publish metrics individual (currently, you have to publish the entire record the metric is in - and a record often contains multiple metrics) -  metrics that go from an _enter-data_ state to _ready-to-publish_ state (by adding a value to the metric in the latest record) behave in a couple of ways:

1. The "Upload Data | Manual Entry" card goes away from the homepage for that particular metric.
2. There is a new Publish data card that appears in the homepage with the latest record title that the metric was in.
  * Example: 
    * You configured two monthly metrics (Funding and Expenses) and have not entered data for them yet in the latest monthly record (June 2023). At this point, two cards will appear (one for Funding with "Upload Data | Manual Entry" action links - and another for Expenses with the same action links). 
    * You enter data for Funding and go back to the home page. You will no longer see a Funding task card, but now see a new task card titled "June 2023" with a "Publish" action link.
    * When you enter data for Expenses now, the same thing happens where the "Upload Data | Manual Entry" task card goes away and you now only see one (the same) task card titled "June 2023" with a "Publish" action link

mxosman, line 144: Metrics **_with_** values are considered ready to publish and receive different prep and rendering than metrics without values and unconfigured metrics.

Metrics without values or not yet configured are straightforwardly rendered. 

Since we don't have the functionality to publish metrics individually (currently, you have to publish the entire record the metric is in - and a record often contains multiple metrics) -  metrics that go from an _enter-data_ state to _ready-to-publish_ state (by adding a value to the metric in the latest record) behave in a couple of ways:

1. The "Upload Data | Manual Entry" card goes away from the homepage for that particular metric.
2. There is a new Publish data card that appears in the homepage with the latest record title that the metric was in.
  * Example: 
    * You configured two monthly metrics (Funding and Expenses) and have not entered data for them yet in the latest monthly record (June 2023). At this point, two cards will appear (one for Funding with "Upload Data | Manual Entry" action links - and another for Expenses with the same action links). 
    * You enter data for Funding and go back to the home page. You will no longer see a Funding task card, but now see a new task card titled "June 2023" with a "Publish" action link.
    * When you enter data for Expenses now, the same thing happens where the "Upload Data | Manual Entry" task card goes away and you now only see one (the same) task card titled "June 2023" with a "Publish" action link

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,338 @@
...
```
mxosman, line 144: Metrics **_with_** values are considered ready to publish and receive different prep and rendering than metrics without values and unconfigured metrics.

Metrics without values or not yet configured are straightforwardly rendered. 

Since we don't have the functionality to publish metrics individually (currently, you have to publish the entire record the metric is in - and a record often contains multiple metrics) -  metrics that go from an _enter-data_ state to _ready-to-publish_ state (by adding a value to the metric in the latest record) behave in a couple of ways:

1. The "Upload Data | Manual Entry" card goes away from the homepage for that particular metric.
2. There is a new Publish data card that appears in the homepage with the latest record title that the metric was in.
  * Example: 
    * You configured two monthly metrics (Funding and Expenses) and have not entered data for them yet in the latest monthly record (June 2023). At this point, two cards will appear (one for Funding with "Upload Data | Manual Entry" action links - and another for Expenses with the same action links). 
    * You enter data for Funding and go back to the home page. You will no longer see a Funding task card, but now see a new task card titled "June 2023" with a "Publish" action link.
    * When you enter data for Expenses now, the same thing happens where the "Upload Data | Manual Entry" task card goes away and you now only see one (the same) task card titled "June 2023" with a "Publish" action link

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+          latestMonthlyAnnualRecordsMetadata?.monthly?.metrics?.find(
+            (latestMonthlyMetric) => latestMonthlyMetric.key === metric.key
+          )?.value;
+        const latestAnnualMetricValue =
+          metricFrequency === "ANNUAL" &&
+          startingMonth &&
+          latestMonthlyAnnualRecordsMetadata?.annual[
+            startingMonth
+          ]?.metrics?.find(
+            (latestAnnualMetric) => latestAnnualMetric.key === metric.key
```
lilidworkin, line 94: Do we also have to check if `latestAnnualMetric.value` is not null? Cause don't sometimes the metrics in this list have null values? Or maybe those have already been filtered out somewhere?

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+        return {
+          reportID,
+          title: metric.display_name,
+          description: metric.description,
+          actionLinks: [taskCardLabelsActionLinks.metricAvailability],
+          metricSettingsParams: `?system=${metric.system.key.toLowerCase()}&metric=${metric.key.toLowerCase()}`,
+        };
+      }) || [];
+  /**
+   * Metrics without values or not yet configured (`allMetricsWithoutValuesOrNotConfigured`) are straightforwardly rendered.
```
lilidworkin, line 141: This took me a sec to understand but it's a clear comment and the code is clear! Basically we start with two lists: metrics that are configured but may or may not have values, and metrics that are unconfigured. And then we rearrange into two different lists: metrics that are configured have values, and metrics that are configured and don't have values + metrics that are unconfigured.

publisher/src/components/Home/Home.tsx
lilidworkin: This makes sense and definitely seems to align with the behavior we all agreed upon!

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+    },
+    {
+      allMetricsWithValues: [],
+      allMetricsWithoutValuesOrNotConfigured: [],
+    } as {
+      [key: string]: TaskCardMetadata[];
+    }
+  );
+  /** Determining whether or not all tasks have been completed */
+  const hasNoEnabledOrUnconfiguredMetricsTaskCardMetadata =
```
lilidworkin, line 165: I'm having a little trouble understanding this part but I'm also not sure I'm thinking clearly -- what is a scenario where we wouldn't have `hasNoEnabledOrUnconfiguredMetricsTaskCardMetadata` -- i.e. that would be false -- but we'd have `hasPublishedLatestAnnualRecords && hasPublishedLatestMonthlyRecord && allMetricsWithoutValuesOrNotConfigured.length === 0` and therefore we'd consider all tasks completed?

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+            allMetricsWithoutValuesOrNotConfigured.map((taskCardMetadata) => (
+              <TaskCard
+                key={taskCardMetadata.title}
+                metadata={taskCardMetadata}
+                reportID={taskCardMetadata.reportID}
+              />
+            ))
+          )}
+
+          {/* Publish-Ready Cards (for Monthly & Annual Records) */}
```
lilidworkin, line 320: Do the Publish cards live outside of the `hasCompletedAllTasks` logic? In other words is it possible to have completed all tasks but still see Publish cards? That's what it looks like to me here but I might be missing something.

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+                metadata={createPublishRecordTaskCardMetadata(
+                  latestMonthlyAnnualRecordsMetadata.monthly.reportTitle,
+                  "MONTHLY"
+                )}
+                reportID={latestMonthlyAnnualRecordsMetadata.monthly.id}
+              />
+            )}
+
+          {/* Publish latest annual record(s) */}
+          {allMetricsWithValues.find(
```
lilidworkin, line 339: Will this render a Publish card for each type of annual report? e.g. if we have a Fiscal year report + metric that's ready to be published and an Annual year report + metric that's ready to be published, will we see two cards?

publisher/src/components/Home/types.ts
lilidworkin: Haha!

publisher/src/components/Reports/DataEntryForm.tsx
lilidworkin: Oh nice, ignore my comment above then. I wonder if there's a way to add this to every page to be safe? Cause like what if we add another scrollable page in the future but forget to add this ... How do other apps solve this problem?

publisher/src/components/Reports/DataEntryForm.tsx
lilidworkin: Good catch!

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+        Object.values(monthlyRecord).length === 0 &&
+        Object.values(annualRecords).length === 0
+      ) {
+        setAgencyMetrics([]);
+        setLoading(false);
+        return;
+      }
+
+      /**
+       * Create annual records metadata objects in the same structure as the response JSON:
```
lilidworkin, line 242: Why do we need to do this? If the response JSON is in this format, how did they get unformatted and need to get reformatted?

lilidworkin (commented):
WOW! This is quite an undertaking. Incredible @mxosman ! Let me know if my comments/questions make sense!

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+                    .status !== "PUBLISHED"
+              );
+        return metric.enabled && recordExistsForMetricAndUnpublished;
+      })
+      .map((metric) => {
+        const metricFrequency = metric.custom_frequency || metric.frequency;
+        const startingMonth = metric.starting_month;
+        const latestMonthlyMetricValue =
+          metricFrequency === "MONTHLY" &&
+          latestMonthlyAnnualRecordsMetadata?.monthly?.metrics?.find(
```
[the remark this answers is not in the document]
alfonsogoberjr: This might be a good place to recommend a functional library like [Ramda](https://ramdajs.com/docs/).

This code may be more efficiently expressed like so:

```Typescript
import { pipe, map, filter, groupBy } from 'ramda';

const metricEnabled = ({ enabled }: Metric) => enabled;

const recordExistsForMetricAndUnpublished = (metric: Metric) => {
    const metricFrequency = metric.custom_frequency || metric.frequency;
    const startingMonth = metric.starting_month;
    /** Does a record exist for this current metric and is that record unpublished? */
    return metricFrequency === "MONTHLY"
        ? Boolean(
            latestMonthlyAnnualRecordsMetadata?.monthly?.id &&
              latestMonthlyAnnualRecordsMetadata.monthly.status !==
                "PUBLISHED"
          )
        : startingMonth &&
          Boolean(
            latestMonthlyAnnualRecordsMetadata?.annual[startingMonth] &&
              latestMonthlyAnnualRecordsMetadata?.annual[startingMonth]
                .status !== "PUBLISHED"
          );
}

const enabledMetricsTaskCardMetadata: TaskCardMetadata[] = pipe(
    filter(metricEnabled),
    filter(recordExistsForMetricAndUnpublished),
    map((metric) => {
        const metricFrequency = metric.custom_frequency || metric.frequency;
        const startingMonth = metric.starting_month;

        const latestMonthlyMetricsHash = pipe(path(['monthly', 'metrics']), groupBy(({ key }) => key))(latestMonthlyAnnualRecordsMetadata)
        const latestAnnualMetricsHash = pipe(path(['annual', startingMonth, 'metrics']), groupBy(({ key }) => key))(latestMonthlyAnnualRecordsMetadata)

        // ...so on and so forth for all your `find()` calls, each one should be a pure function with no side-effects
        const latestAnnualMetricValue =
          metricFrequency === "ANNUAL" &&
          startingMonth &&
          latestAnnualMetricsHash[metric.key]?.value;

        const hasMetricValue =
          metricFrequency === "MONTHLY"
        ? Boolean(metricFrequency === "MONTHLY" && latestMonthlyMetricsHash[metric.key]?.value)
            : Boolean( metricFrequency === "ANNUAL" &&
          startingMonth && latestAnnualMetricsHash[metric.key]?.value);
        const reportID =
          metricFrequency === "MONTHLY"
            ? latestMonthlyAnnualRecordsMetadata?.monthly?.id
            : startingMonth &&
              latestMonthlyAnnualRecordsMetadata?.annual[startingMonth]?.id;
        return {
            reportID,
            title: metric.display_name,
            description: metric.description,
            actionLinks: hasMetricValue
                  ? [taskCardLabelsActionLinks.publish]
                  : [
                      taskCardLabelsActionLinks.uploadData,
                      taskCardLabelsActionLinks.manualEntry,
                    ],
            metricFrequency,
            hasMetricValue,
        };
        })
    )(currentAgencyMetrics)

}) || [];
```

Something like that. Might help organize the logic better

publisher/src/components/Reports/DataEntryForm.tsx
mxosman: Good question - I'm rethinking this solution... Originally, I wanted to avoid adding a scrollable container in the home page (mostly to avoid previous situations where we've had nested scrollable containers) - but, that might be a better solution and could avoid having to update this throughout all together. I think adding it to every page might introduce more complications.

I have to remove this from here anyway when I implement scrolling to the metric from the task card. Thanks for calling this out!

publisher/src/components/Home/Home.tsx
mxosman: Yes - it should differentiate between the two - all based on starting month. So you should see two cards!

publisher/src/components/Home/Home.tsx
mxosman: Great point! I thought about this when implementing it and I don't believe we'll get into that state, but I can't quite remember my reasoning. Either way - let me see if I can tighten this up so there's absolutely no possibility of that happening.

publisher/src/components/Home/Home.tsx
mxosman: Great question! So, I guess what I'm trying to convey here is - the response JSON is in this format `{ <starting month>: AnnualRecord }` and we're keeping the same structure, except instead of an `AnnualRecord` it'll be an `AnnualRecordMetadata` object.

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+        const latestAnnualMetricValue =
+          metricFrequency === "ANNUAL" &&
+          startingMonth &&
+          latestMonthlyAnnualRecordsMetadata?.annual[
+            startingMonth
+          ]?.metrics?.find(
+            (latestAnnualMetric) => latestAnnualMetric.key === metric.key
+          )?.value;
+        const hasMetricValue =
+          metricFrequency === "MONTHLY"
```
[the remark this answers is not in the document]
mxosman: Let me try that out! I didn't split it up into two conditionals because I thought returning one object was clearer but your proposal feels like it might be cleaner.

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,369 @@
...
+  const allTasksCompleteTaskCardMetadata: TaskCardMetadata = {
+    title: "All tasks complete",
+    description: "Your data is updated and published.",
+  };
+  const enabledMetricsTaskCardMetadata: TaskCardMetadata[] =
+    currentAgencyMetrics
+      ?.filter((metric) => {
+        const metricFrequency = metric.custom_frequency || metric.frequency;
+        const startingMonth = metric.starting_month;
+        /** Does a record exist for this current metric and is that record unpublished? */
```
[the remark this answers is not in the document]
mxosman: Absolutely! Love that - thanks for the improvement!

publisher/src/components/Home/Home.tsx
mxosman: I was thinking if you've configured all of your metrics and mark them all as unavailable - then ideally you should be in the `All Tasks Completed` state. Also, if you've input data for all metrics and the latest monthly and annual reports are published - you should be in the `All Tasks Completed` state. Please let me know if that makes sense/I'm forgetting to consider something!

mxosman: I'll make this clearer and add some more comments.

publisher/src/components/Home/Home.tsx
mxosman: Exactly!! We start out with 2 lists - unconfigured metrics and enabled metrics. Then, we look for metrics with values - so we can collapse them into one Publish report card - and metrics without values + unconfigured metrics so we can render those with their individual task cards.

publisher/src/components/DataUpload/DataUpload.tsx
mxosman: Response below: https://github.com/Recidiviz/justice-counts/pull/682/files#r1238898562

publisher/src/components/Home/Home.tsx
mxosman: Here we're going through the metrics within the associated Annual Report to find the matching metric and then checking that metric's value, if that makes sense?

publisher/src/components/Home/Home.tsx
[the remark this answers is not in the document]
mxosman: > Almost certainly doesn't really matter, but whenever we call `find` (O(n)) from within a loop (also O(n)) that raises a flag to me, because that means we are doing something that is O(n^2) complexity (loop within a loop). To avoid this, we could create a set containing all of the metric keys in `latestMonthlyAnnualRecordsMetadata?.monthly?.metrics` once above, and then here we could just check if `metric.key` is in that set (which is only O(1)). LMK if that makes sense!

Totally makes sense - thanks for always keeping an eye out for this! Things are definitely feeling unwieldy here!

> This might be a good place to recommend a functional library like [Ramda](https://ramdajs.com/docs/).
> 

Alfonso - I really appreciate the thorough feedback here and library recommendation! I never used Ramda before. This looks incredibly useful and your explanation and code example make sense to me! I'm going to knock out some of the other feedback here and would like to explore Ramda more and refactor this logic using it either in this PR or a follow up PR. Looking forward to playing around with this.

publisher/src/components/Home/Home.tsx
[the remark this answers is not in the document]
mxosman: So much clearer - thanks, Lili! [d207a0c](https://github.com/Recidiviz/justice-counts/pull/682/commits/d207a0c97738a65d4c968be6cd8254427268995e)

publisher/src/components/Home/Home.tsx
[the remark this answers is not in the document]
mxosman: Here's what this would look like:
[0a8e5e5](https://github.com/Recidiviz/justice-counts/pull/682/commits/0a8e5e532b79615fae64a70d7330ed61f3ca9ae8)

It feels cleaner in many ways but also feels redundant and adds more places to update if something needs to adjust with the Metadata structure. I'm not sure - always curious to hear your thoughts!

publisher/src/components/Home/Home.tsx
mxosman: Added some clarifying comments here for now: [0b84b88](https://github.com/Recidiviz/justice-counts/pull/682/commits/0b84b88d44b76d67779a35ae72cbc5d53ba5ac5e)

Will update if there's something I'm missing/not considering!

publisher/src/components/Home/Home.tsx
mxosman: Thank you for the feedback here! Should feel more secure: [9f4cda7](https://github.com/Recidiviz/justice-counts/pull/682/commits/9f4cda70348009fd7aed9d5afa48829e77b2a229)

publisher/src/components/Reports/DataEntryForm.tsx
mxosman: Added a scrollable container: [97f972d](https://github.com/Recidiviz/justice-counts/pull/682/commits/97f972d3cb4d6097167f56e4d1a45dc52f4e10b5)

This avoids us having to do the scroll-to-top in other areas, but it cuts off oddly at the bottom because of the footer. Since the footer is going to be solidly colored (I believe that PR is merged) - the cut off won't look so odd.


https://github.com/Recidiviz/justice-counts/assets/59492998/22eb8805-d306-4a35-bf27-4028ed0db2c9

publisher/src/components/Home/Home.tsx
```diff
@@ -15,14 +15,264 @@
...
+        )
+      ) || [];
+
+  /**
+   * User has completed all tasks if:
+   *  1. User has configured all metrics and set them all to "Not Available"
+   *  2. User has entered values for all metrics in the latest annual and/or monthly
+   *     records and those records are published
+   */
+  const hasCompletedAllTasks = () => {
```
lilidworkin, line 107: Feel free to ignore this comment (definitely not merge blocking) but I still find this logic a little confusing (although your comments are great). It seems to me it would be clearer to first run a function that generates all task cards. And if this function returns an empty list, then we know that we are in the completed all tasks state. No need to have separate logic for generating tasks cards and determining completed state, since the latter just depends on the former. This approach also feels less brittle, since we might change the logic about what cards get generated when, and right now that means we'd have to also update this function too. Does that makes sense, or is there a reason that might be difficult?

publisher/src/components/Home/helpers.ts
```diff
@@ -0,0 +1,204 @@
...
+  AnnualRecordMetadata,
+  LatestRecordMetadata,
+  taskCardLabelsActionLinks,
+  TaskCardMetadata,
+  TaskCardMetadataValueConfigurationGroup,
+} from ".";
+
+export const metricEnabled = ({ enabled }: Metric) => enabled;
+export const metricNotConfigured = ({ enabled }: Metric) => enabled === null;
+export const metricIsMonthly = (metric: Metric) =>
```
lilidworkin, line 35: Is this right? What if metric.custom_frequency is Annual and metric.frequency is Monthly -- then this function will return true?

publisher/src/stores/ReportStore.ts
```diff
@@ -165,6 +166,40 @@ class ReportStore {
...
+      const latestReportsAndMetrics =
+        (await response.json()) as LatestReportsAgencyMetrics;
+      const allRecords = [
+        latestReportsAndMetrics.monthly_report,
+        ...Object.values(latestReportsAndMetrics.annual_reports),
+      ];
+      if (allRecords.length > 0) {
+        allRecords.forEach(
+          (record) =>
+            record.id &&
```
lilidworkin, line 193: nit/random, but I'm curious if you were seeing records that didn't have an ID? I feel like that shouldn't ever happen

lilidworkin (approved):
Awesome, thank you for addressing all of this feedback and for so carefully considering various ways of organizing the logic! I think this looks great. I did leave one suggestion for a slightly larger change, but even if you do decide to do that, I'd merge this guy in and make that change in a separate PR.

publisher/src/components/Home/Home.tsx
mxosman: That makes a lot of sense - let me think about this more!

publisher/src/components/Home/helpers.ts
mxosman: Ah shoot! That's a really great point. You're absolutely right.

Maybe we can say:
```
export const metricIsMonthly = (metric: Metric) => metric.custom_frequency === "MONTHLY" || (!metric.custom_frequency && metric.frequency === "MONTHLY");
```

This way - we check the `custom_frequency` first - if that's falsy, we will check the metric's natural frequency only if a `custom_frequency` has not been set. There's other ways to express this too, but LMK if this makes sense?

publisher/src/stores/ReportStore.ts
mxosman: I haven't ran across a record that didn't have an `id` - this was likely me satisfying TypeScript - I'll double check if I still need it though! I think one of the typed objects has this as potentially undefined, but I changed some stuff around, so this might not be necessary.

publisher/src/components/Home/Home.tsx
lilidworkin: Cool! Definitely no need to do in this PR

publisher/src/components/Home/helpers.ts
lilidworkin: Yep this looks right!

publisher/src/stores/ReportStore.ts
lilidworkin: Oh cool no it's totally fine to do it for TS, I just wanted to make sure you didn't add this in due to behavior you were seeing

publisher/src/components/Home/helpers.ts
mxosman: Thank you a million for catching this! Phew...

publisher/src/components/Home/Home.tsx
mxosman: Actually - based on the way we create the cards (and our filtering), we don't do need to do the extra checks here! We can just check the below as you proposed (determined based on the output of the task cards generated):

```
  const hasCompletedAllTasks =
    enabledMetricsTaskCardMetadata.length === 0 &&
    unconfiguredMetricsTaskCardMetadata.length === 0;
```

I tested this and it seems to be working as expected. Was this in-line with what you had in mind?

lilidworkin: Yes, that's perfect!

PR #704, opened 2023-06-07
[21327] Add Segment tracking functions based on design doc

alfonsogoberjr:
## Description of the change

Adds tracking functions according to the questions posed [here](https://paper.dropbox.com/doc/Design-Doc-Internal-Dashboard--B5z6zfCM19i6PAV6KGrDfwnrAg-R22hiIqWxF0SJvvQ8kojf)

## Related issues

Closes #21327

### Development

### Code review


publisher/src/analytics.ts
```diff
@@ -194,3 +194,44 @@ export const trackLoadTime = (
...
   });
 };
+
+type TrackingInput = {
+  agencyId: string;
+  metricKey: string;
+};
+
+export const trackLogin = ({ agencyId }: Partial<TrackingInput>): void => {
+  track("publisher_login", {
```
lilidworkin, line 204: I'm wondering how this will interact with the `identify` call we make [here](https://github.com/Recidiviz/justice-counts/blob/bbd9a2d6918e0317cda7c64e6d6ac064397c9245/publisher/src/components/Auth/AuthStore.ts#L101). My understanding is that this also gets called after the user logs in. Or actually maybe it gets called every time you load the page, I'm not sure. Anyway, would be good to figure out if we need both, and in what order they're going to be called.

I guess that also raises another question -- what do we actually mean by "login" -- do we mean you filled out your username/email on the Auth0 modal? What if you login, close your window, and then re-open Publisher -- you'll probably still be logged in -- but do we want to treat that as a separate login? I'm not sure it matters too much and I don't have strong feelings either way, we should just make sure we're clear on the behavior and document it somewhere. (Might be good to get a doc started with a list of the events we're tracking as well as a pointer in the code to where they're tracked?)

lilidworkin: Oh also -- super nit, but I'm not sure we need the "publisher_" prefix on these events? Since they're all being logged from the Publisher app?

publisher/src/analytics.ts
```diff
@@ -194,3 +194,44 @@ export const trackLoadTime = (
...
+    agencyId,
+  });
+};
+
+export const trackMetricConfiguration = ({
+  agencyId,
+  metricKey,
+}: Partial<TrackingInput>): void => {
+  track("publisher_metric_configured", {
```
lilidworkin, line 213: As per earlier conversations -- do we need to also include other information in this event, like the total number of metrics configured so far? Otherwise in order to get that information, will we need to do a sum across events of this type? Which might be fine -- just wanted to confirm/clarify.

publisher/src/analytics.ts
```diff
@@ -194,3 +194,44 @@ export const trackLoadTime = (
...
+    agencyId,
+    metricKey,
+  });
+};
+
+export const trackMetricEnabled = ({
```
lilidworkin, line 219: Now that I think about it, we might be able to get away with just the trackMetricConfigured event, and include a boolean flag for enabled/disabled. Since configuring a metric means (for our purposes here -- in reality it's a little more complicated) either enabling or disabling it (the default state is null, neither enabled nor disabled). So when they enable, we send a configured event with enabled = True, and when they disable, we send a configured event with enabled = False.

Later we'll have to figure out if / how we also want to track enabling/disabling of diaggregations and dimensions. I think we can punt on that for now, but it might be good to chat with @mxosman to get a better understanding of how that part of the flow works.

lilidworkin (commented):
Excited to see this getting started!

lilidworkin:
@alfonsogoberjr quick thing to note, not for this PR -- I know we already have tracking events for creating and publishing reports, but I don't think those events are currently getting called in all the right places (since the code has changed a lot since this was initially written and we've shipped new ways to create reports, like bulk upload, and new ways to publish reports, like bulk publishing). Just something to be aware of -- both @mxosman and I can help identify where the updates need to be made.

mxosman:
This is super exciting, Alfonso! I'm looking forward to seeing these specific trends - happy to help in anyway!

publisher/src/analytics.ts
lilidworkin: @alfonsogoberjr any thoughts on my other comments above? 

1) When will we call `trackLogin` and how is this the same/different from `identify`?
2) Should we call this when the user fills out username/password or when they start a new session?
3) Do we need the `publisher_` prefix?

Curious if @mxosman has thoughts too!

publisher/src/analytics.ts
lilidworkin: @alfonsogoberjr any thoughts on whether it's better/worse to have a single `trackMetricConfiguration` event with an enabled/disabled flag, or to keep these as three separate events?

publisher/src/analytics.ts
```diff
@@ -194,3 +196,44 @@ export const trackLoadTime = (
...
+  });
+};
+
+export const trackMetricConfiguration = ({
+  agencyId,
+  updatedMetricSettings,
+}: Partial<TrackingInput>): void => {
+  track("metric_configured", {
+    agencyId,
+    updatedMetricSettings,
```
lilidworkin, line 217: I'm still not sure I understand how we'll be able to use these events to determine how many metrics have been configured overall. Does the `updatedMetricSettings` include that information, or does it just include information about this individual metric?

publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
```diff
@@ -98,6 +99,13 @@ function MetricAvailability() {
         metricSearchParam,
         enabledStatus
       );
+      if (enabledStatus)
```
lilidworkin, line 102: It looks like `handleUpdateMetricEnabledStatus` gets called when the user clicks the "Not Available" button (i.e. turning a metric off) but that `handleUpdateMetricReportFrequency` gets called when a user clicks the "Monthly" or "Annual" button (i.e. turning a metric on). I don't think `handleUpdateMetricEnabledStatus` ever gets called with `true`? If so we'll need to adjust this and call `trackMetricEnabled` from `handleUpdateMetricReportFrequency`  instead.

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -222,6 +223,7 @@ class MetricConfigStore {
       method: "PUT",
     })) as Response;
 
+    trackMetricConfiguration({ agencyId, updatedMetricSettings });
```
lilidworkin, line 226: So this is going to get called every time a user makes _any_ change to the metric configuration page -- is that what we want? I thought maybe we just wanted to call this when a user configures their metric (i.e. turns it on or off) for the very first time?

lilidworkin (changes requested):
Thanks @alfonsogoberjr ! I think this is heading in the right direction, but I think there are some issues with the logic. Relatedly, I like to ask all engineers to include a "test plan" of some sort to their PR descriptions -- that can either be something like "I added unit tests" or "here's a screenshot/video of how it looks now" or "here are some console.logs that prove that it works." Could you add more detail to your PR description explaining exactly when you expect each of these events to be called, and then show some proof that they were actually called then? There's a TEST_SENDING_ANALYTICS flag that you can set that will log to the console when the event would have actually gotten sent to Segment, which might help!

cc @mxosman who should definitely review this PR as well!

lilidworkin (commented):
@alfonsogoberjr another quick thought about this PR:

I realized the scope/intention is a little unclear to me -- is this meant to be the first PR in a series (i.e. this one just adds tracking functions to analytics.js, but another one actually starts calling those functions in all the right places?) or is it meant to be comprehensive, i.e. all the changes we need to make in order to support the dashboard?

Two things might be helpful here:
1) Adding a more descriptive summary in the PR description to clarify the goal (generally we try to add as much context in the description as possible, though I'm certainly guilty of dropping the ball on this, haha)
2) Making a list of all the P0 items from the design doc, and annotating them with the Segment event we'll use to calculate it. This will help us see if we have full coverage, and if each event contains the right information.

publisher/src/analytics.ts
alfonsogoberjr: Agreed, I realize now the `enabled` field exists on the MetricSettings object

publisher/src/analytics.ts
alfonsogoberjr: I've been thinking about it - since a purely Segment-based accounting isn't going to work, we need some kind of metadata schema stored in our DB instead of ad-hoc requests across multiple data sources. So this PR wouldn't contain that just yet, until the outcome of our meeting is decided.

lilidworkin: Ok cool makes sense!

publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
mxosman: Great catch, Lili! You're right that `handleUpdateMetricEnabledStatus` only gets called when the `enabled` status is false. It's not immediately intuitive and I'm sure we can probably improve how we're naming things - but the selection options for a metric configuration are: 
* Not Available (which uses the `handleUpdateMetricEnabledStatus` function) 
* Monthly (which used the `handleUpdateMetricReportFrequency` function)
* Annual (which used the `handleUpdateMetricReportFrequency` function)

I think a better candidate for this would be the `saveMetricSettings` method in the `MetricConfigStore` which is called by both handler functions above.

publisher/src/analytics.ts
mxosman: If we're still pursuing segment and these tracking functions -- since the `MetricSettings` object can be verbose, I'm wondering if it would simplify things to just extract the `key` and `enabled` properties and use those two to track metric configuration here?

publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
lilidworkin: @mxosman if we call it in `saveMetricSettings` though, won't it get called whenever metric config is updated, not just when metrics are enabled/disabled?

mxosman: That's a good point - though there aren't any other good centralized candidates I can think of to place this - aside from adding the tracking function in each of the above handler functions. If we wanted to put it in a central area like `saveMetricSettings` we could add conditionals to make sure it only runs if/when the enabled property is present in the `updatedMetricSettings` object.

publisher/src/analytics.ts
lilidworkin: @alfonsogoberjr any thoughts on @mxosman 's comment above?

lilidworkin (changes requested):
@alfonsogoberjr , in [this comment](https://github.com/Recidiviz/justice-counts/pull/704#pullrequestreview-1475109244) I asked if you could add a test plan of some sort to your PR, and in [this comment](https://github.com/Recidiviz/justice-counts/pull/704#pullrequestreview-1477105756) I asked if you could make your PR description more detailed in general about the goal/scope of the PR. Thank you!

publisher/src/analytics.ts
alfonsogoberjr: Sorry I missed this comment - that sounds great! I'll update

publisher/package.json
```diff
@@ -9,6 +9,7 @@
     "@auth0/auth0-react": "^1.12.0",
     "@justice-counts/common": "*",
     "@recidiviz/design-system": "^4.1.0",
+    "@storybook/testing-library": "^0.1.0",
```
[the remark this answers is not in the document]
alfonsogoberjr: Linter wouldnt pass without it, for some reason

publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
alfonsogoberjr: Hey I missed these comments with all of the other feedback going around. This makes sense.

publisher/src/components/Reports/CreateReport.tsx
```diff
@@ -14,6 +14,23 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
+/* eslint-disable import/no-extraneous-dependencies */
```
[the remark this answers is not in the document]
alfonsogoberjr: weird, fixing

publisher/package.json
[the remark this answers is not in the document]
alfonsogoberjr: reset branch from main, something must have come over from the storybook branch

alfonsogoberjr:
closed this because of too many concurrent discussions. All requested changes were implemented in new PR [here](https://github.com/Recidiviz/justice-counts/pull/713). There should be no storybook code now

PR #745, opened 2023-06-28
Home Page: Link “Manual Entry” action link to Create Report page for users who do not have a records [5/n]

### LINKED ISSUE #744: [Frontend] Link “Manual Entry” action link to Create Report page for users who do not have a report

mxosman:
## Description of the change

In PR #682 - there was a temporary solution in place for users without records (or without a record tied to a metric - e.g. a monthly metric but the user has 0 monthly records) - if there is no record, the task card doesn't appear. 

After discussing with product and design, we realized a better solution would be to still display all metrics (because users without records can still technically upload a spreadsheet), but link the "Manual Entry" action link to the Create Record page if a metric doesn't have any records to link to.

https://github.com/Recidiviz/justice-counts/assets/59492998/eea4ec5a-6afe-4864-bca5-3bdb60cb5bb5

## Related issues

Closes #744 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/Home.tsx
+++ b/publisher/src/components/Home/Home.tsx
@@ -84,7 +84,7 @@ export const Home = observer(() => {
           { latestMonthlyRecord, latestAnnualRecord },
           createDataEntryTaskCardMetadata
         )
-      ) || [];
+      );
   const unconfiguredMetricsTaskCardMetadata: TaskCardMetadata[] =
     currentAgencyMetrics
       .filter(metricNotConfigured)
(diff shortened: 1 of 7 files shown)


lilidworkin (approved):
Nice, I like this a lot! Can we just quickly test what the experience is for a non-agency-admin, or for someone with read-only permissions? I think neither of those roles is allowed to create reports (double check me on that though). But I'm not sure if that means that the page doesn't load at all for them, or if it means that the page loads, but the "Create Record" button is disabled on it. As long as the experience doesn't break for them, it's fine!

mxosman:
> Nice, I like this a lot! Can we just quickly test what the experience is for a non-agency-admin, or for someone with read-only permissions? I think neither of those roles is allowed to create reports (double check me on that though). But I'm not sure if that means that the page doesn't load at all for them, or if it means that the page loads, but the "Create Record" button is disabled on it. As long as the experience doesn't break for them, it's fine!

Oh great call!! Thank you for thinking of this. Right now, if a user who is a non-agency-admin tries to go to the create record page, they get redirected to the records page. It looks fine to me when there are no records in the record page - but, does it look weird when there is a record or mroe?

No Record:

https://github.com/Recidiviz/justice-counts/assets/59492998/0b1a7a20-fe34-47a5-a72f-a95e7ee53301

With Record:

https://github.com/Recidiviz/justice-counts/assets/59492998/2b254809-9188-4c89-8edf-c9d7c7841278

lilidworkin:
Thanks for testing -- this looks fine to me!!

PR #751, opened 2023-07-05
New Homepage: Bug Fixes [6/n]

mxosman:
## Description of the change

Random bug fixes from testing:
* Fix links back to home to include agency ID
* Fix crash issue due to `latestRecordsAndMetrics` logic in `ReportStore` - properly handles cases when no monthly &/ annual records are sent
* Fix switching agencies bug by updating the `currentSystem` when user switches agencies
* Fix issue with `hasCompletedAllTasks` - now determined by the task card arrays used for rendering (prev. solution was not sufficient for all cases)

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/Home.tsx
+++ b/publisher/src/components/Home/Home.tsx
@@ -124,7 +124,6 @@ export const Home = observer(() => {
     allMetricMetadatasWithoutValuesOrNotConfigured.length === 0 &&
     allMetricMetadatasWithValues.ANNUAL.length === 0 &&
     allMetricMetadatasWithValues.MONTHLY.length === 0;
-
   const welcomeDescription = !hasCompletedAllTasks
     ? "See open tasks below"
     : "Dashboards are updated with the latest published records";
@@ -215,8 +214,7 @@ export const Home = observer(() => {
               {/* Publish-Ready Cards (for Monthly & Annual Records) */}
               {/* Publish latest monthly record */}
               {allMetricMetadatasWithValues.MONTHLY.length > 0 &&
-                latestMonthlyRecord &&
-                latestMonthlyRecord.status !== "PUBLISHED" && (
+                latestMonthlyRecord && (
                   <TaskCard
                     metadata={createPublishTaskCardMetadata(
                       latestMonthlyRecord.reportTitle,
@@ -230,7 +228,6 @@ export const Home = observer(() => {
                 latestMonthlyAnnualRecordsMetadata?.annual &&
                 Object.values(latestMonthlyAnnualRecordsMetadata.annual).map(
                   (record) => {
-                    if (record.status === "PUBLISHED") return null;
                     return (
                       <TaskCard
                         key={record.id}
--- a/publisher/src/components/Home/Home.tsx
+++ b/publisher/src/components/Home/Home.tsx
@@ -84,7 +84,7 @@ export const Home = observer(() => {
           { latestMonthlyRecord, latestAnnualRecord },
           createDataEntryTaskCardMetadata
         )
-      ) || [];
+      );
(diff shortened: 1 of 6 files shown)


publisher/src/components/Home/Home.tsx
```diff
@@ -95,21 +95,7 @@ export const Home = observer(() => {
           { latestMonthlyRecord, latestAnnualRecord },
           createConfigurationTaskCardMetadata
         )
-      ) || [];
```
mxosman, line 98: `|| []` were unnecessary as `currentAgencyMetrics` will always be initialized with an empty array.

publisher/src/components/Home/Home.tsx
```diff
@@ -128,6 +114,22 @@ export const Home = observer(() => {
...
+  /**
+   * User has completed all tasks if:
+   *  1. User has configured all metrics and set them all to "Not Available"
+   *  2. User has entered values for all metrics in the latest annual and/or monthly
+   *     records and those records are published
+   */
+  const hasCompletedAllTasks =
+    allMetricMetadatasWithoutValuesOrNotConfigured.length === 0 &&
+    allMetricMetadatasWithValues.ANNUAL.length === 0 &&
+    allMetricMetadatasWithValues.MONTHLY.length === 0;
```
mxosman, line 126: These are the proper arrays to use to determine this, since we use these to render the task cards - before I mistakenly used the `enabledMetricsTaskCardMetadata` & `unconfiguredMetricsTaskCardMetadata` which still underwent more processing via the `groupMetadatasByValueAndConfiguration`, so it wasn't a reliable source.

publisher/src/components/Home/helpers.ts
```diff
@@ -191,7 +193,7 @@ export const groupMetadatasByValueAndConfiguration = (
     (acc, metric) => {
       const { metricFrequency } = metric;
       if (metric.hasMetricValue) {
-        if (metricFrequency)
+        if (metricFrequency && metric.status !== "PUBLISHED")
```
mxosman, line 196: Exclude metrics with values that are in a published report. We can do this check here and not append a task card to the `allMetricMetadatasWithValues` instead of doing this check in the render.

publisher/src/stores/ReportStore.ts
```diff
@@ -181,18 +181,26 @@ class ReportStore {
...
+      );
+      const hasAnnualRecords = Boolean(annualRecords.length > 0);
+      const hasMonthlyRecord = Boolean(
+        latestRecordsAndMetrics.monthly_report.id
+      );
+      const allRecords = [];
+
+      if (hasMonthlyRecord)
+        allRecords.push(latestRecordsAndMetrics.monthly_report);
+      if (hasAnnualRecords) allRecords.push(...annualRecords);
```
mxosman, line 197: The below (L198 - 202) would sometimes have an empty `{}` in the `allRecords` array which caused some weird behavior - this properly adds the records if they exist to the `allRecords` array.

publisher/src/components/Home/Home.tsx
lilidworkin: Cool, thanks for catching!

publisher/src/components/Home/helpers.ts
lilidworkin: Oh nice, so this change goes along with the changes in Home.tsx right above?

lilidworkin (approved):
Cool! Thanks for testing so thoroughly and catching these!

publisher/src/components/Home/helpers.ts
mxosman: Exactly! This ensures that those arrays only contain the metadatas being rendered - making it a safer check of whether or not a user has completed all tasks.

PR #759, opened 2023-07-11
New Homepage: Playtesting Followups (P0)

### LINKED ISSUE #760: [Frontend] New Homepage + Nav: Playtesting Followups (P0)

mxosman:
## Description of the change

There was an issue with supervision subsystem metric task cards that were incorrectly linking the "Manual Entry" to the Create Record page (thank you for catching this, Michelle!). This was because the supervision subsystem metric objects have a default `null` value in the `starting_month` property. When we create task cards, we need a `starting_month` for annual metrics in order to link them to the appropriate latest annual record.

To solve for this, when we create task cards, before assigning the `startingMonth` we can do a boolean check of `metric.disaggregated_by_supervision_subsystems`. If it's `true`, then we can assume this is a supervision subsystem that will have a `null` value for `starting_month`, so we check the parent `SUPERVISION_<METRIC KEY>` metric object for the `starting_month`. If it's `false`, then we can assume this is not a supervision subsystem, and should have a value for `starting_month`. Do these assumptions feel reliable?

## Related issues

Closes #760

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/helpers.ts
+++ b/publisher/src/components/Home/helpers.ts
@@ -151,11 +151,23 @@ export const createPublishTaskCardMetadata = (
 };
 
 /**
+ * Strips the system name from a given metric key and returns the raw metric key.
+ *
+ * @example
+ * metricKey = "PAROLE_FUNDING"
+ * systemKey = "PAROLE"
+ * returns "FUNDING"
+ */
+const stripSystemKeyFromMetricKey = (metricKey: string, systemKey: string) =>
+  metricKey.replace(`${systemKey}_`, "");
+
+/**
  * Creates configuration and data entry task card metadata objects from metrics linked to
  * their corresponding latest record. This function uses the above `createConfigurationTaskCardMetadata` and
  * `createDataEntryTaskCardMetadata` as callback functions to create the appropriate task card metadata object.
  */
 export const createTaskCardMetadatas = (
+  currentAgencyMetricsByMetricKey: Record<string, Metric[]>,
   metric: Metric,
   recordMetadatas: {
     latestMonthlyRecord: LatestRecordMetadata | undefined;
@@ -169,7 +181,19 @@ export const createTaskCardMetadatas = (
   ) => TaskCardMetadata
 ) => {
   const metricFrequency = metric.custom_frequency || metric.frequency;
-  const startingMonth = metric.starting_month;
+  /**
+   * Supervision subsystems metrics have a default `null` value for `starting_month`.
+   * If this is a supervision subsystem, use the parent supervision metric's starting month,
+   * otherwise use the current metric's starting month.
+   */
+  const startingMonth = metric.disaggregated_by_supervision_subsystems
+    ? currentAgencyMetricsByMetricKey[
+        `SUPERVISION_${stripSystemKeyFromMetricKey(
+          metric.key,
+          metric.system.key
+        )}`
+      ]?.[0].starting_month
+    : metric.starting_month;
   const { latestMonthlyRecord, latestAnnualRecord } = recordMetadatas;
   /** Create Task Card linked to the latest Monthly Record */
   if (metricFrequency === "MONTHLY") {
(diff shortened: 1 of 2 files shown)


lilidworkin:
@mxosman hmm, something feels off here to me. Why is starting month null for supervision subsystem metrics? If you disaggregate Funding, and then set Funding (Parole) to Calendar Year and Funding (Probation) to Fiscal Year, what happens? It seems to me that those should have different starting months. FWIW I don't think this will be a common occurrence lol but the fact that it's not working seems to me like it could be a symptom of something more problematic going on. Do you think you could dig in a little further? Maybe looking at the API requests/responses and also checking what is saved in the DB?

Feel free to pull in @morden35 to help with the backend part of this investigation!

mxosman:
> @mxosman hmm, something feels off here to me. Why is starting month null for supervision subsystem metrics? If you disaggregate Funding, and then set Funding (Parole) to Calendar Year and Funding (Probation) to Fiscal Year, what happens? It seems to me that those should have different starting months. FWIW I don't think this will be a common occurrence lol but the fact that it's not working seems to me like it could be a symptom of something more problematic going on. Do you think you could dig in a little further? Maybe looking at the API requests/responses and also checking what is saved in the DB?

Such a great call out, Lili - thank you for the caution here!

>Why is starting month null for supervision subsystem metrics?

Using the Funding metric for example - when you go to metric settings and Supervision: Funding metric is marked as "Annual", and you proceed to disaggregate it, the Funding (Parole) and Funding (Probation) don't inherit the starting month - their starting months are still `null` until you go in to each one and set it. I think we can still get around this in the FE and infer the starting month if one isn't set - but should we explore updating the BE to have disaggregated annual metrics inherit the starting month? (Happy to explore this if it makes sense to do [and ping you @morden35 if I get lost])

>If you disaggregate Funding, and then set Funding (Parole) to Calendar Year and Funding (Probation) to Fiscal Year, what happens? It seems to me that those should have different starting months.

This is correct - they will have starting months as soon as you set them!

lilidworkin:
> I think we can still get around this in the FE and infer the starting month if one isn't set - but should we explore updating the BE to have disaggregated annual metrics inherit the starting month? 

Oh great point -- I think for now we can just infer on the FE! That feels fine to me. Use the starting month if it's set, else default to the parent's. Thank you @mxosman !

publisher/src/components/Home/helpers.ts
```diff
@@ -150,12 +150,24 @@ export const createPublishTaskCardMetadata = (
   };
 };
 
+/**
+ * Strips the system name from a given metric key and returns the raw metric key.
+ *
+ * @example
+ * metricKey = "PAROLE_FUNDING"
```
lilidworkin, line 157: Love this docstring

publisher/src/components/Home/helpers.ts
```diff
@@ -169,7 +181,20 @@ export const createTaskCardMetadatas = (
...
+   * use the current metric's starting month.
+   */
+  const startingMonth =
+    metric.disaggregated_by_supervision_subsystems && !metric.starting_month
+      ? currentAgencyMetricsByMetricKey[
+          `SUPERVISION_${stripSystemKeyFromMetricKey(
+            metric.key,
+            metric.system.key
+          )}`
+        ]?.[0].starting_month
```
lilidworkin, line 196: It's safe to assume this list will always only have one element, right?

lilidworkin (approved):
Looks great, thanks for handling this super weird edge case @mxosman !

publisher/src/components/Home/helpers.ts
mxosman: Great question! Yes - it should be safe. The `currentAgencyMetricsByMetricKey` is using the `groupBy` helper method to group each metric within the agency metrics array by their metric key. Though - not a bad idea to handle that case since this could crash if somehow it's an empty array. Great call out!

mxosman:
Thank you for all your help and sanity checking!

PR #762, opened 2023-07-11
New Homepage: Playtesting Followups (P1)

### LINKED ISSUE #761: [Frontend] New Homepage + Nav: Playtesting Followups (P1)

mxosman:
## Description of the change

Addresses the following P1 items:
* Fixes “(No Enabled Metrics)” text overflow in Data Viz page
* Adds system name in parenthesis on metric task cards for multi-system agencies (in the “All” filtered stated)
* Fixes Tooltip’s capitalized text-transform
* Updates “Your data is updated and published” copy to “Your data is up-to-date and published”
* Removes Metric Settings button
* Fix Publish review metric checkmarks (sometimes they show, other times they do not)

Demo:

https://github.com/Recidiviz/justice-counts/assets/59492998/cd67bc62-8cbf-4c02-8995-930a2d6cc863

## Related issues

Closes #761 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/Home.tsx
+++ b/publisher/src/components/Home/Home.tsx
@@ -15,7 +15,11 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import { AgencySystems, Metric } from "@justice-counts/common/types";
+import {
+  AgencySystems,
+  Metric,
+  SupervisionSubsystems,
+} from "@justice-counts/common/types";
 import { groupBy } from "@justice-counts/common/utils";
 import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
@@ -160,6 +164,7 @@ export const Home = observer(() => {
       const monthlyRecordMetadata = hasMonthlyRecord
         ? createMonthlyRecordMetadata(monthlyRecord)
         : undefined;
+
       setLatestMonthlyAnnualsRecordMetadata({
         monthly: monthlyRecordMetadata,
         annual: annualRecordsMetadata,
@@ -177,6 +182,28 @@ export const Home = observer(() => {
     return <Loading />;
   }
 
+  const supervisionSubsystemsWithNoTasksFilter = (
+    currentAgencyMetrics,
+    system
+  ) => {
+    const isSupervisionSubsystem = Boolean(
+      SupervisionSubsystems.includes(system.toUpperCase())
+    );
+    if (!isSupervisionSubsystem) return true;
+
+    const supervisionSubsystemsWithTaskCards = currentAgencyMetrics
+      .filter(metricEnabled)
+      .filter((metric) => {
+        const metricSystem = metric.system.key.toLocaleLowerCase();
+        return metricSystem === system.toLowerCase();
+      });
+
+    if (supervisionSubsystemsWithTaskCards.length > 0) {
+      return true;
+    }
+    return false;
+  };
+
   return (
     <Styled.HomeContainer>
       <Styled.WelcomeUser>Welcome, {userFirstName}</Styled.WelcomeUser>
@@ -189,15 +216,17 @@ export const Home = observer(() => {
         <Styled.SystemSelectorContainer>
           <div />
           <Styled.SystemSelectorTabWrapper>
-            {agencySystems?.map((system) => (
-              <Styled.SystemSelectorTab
-                key={system}
-                selected={system === currentSystem}
-                onClick={() => setCurrentSystem(system)}
-              >
-                {system.toLocaleLowerCase()}
-              </Styled.SystemSelectorTab>
-            ))}
+            {agencySystems
+              ?.filter(supervisionSubsystemsWithNoTasksFilter)
+              .map((system) => (
+                <Styled.SystemSelectorTab
+                  key={system}
+                  selected={system === currentSystem}
+                  onClick={() => setCurrentSystem(system)}
+                >
+                  {system.toLocaleLowerCase()}
+                </Styled.SystemSelectorTab>
+              ))}
           </Styled.SystemSelectorTabWrapper>
           <div />
         </Styled.SystemSelectorContainer>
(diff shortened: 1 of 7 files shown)


mxosman:
Hmm - the "Dashboards are updated with the latest published records" copy for these supervision subsystems (without any metrics) still looks funky to me.

Screenshot 2023-07-12 at 12 46 05 PM


If we want to keep the copy and not add to the `hasCompletedAllTasks` logic - we could hide the supervision subsystems tabs if there aren't any. Open to any thoughts!

jwhrec:
Thanks Mahmoud! This all LGTM. For your question around the supervision subsystems, I agree with your suggestion of hiding the tabs.

publisher/src/components/Home/helpers.ts
```diff
@@ -50,6 +50,21 @@ export const createReportTitle = (record: Report, monthName?: string) => {
...
+ * of their systems (under the "All" filter)
+ * @returns "Metric" or "Metric (System Name)"
+ */
+export const formatTaskCardTitle = (
+  title: string,
+  systemName: string,
+  hasMultipleSystemsAndAllSystemsFilter?: boolean
+) => {
+  if (!hasMultipleSystemsAndAllSystemsFilter) return title;
+  return `${title} ${title.includes(systemName) ? `` : `(${systemName})`}`;
```
lilidworkin, line 65: Just curious -- is it ever the case that title will already include system name? Certainly can't hurt to include this, just wondering!

mxosman: Yes - I noticed the supervision subsystems already come with their subsystem name - caught it when I saw them rendering as (e.g. "Funding (Parole) (Parole)")

alfonsogoberjr (approved):
Nothing new for me to comment here, LGTM

publisher/src/components/Home/Home.tsx
```diff
@@ -236,6 +249,15 @@ export const Home = observer(() => {
...
                   (record) => {
+                    /** Only display records w/ metrics that match the current system */
+                    if (
+                      Object.values(record.metrics).filter(
+                        ([metric]) =>
+                          metricBelongsToCurrentSystem(metric) &&
+                          metricEnabled(metric)
+                      )?.length === 0
+                    )
+                      return null;
```
mxosman, line 260: This change was due to another edge-case I caught with supervision subsystems specifically w/ annual records.

Here's the scenario I ran into:
* Systems: Prison, Supervision, Parole, Probation
* There is an existing CY annual record for Prison & Supervision metrics
* I disaggregated the Supervision: Funding metric (now it appears available in Parole and Probation)
* I marked Probation: Funding as unavailable - so now we're just left with Parole: Funding
* I change the reporting frequency for Parole: Funding to FY - and created an annual FY record

Without the above change, all of the annual record Publish cards appear when you filter the homepage to only show Parole task cards. This is definitely an edge-case because typically, even disaggregated metrics still fall under the same frequency, and I can't imagine a disaggregated supervision metric having multiple annual records at different frequencies. 

Either way, this basically goes through each annual record's metrics and only displays the annual records that have enabled metrics that match the current system. LMK if this isn't entirely clear - finding it a bit tough to explain.

publisher/src/components/Home/helpers.ts
```diff
@@ -36,6 +42,41 @@ export const metricIsMonthly = (metric: Metric) =>
...
+    .filter((metric) => {
+      const metricSystem = metric.system.key.toLocaleLowerCase();
+      return metricSystem === system.toLocaleLowerCase();
+    });
+
+  if (enabledSupervisionSubsystemMetrics.length > 0) {
+    return true;
+  }
+  return false;
+};
```
mxosman, line 78: So, this filter is a new addition that's used here when rendering the filter options:

Screenshot 2023-07-14 at 10 12 08 AM

```
            {agencySystems
              ?.filter((system) =>
                supervisionSubsystemsWithEnabledMetrics(
                  system,
                  currentAgencyMetrics
                )
              )
              .map((system) => (
              ...
```

Hiding the supervision subsystems with no metrics/tasks was super tricky. After playing around with this, I realized that we can just look and see if the supervision subsystem has enabled metrics. If so, we can assume 2 things:
1. The metric has been configured (clicking on Disaggregated in the parent metric sets that parent metric to 'Not Available' and sets the supervision subsystem metric(s) as enabled)
2. The supervision subsystem will have relevant task cards generated for the enabled metrics (including a true "All tasks completed" if those subsystem metrics are published in a record)

Let me know if this feels too confusing. I tested this and it's consistently giving me the expected behavior.

mxosman:
Sorry, Lili and Alfonso - I made a couple more changes and added some clarification comments. After that - this should be all set. Many many thanks for your time!

publisher/src/components/Home/Home.tsx
lilidworkin: @mxosman I didn't think this through too carefully but I'm sure this is fine :D Your fix seems safe enough!

publisher/src/components/Home/helpers.ts
lilidworkin: @mxosman this logic sounds fine, I'm not sure what you meant by "Hiding the supervision subsystems with no metrics/tasks" though -- I thought if a subsystem has no tasks, it shows "All tasks completed". And I'm not sure what it would mean to have no metrics -- all metrics are disabled? But then wouldn't we just show "All tasks completed" since you've configured everything and there are no tasks? Maybe I'm missing something.

lilidworkin (approved):
One question inline but this looks good to go!

publisher/src/components/Home/helpers.ts
lilidworkin: Are you talking about hiding the system from the tab bar? Why would we want to do that?

lilidworkin: Ohh I missed your comment above! Okay yea actually this makes a lot of sense. I like this solution! Thank you for catching this and coming up with a product fix!

mxosman: Sounds great, Lili! I also made an enhancement task here if we ever want to revisit the copy in those rare cases where a user turns all metrics off: #775

PR #758, opened 2023-07-10
[Justice Counts] Connect FE to BE Standalone Bulk Upload Review Page

morden35:
## Description of the change

This PR connects the FE code for the Bulk Upload Standalone Review Page to the BE api endpoint that fetches the stored upload review data (stored in gcp bucket).

Accompanying BE PR: https://github.com/Recidiviz/recidiviz-data/pull/21959

Frontend and Backend working together:

https://github.com/Recidiviz/justice-counts/assets/82831800/9b461163-1c2d-414d-a642-ead50a91ec24

## Type of change

> All pull requests must have at least one of the following labels applied (otherwise the PR will fail):

| Label                       	| Description                                                                                               	|
|-----------------------------	|-----------------------------------------------------------------------------------------------------------	|
| Type: Bug                   	| non-breaking change that fixes an issue                                                                   	|
| Type: Feature               	| non-breaking change that adds functionality                                                               	|
| Type: Breaking Change       	| fix or feature that would cause existing functionality to not work as expected                            	|
| Type: Non-breaking refactor 	| change addresses some tech debt item or prepares for a later change, but does not change functionality    	|
| Type: Configuration Change  	| adjusts configuration to achieve some end related to functionality, development, performance, or security 	|
| Type: Dependency Upgrade      | upgrades a project dependency - these changes are not included in release notes                             |

## Related issues

Closes #XXXX

### Development

**This box MUST be checked by the submitter prior to merging**:
- [x] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [ ] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #767, opened 2023-07-13
[Justice Counts] Connect FE to BE Standalone Bulk Upload Errors/Warnings Page

morden35:
## Description of the change

This PR connects the FE code for the Bulk Upload Standalone Errors/Warnings Page to the BE api endpoint that fetches the stored upload errors/warnings data (stored in gcp bucket).

Accompanying BE PR: https://github.com/Recidiviz/recidiviz-data/pull/21959

Frontend and Backend working together:

https://github.com/Recidiviz/justice-counts/assets/82831800/44818c41-6f8d-43dd-922a-41d20c251189

## Type of change

## Related issues

Related to #758

### Development

### Code review

### DIFF
--- a/publisher/src/App.tsx
+++ b/publisher/src/App.tsx
@@ -20,6 +20,8 @@ import React, { ReactElement, useEffect } from "react";
 import { Navigate, Route, Routes, useLocation } from "react-router-dom";
 
 import { trackNavigation } from "./analytics";
+import ShareSpreadsheetReview from "./components/DataUpload/ShareSpreadsheet";
+import ShareUploadErrorWarnings from "./components/DataUpload/ShareUploadErrorWarnings";
 import Footer from "./components/Footer";
 import { AppWrapper, PageWrapper } from "./components/Forms";
 import { REPORTS_LOWERCASE } from "./components/Global/constants";
@@ -80,6 +82,14 @@ const App: React.FC = (): ReactElement => {
                   }`}
                 />
               }
+            />
+            <Route
+              path="/upload/:spreadsheetId/errors-warnings"
+              element={<ShareUploadErrorWarnings />}
+            />
+            <Route
+              path="/upload/:spreadsheetId/review-metrics"
+              element={<ShareSpreadsheetReview />}
             />
           </Routes>
         ) : (
--- a/publisher/src/App.tsx
+++ b/publisher/src/App.tsx
@@ -20,8 +20,6 @@ import React, { ReactElement, useEffect } from "react";
 import { Navigate, Route, Routes, useLocation } from "react-router-dom";
 
 import { trackNavigation } from "./analytics";
-import ShareSpreadsheetReview from "./components/DataUpload/ShareSpreadsheet";
-import ShareUploadErrorWarnings from "./components/DataUpload/ShareUploadErrorWarnings";
 import Footer from "./components/Footer";
 import { AppWrapper, PageWrapper } from "./components/Forms";
 import { REPORTS_LOWERCASE } from "./components/Global/constants";
@@ -82,14 +80,6 @@ const App: React.FC = (): ReactElement => {
                   }`}
                 />
               }
-            />
-            <Route
-              path="/upload/:spreadsheetId/errors-warnings"
-              element={<ShareUploadErrorWarnings />}
-            />
-            <Route
-              path="/upload/:spreadsheetId/review-metrics"
-              element={<ShareSpreadsheetReview />}
             />
           </Routes>
         ) : (
--- a/publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
+++ b/publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
@@ -15,6 +15,10 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import { Button } from "@justice-counts/common/components/Button";
+import { MIN_DESKTOP_WIDTH } from "@justice-counts/common/components/GlobalStyles";
+import { HeaderBar } from "@justice-counts/common/components/HeaderBar";
+import { useWindowWidth } from "@justice-counts/common/hooks";
 import {
   ReportOverview,
   SupervisionSubsystems,
@@ -25,7 +29,10 @@ import { useNavigate, useParams } from "react-router-dom";
 
 import { useStore } from "../../stores";
 import { PageWrapper } from "../Forms";
+import { REPORTS_LOWERCASE } from "../Global/constants";
+import { useHeaderBadge } from "../Header/hooks";
 import { Loading } from "../Loading";
+import { DataUploadContainer } from ".";
 import ShareSpreadsheet from "./ShareSpreadsheet";
 import { ErrorsWarningsMetrics } from "./types";
 import { UploadErrorsWarnings } from "./UploadErrorsWarnings";
@@ -34,6 +41,8 @@ import { processUploadResponseBody } from "./utils";
 function ShareUploadErrorWarnings() {
   const navigate = useNavigate();
   const { userStore, reportStore } = useStore();
+  const headerBadge = useHeaderBadge();
+  const windowWidth = useWindowWidth();
 
   const [errorsWarningsMetrics, setErrorsWarningsMetrics] =
     useState<ErrorsWarningsMetrics>();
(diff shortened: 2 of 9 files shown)


morden35:
1 thing that I think we still need to do is gate the standalone review page and standalone errors/warnings page to users that belong to the associated agency.

For example, I can currently visit: http://localhost:3000/agency/148/upload/128/errors-warnings

Even though the spreadsheet with id=128 belongs to agency 151 (Jails)

publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
```diff
@@ -85,48 +87,32 @@ function ShareUploadErrorWarnings() {
...
-  //       setLoadingError(result.message);
-  //     }
-  //
-  //     setNewAndUpdatedReports({
-  //       newReports: result.new_reports || [],
-  //       updatedReports: result.updated_reports || [],
-  //       unchangedReports: result.unchanged_reports || [],
-  //     });
-  //
-  //     const errorsWarningsAndMetrics = processUploadResponseBody(
```
lilidworkin, line 101: Do we need lines 101-109?

lilidworkin (commented):
Nice!!

What happens if you click "Continue" on this page? If it doesn't work, not blocking, we can address in a separate PR. I haven't yet given any thought to how these two standalone pages link together. It might actually just work out of the box but I genuinely don't know, haha.

As for your point about authentication -- really great call, thanks so much for thinking of that! I left a comment on your BE PR about how to fix :)

lilidworkin:
> For example, I can currently visit: http://localhost:3000/agency/148/upload/128/errors-warnings

Oh hm maybe this isn't the right URL -- maybe we should change the URL to just be http://localhost:3000/128/errors-warnings, and then the BE should check if the user has access to whatever agency spreadsheet 128 belongs to.

morden35:
> What happens if you click "Continue" on this page? If it doesn't work, not blocking, we can address in a separate PR. I haven't yet given any thought to how these two standalone pages link together. It might actually just work out of the box but I genuinely don't know, haha.

Clicking 'Continue' works and leads to the standard (not standalone) review page (see video)

publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
morden35: I dont think so because `errorsWarningsAndMetrics` and `hasErrorsOrWarnings` are already initialized [above](https://github.com/Recidiviz/justice-counts/pull/767/files#diff-007f291ff8b62583d3ffac669b13f55aa82368c45ab3188f4233d031c1b5694dL67) on lines 69 and 70.

publisher/src/App.tsx
```diff
@@ -81,6 +83,14 @@ const App: React.FC = (): ReactElement => {
                 />
               }
             />
+            <Route
```
morden35, line 86: Moving the routes here changes the urls to:
http://localhost:3000/upload/128/errors-warnings
http://localhost:3000/upload/128/review-metrics

However now the header does not show on the errors-warnings page, the 'Continue' button on the errors-warnings page breaks, and the 'Publish' and 'Exit without Publishing' buttons break

morden35: Maybe @mxosman could help me with this?

The goal here is to allow users to access errors-warnings and review-metrics pages for any spreadsheet associated with their agency (and not access pages not associated with their agency).

I'm not sure we care so much about the url in this case, but rather we care that users only have access to the pages associated with their agency's spreadsheets, right?

I'm wondering if the code we added to the BE is sufficient to accomplish this?

```
      user = UserAccountInterface.get_user_by_auth0_user_id(
          session=current_session,
          auth0_user_id=get_auth0_user_id(request_dict=request_json),
      )

      raise_if_user_is_not_in_agency(user=user, agency_id=int(agency_id))
```

morden35: Because there, the agency_id is linked to the spreadsheet, so if the user requests the page for a spreadsheet that does not belong to their agency, raise_if_user_is_not_in_agency() will fail

mxosman: That makes sense to me, Michelle! Then the FE would receive the error from that request and the error message will display in L115 `return <PageWrapper>Error: {loadingError}</PageWrapper>`. 

Just a few notes after playing around with this:
* I do see the header missing - that makes sense actually, since the `UploadErrorsWarnings` component renders within the `DataUpload` and `DataUpload` has the header bar we need. We could for now copy that HeaderBar and its dependencies over (the one from the `DataUpload`). It will create two places to make updates - but we can address that in a follow up PR if we want to keep the momentum going with what you have here!
* I see the same error/warnings page as your demo regardless of which files I upload (11 metrics saved & 1 error)
* I am able to click Continue with no problem and go to the standard Publish review page and click the Publish button and get a success modal
* The records don't appear to be published after clicking publish - I noticed the report IDs that were sent to the backend to update the status didn't match the reports within the agency - I wonder if this is related somehow to seeing the same error/warnings page?

publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
lilidworkin: I think that is using the wrong data though -- we do `import { data } from "../../mocks/spreadsheetReviewData"` at the top of the file, and then on line 67 we use that mock data ` const errorsWarningsAndMetrics = processUploadResponseBody(data);`

publisher/src/App.tsx
lilidworkin: > I see the same error/warnings page as your demo regardless of which files I upload (11 metrics saved & 1 error)

This could be because of my comment above, re: using the wrong `data` variable

publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
morden35: oh wow thank you so much for catching this

mxosman: Oh nice!!! Completely missed that!

publisher/src/App.tsx
morden35: I believe everything in this thread has been addressed! Please do test that this is true!

publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
```diff
@@ -64,83 +74,81 @@ function ShareUploadErrorWarnings() {
...
-      <UploadErrorsWarnings
-        errorsWarningsMetrics={errorsWarningsMetrics}
-        newAndUpdatedReports={newAndUpdatedReports}
-        selectedSystem={userSystems.length === 1 ? userSystems[0] : undefined}
-        resetToNewUpload={() => navigate(`/agency/${agencyId}/upload`)}
-        // fileName={reportStore.spreadsheetReviewData[spreadsheetId].fileName}
-        fileName="new file"
-      />
+      <DataUploadContainer>
+        <HeaderBar
```
lilidworkin, line 125: Just a quick question about this -- I thought you said that the header only broke if we changed the URLs, which we're not doing anymore, right? Or is this a different header? Would you mind taking a screenshot of what it looks like with/without this header just so I can see quickly?

lilidworkin (commented):
@morden35 this looks great! I just have a couple questions about the header situation. I'll leave the final review to @mxosman !

publisher/src/components/DataUpload/ShareUploadErrorWarnings.tsx
morden35: > I thought you said that the header only broke if we changed the URLs, which we're not doing anymore, right?

Not sure what this is referring to!

With header (same as standard errors-warnings page):
Screenshot 2023-07-18 at 10 58 34 AM

Without header:
Screenshot 2023-07-18 at 10 59 36 AM

lilidworkin: Got it! Ok cool cool!

mxosman (approved):
🚀

morden35:
> Hey kudos to you, Michelle! This looks fantastic! Things feel like they work as expected - was able to publish all the records that appeared on the uploads. So cool to see these standalone pages working!
> 
> [I also filed a follow up task for any one of us later to find a way to deduplicate the HeaderBar: #774 ]

Hooray! Thank you so much!

PR #780, opened 2023-07-19
Update Delete button behavior in Records Overview page

### LINKED ISSUE #779: Delete button in Publisher Records Page
Within the records page, when a user presses the delete button it should not delete any files but instead a popup should appear with the following message: "Please reach out to justice-counts-support@csg.org if you would like to delete a file."

This is the same message as appears for the delete button in the uploaded files page.

mxosman:
## Related issues

Closes #779 

### Development

### Code review

### DIFF
--- a/publisher/src/pages/Reports.tsx
+++ b/publisher/src/pages/Reports.tsx
@@ -118,6 +118,10 @@ const Reports: React.FC = () => {
   );
   const [isRemoveRecordsModalOpen, setIsRemoveRecordsModalOpen] =
     useState(false);
+  const [
+    isUnauthorizedRemoveRecordsModalOpen,
+    setIsUnauthorizedRemoveRecordsModalOpen,
+  ] = useState(false);
 
   const isAdmin =
     userStore.isJusticeCountsAdmin(agencyId) ||
@@ -170,6 +174,7 @@ const Reports: React.FC = () => {
   };
 
   const handleRemoveRecords = () => {
+    if (!isJCAdmin) return;
     reportStore.deleteReports(selectedRecords, agencyId);
     clearBulkAction();
     clearAllSelectedRecords();
@@ -236,17 +241,14 @@ const Reports: React.FC = () => {
     {
       key: "deleteAction",
       label: "Delete",
-      onClick: () => selectBulkAction("delete"),
+      onClick: () =>
+        isJCAdmin
+          ? selectBulkAction("delete")
+          : setIsUnauthorizedRemoveRecordsModalOpen(true),
       color: "red",
       noHover: true,
     },
-  ].filter((option) => {
-    // Gates `Delete` action to Justice Counts Admins only
-    if (option.key === "deleteAction" && !isJCAdmin) {
-      return false;
-    }
-    return true;
-  }) as DropdownOption[];
+  ] as DropdownOption[];
   const tabbedBarOptions: TabOption[] = Object.entries(
     ReportStatusFilterOptionObject
   ).map(([key, value]) => ({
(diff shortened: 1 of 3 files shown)


publisher/src/pages/Reports.tsx
```diff
@@ -170,6 +175,7 @@ const Reports: React.FC = () => {
   };
 
   const handleRemoveRecords = () => {
+    if (!isJCAdmin) return;
```
mxosman, line 178: Just for extra safety.

publisher/src/components/Modals/UnauthorizedDeleteActionModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+}> = ({ closeModal }) => {
+  return (
+    <Modal
+      title={
+        <>
+          Please reach out to{" "}
+          <a href="mailto:justice-counts-support@csg.org">
+            justice-counts-support@csg.org
+          </a>{" "}
+          if you would like to delete a file.
```
lilidworkin, line 32: Should this say "record" or "file" depending on the entry point?

jwhrec:
good catch! Yes that'd be great.

publisher/src/components/Modals/UnauthorizedDeleteActionModal.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </>
+      }
+      description=""
+      primaryButton={{
```
alfonsogoberjr, line 36: Why do we pass props directly through to Modal? Can we use `children` instead?

alfonsogoberjr (approved):

publisher/src/components/Modals/UnauthorizedDeleteActionModal.tsx
mxosman: Good question - all the modals we've used so far have a similar design structure. If we made it more generic, we'd be duplicating a lot of the same structure to pass through as `children`. Totally see an argument for having a generic version of the modal that will wrap around any set of child components! We just haven't deviated much from how the modals look.

publisher/src/components/Modals/UnauthorizedDeleteActionModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { removeAgencyFromPath } from "../../utils";
+import { REPORT_LOWERCASE, REPORTS_LOWERCASE } from "../Global/constants";
+
+export const UnauthorizedDeleteActionModal: React.FC<{
+  closeModal: () => void;
+}> = ({ closeModal }) => {
+  const location = useLocation();
+  const isRecordsPage =
+    removeAgencyFromPath(location.pathname) === REPORTS_LOWERCASE;
+  const typeOfAssetDisplayName = isRecordsPage ? REPORT_LOWERCASE : "file";
```
mxosman, line 31: I feel like it's safe to say the Records overview page is the only place where you could delete a record - so, if this modal is not in the records page, we can assume it's a non-record asset (e.g. "file"). LMK if this doesn't feel like a safe assumption and I can make this specifically check whether we're in the Uploaded Files page. I figured for now, this is safe to assume. (Thanks for catching this, Lili)

lilidworkin: I think this is fine! The way I would have done it probably would have been to pass an enum parameter to this modal with the name of the object to delete, e.g. either FILE or RECORD. Then we could expand that enum in the future if we ever wanted to use this modal in other places for other assets. But that may be a premature optimization!

mxosman: That's a great call, Lili! I like that idea of having an enum that we could expand in the future. Will update!

mxosman: Was this in-line with what you were thinking? https://github.com/Recidiviz/justice-counts/pull/780/commits/65ebdf4bbad3e476b342ca376deeb9abee3f5a2c

Not entirely feeling the name `typeOfAsset` either - will think of something better.

lilidworkin: Yea that looks perfect!

PR #782, opened 2023-07-20
[#777] force reload on change dropdown selector event

### LINKED ISSUE #777: [Justice Counts] Uploaded Files bug
While debugging a separate issue, I noticed that when you are in the 'Uploaded Files' page within 'Settings, if you change agencies, the url changes but the page does not update:

https://images.zenhubusercontent.com/529290376/f2a6fecb-f256-4eee-9c29-07954e8ac761/loom_message___19_july_2023.mp4

Steps to reproduce:
1. Log in to Publisher
2. Select the following Agency: Columbus Police Department
3. Navigate to Setting > Uploaded Files (at the time of writing this ticket, CPD had attempted 1 file upload)
4. Select the following Agency: Iowa Department of Corrections (IDOC)
At this point, the URL and 'Welcome, <Name> at Iowa Department of Corrections (IDOC)' banner should have changed. However the contents of the Uploaded Files page has not.
5. Navigate to 'Team Management' and then back to 'Uploaded Files'
At this point, the page should have refreshed and populated the 'Uploaded Files' with the proper files associated with IDOC.

alfonsogoberjr:
## Description of the change

Force page reload when selecting Agency dropdown items

## Type of change

Bug

## Related issues

Closes #777

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataUpload/UploadedFiles.tsx
+++ b/publisher/src/components/DataUpload/UploadedFiles.tsx
@@ -30,13 +30,13 @@ import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 
+import { UploadedFile, UploadedFileStatus } from ".";
 import { useStore } from "../../stores";
 import { formatSystemName } from "../../utils";
-import downloadIcon from "../assets/download-icon.png";
 import { SYSTEM_CAPITALIZED } from "../Global/constants";
 import { ContainedLoader } from "../Loading";
+import downloadIcon from "../assets/download-icon.png";
 import { TeamMemberNameWithBadge } from "../primitives";
-import { UploadedFile, UploadedFileStatus } from ".";
 import {
   ActionsContainer,
   DateUploaded,
@@ -315,7 +315,7 @@ export const UploadedFiles: React.FC = observer(() => {
       fetchListOfUploadedFiles();
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    []
+    [agencyId]
   );
 
   if (isLoading) {
--- a/publisher/src/components/DataUpload/UploadedFiles.tsx
+++ b/publisher/src/components/DataUpload/UploadedFiles.tsx
@@ -30,13 +30,13 @@ import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 
-import { UploadedFile, UploadedFileStatus } from ".";
 import { useStore } from "../../stores";
 import { formatSystemName } from "../../utils";
+import downloadIcon from "../assets/download-icon.png";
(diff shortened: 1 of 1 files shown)


lilidworkin:
Thanks for the quick fix @alfonsogoberjr ! @mxosman curious what you think of this approach? I imagine there was a reason we weren't doing this before? How are we currently updating the state on agency change, and is there a way to apply that existing logic to the Uploaded Files page?

mxosman:
> Thanks for the quick fix @alfonsogoberjr ! @mxosman curious what you think of this approach? I imagine there was a reason we weren't doing this before? How are we currently updating the state on agency change, and is there a way to apply that existing logic to the Uploaded Files page?

Ooo - this is interesting. I probably would've just added the `agencyId` to the `UploadedFiles` dependency array so it re-renders:

https://github.com/Recidiviz/justice-counts/blob/7579815815a1e093ffc0a5c477388fb7e3d619b5/publisher/src/components/DataUpload/UploadedFiles.tsx#L313-L319)

But, this solution doesn't have that granular concern - it just updates the path (which has the new `agencyId`) and reloads the page. I'm actually not sure which would be best to do! On a component level, I think it makes sense to add the dependency because the useEffect should depend on the `agencyId` and refetch every time that changes - but I like how straightforward and agnostic this solution is.

yarn.lock
```diff
@@ -4762,6 +4758,11 @@
     "@types/qs" "*"
     "@types/serve-static" "*"
 
+"@types/file-saver@^2.0.5":
```
mxosman, line 4761: I think this yarn.lock might be an accidental commit?

lilidworkin:
>  I probably would've just added the agencyId to the UploadedFiles dependency array so it re-renders:

Yea, I think this is probably the way to go, since that's how we handle agency switching everywhere else? What do you think @alfonsogoberjr ? Curious to hear if you considered that and what you think the pros/cons of each approach is.

alfonsogoberjr:
> > Thanks for the quick fix @alfonsogoberjr ! @mxosman curious what you think of this approach? I imagine there was a reason we weren't doing this before? How are we currently updating the state on agency change, and is there a way to apply that existing logic to the Uploaded Files page?
> 
> Ooo - this is interesting. I probably would've just added the `agencyId` to the `UploadedFiles` dependency array so it re-renders:
> 
> https://github.com/Recidiviz/justice-counts/blob/7579815815a1e093ffc0a5c477388fb7e3d619b5/publisher/src/components/DataUpload/UploadedFiles.tsx#L313-L319
> 
> )
> But, this solution doesn't have that granular concern - it just updates the path (which has the new `agencyId`) and reloads the page. I'm actually not sure which would be best to do! On a component level, I think it makes sense to add the dependency because the useEffect should depend on the `agencyId` and refetch every time that changes - but I like how straightforward and agnostic this solution is.

Yeah way better

mxosman (approved):
LGTM!

PR #795, opened 2023-07-28
[#771][Agency Dashboard] Only allowlisted agencies should have a dashboard (all others 404)

alfonsogoberjr:
## Description of the change

Adds an allowlist for the DashboardView

## Type of change

Type: Feature

## Related issues

Closes #771

### Development

**This box MUST be checked by the submitter prior to merging**:
- [x] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [x] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #809, opened 2023-08-02
[#771][Agency Dashboard] Only allowlisted agencies should have a dashboard (all others 404)

### LINKED ISSUE #771: [Agency Dashboard] Only allowlisted agencies should have a dashboard (all others 404)
Allowlist is here: https://docs.google.com/document/d/1ga2GWL8MnEJLkzzjt5DlBJdJSyKf_-fgVzXn2-IBPCw/edit

This sort of blocks on https://app.zenhub.com/workspaces/justice-counts--622109ea3bb6e0001d2d06ad/issues/gh/recidiviz/recidiviz-data/22171, but we can still get started on this by creating a dummy allowlist.

alfonsogoberjr:
## Description of the change

Adds corrections and feedback from previous PR 

Closes #771

## Test Plan

Test at https://alfonsotest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/

I added staging to the allowlist logic as well so this is testable before merging. 

- Navigate to https://alfonsotest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/agency/department-of-corrections-%5Bcla-demo%5D/dashboard
- Verify that it loads.
- Navigate to a different agency not on the allowlist. 
- Verify that it 404s

### Development

### Code review


agency-dashboard/src/DashboardView/DashboardView.tsx
```diff
@@ -167,6 +167,12 @@ export const DashboardView = observer(() => {
   }, []);
 
   useEffect(() => {
+    if (
```
lilidworkin, line 170: Why is this block duplicated?

agency-dashboard/src/utils/env.ts
```diff
@@ -16,8 +16,8 @@
 // =============================================================================
 
 enum envRegexes {
-  PRODUCTION = "/publisher.justice-counts.org/i",
-  STAGING = "/publisher-staging.justice-counts.org/i",
+  PRODUCTION = "/dashboard.justice-counts.org/i",
+  STAGING = "/dashboard-staging.justice-counts.org/i",
```
lilidworkin, line 20: I just realized this won't work with playtesting URLs, which are different ... I guess that's a reason to favor the API approach. But I don't want to spend too long on this. I think it's okay if playtesting URLs are treated as development for this purpose, but let's remember this in case it causes issues in the future.

lilidworkin: @alfonsogoberjr just want to confirm that you did see Michelle and Mahmoud's pointers about how to do this via the API [here](https://github.com/Recidiviz/justice-counts/pull/795#discussion_r1279926930), in case we have to change it later?

lilidworkin (changes requested):
Can we delete this rogue file as per my comment [here](https://github.com/Recidiviz/justice-counts/pull/795/files#r1281913615)?

agency-dashboard/src/utils/env.ts
alfonsogoberjr: Confirmed

agency-dashboard/src/DashboardView/DashboardView.tsx
alfonsogoberjr: Rebase gone wrong

agency-dashboard/src/DashboardView/DashboardView.tsx
```diff
@@ -157,6 +158,7 @@ export const DashboardView = observer(() => {
   useAsyncEffect(async () => {
     try {
       await agencyDataStore.fetchAgencyData(slug as string);
+      setEnv(await getEnv());
```
lilidworkin, line 161: I'm curious about the pros/cons of this approach, where we make the backend API call here, as opposed to the approach in Publisher, where it's called in the API store [here](https://github.com/Recidiviz/justice-counts/blob/cb38e2e71d87388dfa9d4e3ce1bf30d32ecd024d/publisher/src/stores/API.ts#L50), and the result is stored in an `environment` property on that API to be referenced later. If we do it the way it is here, will we be calling the API more times than we need to? Like if we switch back and forth between different agencies, will we call the API each time?

alfonsogoberjr: I tend to agree, because the API store lives in a React context, where it is mutable and only rendered once. 

There are more problems with this approach than just this, however. The endpoint is behind auth, so [this ticket](https://app.zenhub.com/workspaces/justice-counts--622109ea3bb6e0001d2d06ad/issues/gh/recidiviz/recidiviz-data/22928) blocks this now.

agency-dashboard/src/utils/env.ts
mxosman: Good call about the playtesting URLs - didn't think of that. I think we'll also need to add a couple lines to the `/api/init` endpoint to add in an explicit check and return of the "production" environment (atm, we just have "unknown", "staging" and "local")

agency-dashboard/src/DashboardView/DashboardView.tsx
mxosman: Yeah, it's returning 401 unauthorized when called from agency dashboards. I was hoping it would work out of the box, but I remembered we never set up any auth handling in agency dashboards. Since that endpoint's auth decorator requires an auth token, would it make sense for us to add a new endpoint without the auth decorator just to get the environment?

alfonsogoberjr: I can just mirror the Publisher API store pattern for the entire agency dashboard app, but this will take some time. Since the api endpoint needs to be changed, that pushes back the timeline for this specific ticket. Other agency dashboard work (such as [here](https://github.com/Recidiviz/justice-counts/pull/810/files)) can continue.

mxosman: Since this app is mostly public - I'm not sure if it's worth the lift just to get the environment. I can work on the endpoint this morning to unblock.

lilidworkin: Just to make sure I understand -- doesn't this current approach (outside of the Auth store, but still calling the API endpoint) run into the same trouble?

lilidworkin: Yes, 100% agreed @mxosman ! If you could pair with @alfonsogoberjr on this that would be awesome, it shouldn't be hard at all.

lilidworkin: Actually @mxosman I think @alfonsogoberjr should do the endpoint! But maybe you can just help him out?

alfonsogoberjr: It's done, try https://alfonsotest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/api/env

lilidworkin (commented):
Looks right to me! Please add a test plan.

agency-dashboard/src/stores/API.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      return response;
+    } catch (error) {
+      if (error instanceof Error) {
+        // TODO add some tracking here
+        throw error;
+      }
+      throw error;
+    }
+  }
```
mxosman, line 78: Nice! Not for this PR, but might be worth making a note for us to either use this `request` method throughout and deprecate the `/networking.ts` helper, or if we're good with that helper, we could probably get rid of this method in the store and import the `request` from `/networking.ts`?

alfonsogoberjr: Oh I forgot about this duplication. Will consolidate

lilidworkin (approved):
@mxosman should probably also take a look, but this lgtm. Thanks for the detailed test plan!

mxosman (approved):
LGTM! Thanks, Alfonso!

PR #810, opened 2023-08-03
[#792] [Agency Dashboard] Part 2: Capacity and Costs Category Page (v2) 

### LINKED ISSUE #792: [Agency Dashboard] Part 2: Capacity and Costs Category Page (v2)
Figma [here](https://www.figma.com/file/6VEZgDlhYCTR0Jnpyh63R9/CLA-Demo?type=design&node-id=228%3A21&mode=design&t=5eCxdkM0Dhwj4EvP-1) -- the "*with* breakdowns" version.

Also covers adding the LineChart component from #22805 to this page.

Target Date: 08/02/23

alfonsogoberjr:
## Description of the change

[x] first pass: customize legend to fit mockups
[x] second pass: reorganize logic for testability and proper dependency handling
[x] third pass: implement last-minute Figma changes
Screenshot 2023-08-08 at 1 13 42 PM

## Related issues

Closes #792 

## Test Plan

Playtesting link: https://alfonsotest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/agency/department-of-corrections-[cla-demo]/capacity-and-costs

### Development

### Code review


common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -31,39 +33,64 @@ export function CategoryOverviewLineChart({
...
 }: LineChartProps) {
-  const renderLines = () => {
-    // each Recharts Bar component defines a category type in the stacked bar chart
-    let lineDefinitions: JSX.Element[] = [];
-    dimensions.forEach((dimension, index) => {
-      const newLine = (
-        <Line
-          key={dimension}
-          dataKey={dimension}
-          stroke={Object.values(palette.dataViz)[index]}
```
alfonsogoberjr, line 42: This was being recomputed on every re-render, and since `Object.values` doesn't guarantee order, so it ended up being different each time. This change standardizes it into a single `colorDict` made from transforming the `palette.dataViz` properties based on their keys (which luckily happen to contain numbers already).

common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -31,39 +33,64 @@ export function CategoryOverviewLineChart({
...
-      margin={{ top: 5, right: 30, left: 20, bottom: 5 }}
-    >
-      <XAxis
-        dataKey={(datapoint) => {
-          const [, , , year] = splitUtcString(datapoint.start_date);
-          return year;
-        }}
-      />
-      <Tooltip />
-      <Legend />
```
alfonsogoberjr, line 65: The built-in legend provided by recharts was significantly different from our designs. Turns out you can customize them.

lilidworkin (commented):
Thanks @alfonsogoberjr ! Not sure if this is ready for review or not (looks like it's currently in draft?) but just a couple quick thoughts: 
* I like how you made that checklist in the description to clarify what parts this PR addresses, that was super helpful for me! 
* Since this PR doesn't fully address the "Part 2: Capacity and Costs Category Page (v2)"task, I would name it something more appropriate like "[Agency Dashboard] Styling the legend for line chart component". Also, instead of "Closes #task" you can say "Contributes to #task".

alfonsogoberjr:
> Thanks @alfonsogoberjr ! Not sure if this is ready for review or not (looks like it's currently in draft?) but just a couple quick thoughts:
> 
> * I like how you made that checklist in the description to clarify what parts this PR addresses, that was super helpful for me!
> * Since this PR doesn't fully address the "Part 2: Capacity and Costs Category Page (v2)"task, I would name it something more appropriate like "[Agency Dashboard] Styling the legend for line chart component". Also, instead of "Closes #task" you can say "Contributes to #task".

Correct it's not finished, but I was planning for it to contain the full scope. I've already added the hover responses, just need to refine the switching logic becuase this recharts library is sensitive to excessive re-renders.

common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -31,39 +33,61 @@ export function CategoryOverviewLineChart({
...
 }: LineChartProps) {
-  const renderLines = () => {
+  const colorDict = pipe(
+    invertObj,
+    mapObjIndexed(
+      (colorName: string, fill) =>
+        dimensions[Number(colorName.replace("bar", "")) - 1]
+    ),
+    invertObj
+  )(palette.dataViz);
```
mxosman, line 43: Trying to understand this... 
1. We're taking in the `palette.dataViz` object and first inverting it `e.g. { [value]: [key] }` (e.g. e.g.`{ "#133158": "bar1" }`)
2. Then, we are mapping it and updating the value (which was previously the key?) to whatever index of the `dimensions` array (based on the # suffix in the `bar#` name) - which I think are just the dimension names?
3. Then we are inverting the object again so we're back to our usual key/value except now it's `{ [dimension name]: [color value] }`?

Is `fill` doing anything here?

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -112,13 +124,52 @@ export const CategoryOverview = observer(() => {
...
+
+  useEffect(() => {
+    console.log(selectedYear, selectedMetricKey);
+    if (selectedYear && selectedMetricKey)
+      setLineChartData(
+        getLineChartData(selectedMetricKey).filter(
+          (datapoint) => getDatapointYear(datapoint) === selectedYear
+        )
+      );
+  }, [selectedYear, selectedMetricKey]);
```
mxosman, line 167: Should we add `getLineChartData` to the dependency array? [Assuming it doesn't create issues!]

mxosman:
One last comment and nit regarding the styling and the loading of the datapoints. 
* For some reason, the datapoints don't load for multiple years? Sometimes they don't load at all, other times there's just one year loaded.
* Nit: there's about a ~58px gap in between the last bar chart and the legend in Figma - legend feels a bit too close atm

Screenshot 2023-08-03 at 1 18 50 PM

lilidworkin:
@alfonsogoberjr at the point that this is ready for review, can you put it on playtesting and share it with the team?

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -63,10 +66,19 @@ export const CategoryOverview = observer(() => {
...
   );
+  const [lineChartData, setLineChartData] = useState<Datapoint[]>([]);
 
-  const filterDatapoints = (datapoints: Datapoint[]) => {
-    return dataRangeFilter === "recent" ? datapoints.slice(-5) : datapoints;
-  };
+  const categoryMetrics =
+    agencyDataStore.metricsByCategory[categoryData[category].key];
+  const [selectedMetricKey, setSelectedMetricKey] = useState<string>();
+  const [selectedYear, setSelectedYear] = useState<string>();
```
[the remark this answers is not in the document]
alfonsogoberjr: good catch

agency-dashboard/src/utils/formatting.ts
```diff
@@ -13,6 +13,15 @@
...
+
+import { splitUtcString } from "@justice-counts/common/components/DataViz/utils";
+import { Datapoint } from "@justice-counts/common/types";
+
 // =============================================================================
 export const slugify = (str: string): string =>
   str?.replace(/\s/g, "-")?.toLowerCase();
+
+export const getDatapointYear = (datapoint: Datapoint) => {
+  const [, , , year] = splitUtcString(datapoint.start_date);
```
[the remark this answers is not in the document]
alfonsogoberjr: Yeah I really hate this method

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -183,42 +234,27 @@ export const CategoryOverview = observer(() => {
...
-                          agencyDataStore.datapointsByMetric[metric.key]
-                            .aggregate,
-                          DataVizTimeRangesMap.All,
-                          "Count"
-                        )
-                      )}
+                      width={1059}
+                      maxBarSize={193}
+                      onHoverBar={(data: ResponsiveBarData) => {
+                        console.log(data, getDatapointYear(data));
```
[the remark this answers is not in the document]
alfonsogoberjr: still WIP, will remove before converting to normal PR

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -112,13 +124,52 @@ export const CategoryOverview = observer(() => {
...
+          DataVizTimeRangesMap.All,
+          "Count"
+        )
+      );
+    },
+    [filterDatapoints, agencyDataStore.datapointsByMetric]
+  );
+
+  useEffect(() => {
+    console.log(selectedYear, selectedMetricKey);
```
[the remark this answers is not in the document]
alfonsogoberjr: same as above

common/components/DataViz/types.ts
```diff
@@ -38,4 +38,40 @@ export type ResponsiveBarChartProps = {
...
+  name: string;
+  payload: Datapoint;
+  type: any;
+  unit: any;
+  value: number;
+};
+
+export type ResponsiveBarData = Datapoint &
+  BarChartBackground &
+  Record<string, any> & {
```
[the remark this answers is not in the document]
alfonsogoberjr: Agreed, will refine that further

common/components/DataViz/CategoryOverviewLineChart.tsx
alfonsogoberjr: You got it exactly right. And no, fill can be removed.

alfonsogoberjr:
> One last comment and nit regarding the styling and the loading of the datapoints.
> 
> * For some reason, the datapoints don't load for multiple years? Sometimes they don't load at all, other times there's just one year loaded.
> * Nit: there's about a ~58px gap in between the last bar chart and the legend in Figma - legend feels a bit too close atm
> 
> Screenshot 2023-08-03 at 1 18 50 PM

Working on these issues!

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
alfonsogoberjr: Yes

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -183,42 +234,27 @@ export const CategoryOverview = observer(() => {
...
+                      data={getBarChartData(metric)}
                       dimensionNames={[DataVizAggregateName]}
                       ref={ref}
                     />
-                  </Styled.MetricDataVizContainer>
-                  <CategoryOverviewLineChart
-                    data={filterDatapoints(
-                      transformDataForBarChart(
+                    <CategoryOverviewLineChart
+                      data={lineChartData}
```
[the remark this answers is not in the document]
alfonsogoberjr: Thank you!

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -62,10 +67,66 @@ export const CategoryOverview = observer(() => {
...
-    return dataRangeFilter === "recent" ? datapoints.slice(-5) : datapoints;
-  };
+  const filterDatapoints = useCallback(
+    (datapoints: Datapoint[]) => {
+      return dataRangeFilter === "recent" ? datapoints.slice(-5) : datapoints;
+    },
+    [dataRangeFilter]
+  );
+
+  useEffect(() => {
```
alfonsogoberjr, line 92: This was carefully trial-and-errored, it turns out the agencyDataStore itself can't be used in the dependency array, it doesn't do anything. You have to add individual properties.

alfonsogoberjr:
cc @jwhrec

common/components/DataViz/CategoryOverviewBreakdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { getSumOfDimensionValues } from "./utils";
+
+export const CategoryOverviewBreakdown: FunctionComponent<{
+  data: Record<
+    keyof Datapoint,
+    { value: number | string | null; fill: string }
+  >;
+  dimensions: string[];
+}> = ({ data, dimensions }) => {
+  const renderText = (val: number | string | null, maxValue: number) => {
```
alfonsogoberjr, line 35: Same function from Tooltip. Could use some refinement.

common/hooks/useBarChart.ts
```diff
@@ -0,0 +1,61 @@
+// Recidiviz - a data platform for criminal justice reform
```
alfonsogoberjr, line 1: These three hooks were created to better encapsulate view logic and separate it from the components themselves. This is a pattern that allows for easier testability and better organization of functionality. The nomenclature is intended to reflect this - useBarChart is intended to be used with BarChart, useLineChart with LineChart, and so on.

common/components/DataViz/CategoryOverviewBreakdown.tsx
lilidworkin: Is it easy to pull this out into a shared utils file?

common/components/DataViz/CategoryOverviewLineChart.tsx
lilidworkin: Oh awesome, great fix!

common/hooks/useBarChart.ts
lilidworkin: Awesome!

common/components/DataViz/CategoryOverviewBreakdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+    if (typeof val !== "number") {
+      return "Not Reported";
+    }
+
+    let percentText = `${val !== 0 ? Math.round((val / maxValue) * 100) : 0}%`;
```
lilidworkin, line 40: Small potential bug: when I tested locally, I noticed all the percents are the same:
Screenshot 2023-08-08 at 10 46 19 AM

lilidworkin (commented):
Looks awesome! Left a few small comments, will leave the bulk of the review to @mxosman .

The playtesting link didn't work for me, I think it might be out of data, mind redeploying so @jwhrec can take a look? Thanks!

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
mxosman: >it turns out the agencyDataStore itself can't be used in the dependency array, it doesn't do anything

Yeah, the `agencyDataStore` will always reference the same object - so even if it's properties change, the reference to the object will not change, and just adding this by itself will never trigger a re-render. We mostly add it to the dependency array to satisfy the linter, but acknowledging that it will never be the thing to trigger a re-render.

RE: this `useEffect` block

This is not blocking for me as I'm sure this probably works fine as is, but I feel like this is the same trap we fell into last time - it might not be super reliable and will trigger multiple unnecessary re-renders.

Let's take the `agencyDataStore.loading` for example.

If we wrote this in-line on this component function's body (which is wrapped in the `observer` HoC):
```
export const CategoryOverview = observer(() => {
    ...
    const loading = agencyDataStore.loading;

    useAsyncEffect(() => {
       await agencyDataStore.someAsyncFetchThatUpdatesTheStoresLoadingValue();
       ...
     })
     
    return ...
});
```

It would behave exactly as writing this:
```
export const CategoryOverview = () => {
    ...
    const [loading, setIsLoading] = useState(true);

    useEffect(() => {
        ...
        await someAsyncFetch();
        setIsLoading(false);
    }, [])

    return ...
};
```

Both will re-render the `CategoryOverview` component when the `loading` value changes.

In this code - we're doing both and duplicating re-renders. When `agencyDataStore.fetchAgencyData(slug)` is called, it will update the store's `loading`, which will trigger a component re-render. But, the store's `loading` is also inside of a useEffect's dependency array and being set in local state, so once the mobX re-render happens because `loading` was updated in the store, this useEffect will trigger a duplicate re-render, because it detected that the `agencyDataStore.loading` was changed as well.

So we have two renders happening... for the one value that's changing (in this small example). My suggestion is that we lose this extra useEffect + setting mobX values (which are already being tracked) to local `useState`. We can trust mobX to handle the re-rendering and nail it every time.

[There's really not a lot of state changes happening here anyway. Once we call the endpoint via `agencyDataStore.fetchAgencyData(slug)` it loads all the values, and we consume them via the charts. Nothing else gets updated or happens unless you click into another agency.]

This is not a demo blocker for me, even if we have extra re-renders here, it'll still give us what we need, but just wanted to point it out. LMK if it's easier to sync on this if I'm not explaining it properly via writing.

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
```diff
@@ -174,51 +221,25 @@ export const CategoryOverview = observer(() => {
...
+                      width={620}
+                      data={getBarChartData(metric)}
                       dimensionNames={[DataVizAggregateName]}
                       metric={metric.display_name}
                       ref={ref}
                     />
+                    <CategoryOverviewLineChart
+                      data={getLineChartData(metric)}
+                      dimensions={getLineChartDimensions(metric)}
+                    />
```
mxosman, line 241: Not blocking - just a small thing I noticed - also not sure if this is a recharts bug:

Some times I get the little circles on each datapoint in the line chart and other times they go away

Screenshot 2023-08-08 at 11 15 13 AM

Screenshot 2023-08-08 at 11 14 57 AM

common/components/DataViz/CategoryOverviewBreakdown.tsx
mxosman: +1

tiny nit: we could give it a more descriptive name

common/components/DataViz/CategoryOverviewBreakdown.tsx
mxosman: +1

common/components/DataViz/CategoryOverviewBreakdown.tsx
alfonsogoberjr: Yeah for sure, I'll do that

common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -39,31 +66,62 @@ export function CategoryOverviewLineChart({
...
+                      ? Number(datapoint[dimension])
+                      : acc,
+                  0
+                )
+              )(data);
+              return reducedDatapoints > total ? reducedDatapoints : total;
+            },
+            0,
+            dimensions
+          )}
```
mxosman, line 109: nit: I think this could be extracted into a function that's passed here to avoid towering the prop

common/components/DataViz/CategoryOverviewBreakdown.tsx
alfonsogoberjr: I noticed that too - but then I compared the values and they seem to all be that percentage of their totals, rounded to the nearest whole. If you add them up they add up to around 100%, so I think the demo values are just evenly split.

common/components/DataViz/MetricsCategoryBarChart.tsx
```diff
@@ -30,10 +31,9 @@ import { rem } from "../../utils";
 import { palette } from "../GlobalStyles";
 import { CustomCursor, CustomYAxisTick } from "./BarChartComponents";
 import Tooltip from "./Tooltip";
-import { ResponsiveBarChartProps, TickProps } from "./types";
-import { splitUtcString } from "./utils";
+import { ResponsiveBarChartProps, ResponsiveBarData, TickProps } from "./types";
 
-const MAX_BAR_SIZE = 150;
+const MAX_BAR_SIZE = 250;
```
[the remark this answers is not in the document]
alfonsogoberjr: This was an experiment I forgot to clean up. Will change back to 150

common/hooks/useBarChart.ts
mxosman: Nice!

common/components/DataViz/CategoryOverviewBreakdown.tsx
alfonsogoberjr: It's possible these are being compared to the wrong maxValue however? I think maxValue is obtained by adding up the dimensions on a single datapoint object. Maybe that's the bug? I'm not sure what else to use as a maxValue instead, if that's the case

common/components/DataViz/CategoryOverviewBreakdown.tsx
```diff
@@ -0,0 +1,114 @@
...
+  color: ${({ color }) => color};
+  text-align: left;
+`;
+
+const LegendValue = styled.span`
+  font-size: 14px;
+  line-height: 22px;
+  margin-left: auto;
+  text-align: right;
+`;
```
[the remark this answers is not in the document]
alfonsogoberjr: Good catch

common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -31,6 +45,19 @@ export function CategoryOverviewLineChart({
...
+      pipe(
+        invertObj,
+        mapObjIndexed(
+          (colorName: string, fill) =>
+            dimensions[Number(colorName.replace("bar", "")) - 1]
+        ),
+        invertObj
+      )(palette.dataViz),
+    [dimensions]
+  );
```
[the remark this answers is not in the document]
alfonsogoberjr: That's strange, I will run a filter over to keys to filter out undefined

common/components/DataViz/CategoryOverviewLineChart.tsx
alfonsogoberjr: Thanks for clarifying though, I didn't know that!

common/components/DataViz/CategoryOverviewBreakdown.tsx
alfonsogoberjr: Fixed!

common/components/DataViz/MetricsCategoryBarChart.tsx
```diff
@@ -52,7 +52,18 @@ const tickStyle = {
...
+      data,
+      dimensionNames,
+      width,
+      metric,
+      maxBarSize,
+      onHoverBar,
+    }: ResponsiveBarChartProps,
+    ref
+  ) => {
+    const [colors] = useState<string[]>(Object.values(palette.dataViz));
```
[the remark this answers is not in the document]
alfonsogoberjr: no because the issue was a discrepancy between the order when it's run for the purpose of the breakdown vs when it's run for the purpose of the line itself. There's an additional render between when those two things happen, moving it into the colorDict gave it sanity

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
alfonsogoberjr: I noticed this too - usually when the dots don't render, they end up rendering eventually. Might be a browser limitation?

common/components/DataViz/CategoryOverviewBreakdown.tsx
```diff
@@ -0,0 +1,74 @@
...
+    )})`}</LegendTitle>
+    {dimensions.map((dimension) => (
+      <LegendItem>
+        <LegendBullet color={data[dimension]?.fill}>▪</LegendBullet>
+        <LegendName color={palette.solid.black}>{dimension}</LegendName>
+        <LegendValue>
+          {renderPercentText(
+            data[dimension]?.value,
+            pipe(
+              values as unknown as (
```
alfonsogoberjr, line 55: This was... annoying. Ramda and types are weird as you know. This was the only way eslint would pass

agency-dashboard/src/CategoryOverview/CategoryOverview.tsx
mxosman: They seem to be showing up consistently now looking at your playtesting link! Please disregard.

mxosman (approved):
LGTM! Thank you so much, Alfonso!

Issue #770, opened 2023-07-17
[Agency Dashboard] Change URL from </agency/<agency_id> to </agency-slug>


lilidworkin:
Agency slug is a lowercase, hyphen separated version of agency name. This is blocked by having a backend endpoint that accepts the agency slug instead of the agency id, because otherwise the FE has no way of converting the slug to an ID to hit the existing endpoint.

Note: PR with backend endpoint is [here](https://github.com/Recidiviz/recidiviz-data/pull/22258)

PR #819, opened 2023-08-04
Changes to bar charts

ivan-kishko:
## Description of the change

> Add dollar sign to money metric values and add metric name to total tooltip

## Type of change
Type: Non-breaking refactor

> All pull requests must have at least one of the following labels applied (otherwise the PR will fail):

| Label                       	| Description                                                                                               	|
|-----------------------------	|-----------------------------------------------------------------------------------------------------------	|
| Type: Bug                   	| non-breaking change that fixes an issue                                                                   	|
| Type: Feature               	| non-breaking change that adds functionality                                                               	|
| Type: Breaking Change       	| fix or feature that would cause existing functionality to not work as expected                            	|
| Type: Non-breaking refactor 	| change addresses some tech debt item or prepares for a later change, but does not change functionality    	|
| Type: Configuration Change  	| adjusts configuration to achieve some end related to functionality, development, performance, or security 	|
| Type: Dependency Upgrade      | upgrades a project dependency - these changes are not included in release notes                             |

## Related issues

Closes #815 

### Development

**This box MUST be checked by the submitter prior to merging**:
- [x] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [ ] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #821, opened 2023-08-04
Remove useEffects from AgencyOverview

alfonsogoberjr:
## Description of the change

Removes dependency tracking and management from AgencyOverview

## Related issues

Relates to #770

## Test Plan

* Navigate to playtesting https://alfonsotest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/
* Click on an agency
* Verify that agency loads after some loading time during which you see the spinner

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
+++ b/agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
@@ -22,11 +22,10 @@ import {
   AgencySystems,
   DataVizAggregateName,
   DataVizTimeRangesMap,
-  Metric,
 } from "@justice-counts/common/types";
 import { printDateAsShortMonthYear } from "@justice-counts/common/utils";
 import { observer } from "mobx-react-lite";
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
 import useAsyncEffect from "use-async-effect";
 
@@ -78,22 +77,10 @@ export const AgencyOverview = observer(() => {
   const [currentSystem, setCurrentSystem] = useState<AgencySystems>(
     availableSystems[0]
   );
-  const [agencyHasAvailableSystems, setAgencyHasAvailableSystems] =
-    useState<boolean>(false);
-  const [
-    metricsByAvailableCategoriesAndSystems,
-    setMetricsByAvailableCategoriesAndSystems,
-  ] = useState<Metric[]>();
-  const [
-    metricsByAvailableCategoriesAndSystemsWithData,
-    setMetricsByAvailableCategoriesAndSystemsWithData,
-  ] = useState<Metric[]>();
-  const [agencyDescription, setAgencyDescription] = useState<string>(
-    agencyDataStore.agencySettingsBySettingType.PURPOSE_AND_FUNCTIONS?.value
-  );
-  const [agencyHomepageUrl, setAgencyHomepageUrl] = useState<string>(
-    agencyDataStore.agencySettingsBySettingType.HOMEPAGE_URL?.value
-  );
+  const agencyDescription =
+    agencyDataStore.agencySettingsBySettingType.PURPOSE_AND_FUNCTIONS?.value;
+  const agencyHomepageUrl =
+    agencyDataStore.agencySettingsBySettingType.HOMEPAGE_URL?.value;
 
   useAsyncEffect(async () => {
     try {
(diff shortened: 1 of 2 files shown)


agency-dashboard/src/NotFound/NotFound.tsx
```diff
@@ -22,23 +22,25 @@ import { HeaderBar } from "../Header";
...
-          </span>
-          <a href="mailto:justice-counts-support@csg.org">
-            justice-counts-support@csg.org
-          </a>
-        </NotFoundText>
-      </NotFoundWrapper>
-      <Footer />
-    </>
-  );
+  return <NotFoundComponent />;
```
mxosman, line 25: nit: I know this was a previous change - but I don't think this needs to be an observer component. Also, do we need to re-export this or can we just export the component itself in L28 and name it `NotFound`?

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
```diff
@@ -107,69 +95,31 @@ export const AgencyOverview = observer(() => {
...
-    ) {
-      navigate("/404");
-    }
-  }, [
-    navigate,
-    agencyDataStore,
-    agencyHasAvailableSystems,
-    metricsByAvailableCategoriesAndSystemsWithData,
-  ]);
+  if (agencyHasNoAvailableSystemsOrHasNoData && !agencyDataStore.loading)
```
mxosman, line 119: nit: I think maybe you can drop the `!agencyDataStore.loading` check here since it will be implied when we reach this line

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
```diff
@@ -107,69 +95,31 @@ export const AgencyOverview = observer(() => {
...
+  const metricsByAvailableCategoriesAndSystemsWithData =
+    metricsByAvailableCategoriesAndSystems.filter(
+      (metric) =>
+        agencyDataStore.datapointsByMetric[metric.key].aggregate.filter(
+          (dp) => dp[DataVizAggregateName] !== null
+        ).length > 0
+    );
+  const agencyHasNoAvailableSystemsOrHasNoData =
+    !agencyHasAvailableSystems ||
+    metricsByAvailableCategoriesAndSystemsWithData.length === 0;
```
mxosman, line 115: [Pure assumption] If we still wanted to use `navigate` we could put all of this after L88 `await agencyDataStore.fetchAgencyData(slug as string);` in the `useAsyncEffect` and do the `agencyHasNoAvailableSystemsOrHasNoData` check. I don't think anything else aside from this check is dependent on these.

Not necessary since we're returning the NotFound in-line, but throwing out a potential option if we preferred navigating away. I feel like having it inline helps that weird blip of the component rendering and the navigating away.

mxosman (commented):
LGTM, Alfonso! Just a couple nits! We're merging this in after Ivan's https://github.com/Recidiviz/justice-counts/pull/819, right?

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
```diff
@@ -78,22 +78,10 @@ export const AgencyOverview = observer(() => {
...
-  const [agencyDescription, setAgencyDescription] = useState<string>(
-    agencyDataStore.agencySettingsBySettingType.PURPOSE_AND_FUNCTIONS?.value
-  );
-  const [agencyHomepageUrl, setAgencyHomepageUrl] = useState<string>(
-    agencyDataStore.agencySettingsBySettingType.HOMEPAGE_URL?.value
-  );
+  const agencyDescription =
+    agencyDataStore.agencySettingsBySettingType.PURPOSE_AND_FUNCTIONS?.value;
+  const agencyHomepageUrl =
+    agencyDataStore.agencySettingsBySettingType.HOMEPAGE_URL?.value;
```
mxosman, line 84: These are being tracked and the component will re-render if the values change/point to something different.

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
```diff
@@ -107,69 +95,31 @@ export const AgencyOverview = observer(() => {
...
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [metricsByAvailableCategoriesAndSystems, agencyDataStore]);
+  const agencyHasAvailableSystems = agencyDataStore.agency?.systems?.some(
+    (system) => availableSystems.includes(system)
+  );
+  const metricsByAvailableCategoriesAndSystems = agencyDataStore.metrics.filter(
+    (metric) =>
+      Object.keys(orderedCategoriesMap).includes(metric.category) &&
+      availableSystems.includes(metric.system.key)
+  );
```
mxosman, line 105: These are also being tracked and the component will re-render if the `agencyDataStore` values (`agencyDataStore.agency.systems` or `agencyDataStore.agency`, or `agencyDataStore.metrics`) change/point to something different.

agency-dashboard/src/NotFound/NotFound.tsx
alfonsogoberjr: Good point!

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
alfonsogoberjr: I'm hesitant to mess around with this more in this PR but this will be addressed in the doc I'm writing about managing state locally in React

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
alfonsogoberjr: I think that was the intention, because the only way to prevent that is by putting them back in useEffect calls

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
alfonsogoberjr: Same as my comment above

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
mxosman: Yeah - for sure! The flow you have here works just fine!

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
mxosman: I meant that these are observables that mobx is tracking (by accessing the properties, they become observables). This behaves similar to something like (rough sketch):

```
const [agencyDescription, setAgencyDescription] = useState(initialValue);

useEffect(() => {
  setAgencyDescription(agencyDataStore.agencySettingsBySettingType.PURPOSE_AND_FUNCTIONS?.value)
}, [agencyDataStore.agencySettingsBySettingType.PURPOSE_AND_FUNCTIONS?.value])

```

mxosman (approved):

PR #844, opened 2023-08-10
Fix bounce back bug in publisher configuration page

### LINKED ISSUE #677: [Bug][Frontend] Sometimes when you click on the Data Viz page and click on another link, it briefly takes you to your destination and suddenly snaps you back to you back to the Data Viz page
There is a strange intermittent bug when you click to go to the Data (visualization) page and go to another page - there are moments when you get redirected back to the Data page. It doesn't happen all the time and there's no clear way to trigger this. Some times it takes a few seconds before you get snapped back, sometimes it happens immediately, sometimes not at all. It feels like there's something funny going on with either the routing or some state that the Data page updating after you leave that page. I was able to capture a video of this behavior happening a couple of times:

https://github.com/Recidiviz/justice-counts/assets/59492998/e3f46adf-2536-4182-b7c4-3ff8e849012e

ivan-kishko:
## Description of the change

> Fix bug where you get thrown back to data/config page after you switch to another page

## Type of change
Type: Bug

## Related issues

Closes #677 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricsConfiguration/MetricsConfiguration.tsx
+++ b/publisher/src/components/MetricsConfiguration/MetricsConfiguration.tsx
@@ -15,83 +15,31 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import { showToast } from "@justice-counts/common/components/Toast";
 import React, { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 
 import { useStore } from "../../stores";
-import { SYSTEM_CAPITALIZED, SYSTEM_LOWERCASE } from "../Global/constants";
 import { Loading } from "../Loading";
 import { useSettingsSearchParams } from "../Settings";
 import MetricConfig from "./Configuration";
 import { MetricsOverview } from "./MetricsOverview";
 
 export function MetricsConfiguration() {
-  const [settingsSearchParams, setSettingsSearchParams] =
-    useSettingsSearchParams();
+  const [settingsSearchParams] = useSettingsSearchParams();
   const { agencyId } = useParams() as { agencyId: string };
-  const { userStore, metricConfigStore } = useStore();
-  const { initializeMetricConfigStoreValues, getMetricsBySystem } =
-    metricConfigStore;
+  const { metricConfigStore } = useStore();
+  const { initializeMetricConfigStoreValues } = metricConfigStore;
 
-  const { system: systemSearchParam, metric: metricSearchParam } =
-    settingsSearchParams;
+  const { metric: metricSearchParam } = settingsSearchParams;
 
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [loadingErrorMessage, setLoadingErrorMessage] = useState<string>();
 
   const initializeMetricConfiguration = async () => {
     setIsLoading(true);
-
     const response = await initializeMetricConfigStoreValues(agencyId);
     if (response instanceof Error) {
       return setLoadingErrorMessage(response.message);
-    }
-
-    const currentAgency = userStore.getAgency(agencyId);
-
-    // now when agency is in the url we still have to check external link
-    // with system and metric search params in it if they belong to agency
-    // (system to agency and metric to system)
-    // if system in agency, go further, if not change system to 1st one in agency
-    if (systemSearchParam && currentAgency?.systems) {
-      const isUrlSystemParamInCurrentAgencySystems =
-        !!currentAgency?.systems.find((system) => system === systemSearchParam);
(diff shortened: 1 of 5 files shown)


ivan-kishko:
Hello @lilidworkin @mxosman. So this is WIP PR because I only fixed metric config page. Let me explain the idea of how I decided to fix that and if you like it I will try to fix data page (explore data now) the same way. So basically what was causing the bug (I wrote about it in task itself) is that after loading data we change state also with setting search params if there are none because we needed like 1 source of truth for app to realize which system and metric to display. The only idea I came up for now is to remove this initial search param setting, and instead if there is no system param just display first from agency. Problem with that is that before logic that I removed was checking whether given system in search params belongs to agency (you know like if you share link) and if not we switch to first in agency as well (example would be like http://localhost:3000/agency/150/data?system=some_jibberish). So I decided if that is the case we just show Not Found Page and if user click "Set Up Metrics" link he goes to first system in agency and see those system metrics. Something similar can be done if system is in agency but metric is not (example http://localhost:3000/agency/161/metric-config?system=jails&metric=jails_jibberish)

ivan-kishko:
Added show not found page if metric is not in system, also home button in not found redirected you to records page and since with new ui we have literally home button I fixed link to go there

publisher/src/components/MetricsConfiguration/MetricsConfiguration.tsx
```diff
@@ -15,84 +15,32 @@
...
   const [loadingErrorMessage, setLoadingErrorMessage] = useState<string>();
 
   const initializeMetricConfiguration = async () => {
     setIsLoading(true);
-
     const response = await initializeMetricConfigStoreValues(agencyId);
     if (response instanceof Error) {
       return setLoadingErrorMessage(response.message);
     }
-
```
alfonsogoberjr, line 50: So this logic was completely removed?

alfonsogoberjr (approved):

alfonsogoberjr:
@mxosman or @lilidworkin should also approve before merge

mxosman:
> Hello @lilidworkin @mxosman. So this is WIP PR because I only fixed metric config page. Let me explain the idea of how I decided to fix that and if you like it I will try to fix data page (explore data now) the same way. 

> So basically what was causing the bug (I wrote about it in task itself) is that after loading data we change state also with setting search params if there are none or system/metric is not in agency/system because we needed like 1 source of truth for app to realize which system and metric to display. 

> The only idea I came up for now is to remove this initial search param setting logic, and instead if there is no system param just display first from agency. Problem with that is that before logic that I removed was checking whether given system in search params belongs to agency (you know like if you share link) 
and if not we switch to first system in agency as well (example would be like http://localhost:3000/agency/150/data?system=some_jibberish switch to http://localhost:3000/agency/150/data?system=legit_system). 

> So I decided if that is the case we just show Not Found Page and if user click "Set Up Metrics" link he goes to first system in agency and see those system metrics. Something similar can be done if system is in agency but metric is not (example http://localhost:3000/agency/161/metric-config?system=jails&metric=jails_jibberish)

Hi Ivan - thank you for this recap and the explanation! I actually really like this solution. I'm personally OK with us showing a Not Found page if the query params are wrong vs. trying to correct the URL with the side effect of having this bounce back. 

cc'ing @lilidworkin to also get her thoughts - are you OK with the UI showing the NotFound page if the user has a typo in the query params (or the query params have a system they don't belong to) instead of redirecting to the first system? We shouldn't have issues internally with users getting to where they need to go and the query params being right (unless we have a bug) - but, I could see issues come up as Ivan pointed out w/ typos when sharing a link (or someone clicking on a link that has the `system` set to one they don't belong to).

> Added show not found page if metric is not in system, also home button in not found redirected you to records page and since with new ui we have literally home button I fixed link to go there

Nice touch! Thank you!

publisher/src/components/MetricsConfiguration/MetricsConfiguration.tsx
mxosman: Yeah - this is the logic that had that intermittent bounce back effect if you went to the Metric Settings page and another page very quickly (example https://github.com/Recidiviz/justice-counts/issues/677).

mxosman (approved):
Thank you, Ivan! Will give @lilidworkin a chance to review as well.

ivan-kishko:
I guess I merge this part and make a separate task for fixing this on data page since here it is more difficult, and I am still not sure why

lilidworkin (approved):
@ivan-kishko thanks for the thoughtful explanation! Your solution sounds totally fine with me. And agreed, we can merge this in and you can work on fixing on the data page separately. Thanks!

PR #868, opened 2023-08-18
Superagencies: Only show superagency system data and add copy explaining how to view child agency data

### LINKED ISSUE #875: [Frontend][Publisher] Superagencies: Only show superagency data in Explore Data

mxosman:
## Description of the change

Update Explore Data to only show Superagency data for superagencies and add a message with instructions on how to view child agency data.

Before:
Screenshot 2023-08-18 at 6 35 08 PM

After:
Screenshot 2023-08-18 at 6 35 20 PM

## Related issues

Closes #875, #876

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataViz/MetricsDataChart.styled.tsx
+++ b/publisher/src/components/DataViz/MetricsDataChart.styled.tsx
@@ -161,8 +161,14 @@ export const DisclaimerTitle = styled.div`
   ${typography.sizeCSS.small}
 `;
 
-export const DisclaimerText = styled.div`
+export const DisclaimerText = styled.div<{
+  textColor?: string;
+  topSpacing?: boolean;
+}>`
   ${typography.sizeCSS.normal}
+  ${({ topSpacing }) => topSpacing && `padding-top: 24px;`};
+  color: ${({ textColor }) =>
+    textColor === "orange" ? palette.solid.orange : palette.solid.darkgrey};
 `;
 
 export const DisclaimerLink = styled.span`
--- a/publisher/src/components/DataViz/MetricsDataChart.styled.tsx
+++ b/publisher/src/components/DataViz/MetricsDataChart.styled.tsx
@@ -18,17 +18,12 @@
 import {
   HEADER_BAR_HEIGHT,
   MIN_DESKTOP_WIDTH,
-  MIN_TABLET_WIDTH,
   palette,
   typography,
 } from "@justice-counts/common/components/GlobalStyles";
 import styled from "styled-components/macro";
 
 const INNER_PANEL_LEFT_CONTAINER_MAX_WIDTH = 314;
-const STICKY_HEADER_WITH_PADDING_HEIGHT = 48;
-const DROPDOWN_WITH_MARGIN_HEIGHT = 79;
-const FIXED_HEADER_WITH_DROPDOWN_HEIGHT = 92;
-const FIXED_HEADER_WITHOUT_DROPDOWN_HEIGHT = 24;
 
 export const NoEnabledMetricsMessage = styled.div`
   min-height: 100%;
@@ -59,13 +54,6 @@ export const MetricsViewPanel = styled.div<{
 
   @media only screen and (max-width: ${MIN_DESKTOP_WIDTH}px) {
     justify-content: unset;
-  }
-
-  @media only screen and (max-width: ${MIN_TABLET_WIDTH}px) {
-    padding-top: ${({ hasSystemsDropdown }) =>
-      hasSystemsDropdown
-        ? `${FIXED_HEADER_WITH_DROPDOWN_HEIGHT + 24}px`
-        : `${FIXED_HEADER_WITHOUT_DROPDOWN_HEIGHT + 24}px`};
   }
 `;
 
@@ -216,7 +204,6 @@ export const CurrentMetricsSystem = styled.div<{
     text-transform: capitalize;
     padding-bottom: 12px;
     padding-top: 24px;
-    position: fixed;
     top: ${({ isSuperagency }) =>
       isSuperagency
         ? `${HEADER_BAR_HEIGHT + 76}px;`
@@ -247,6 +234,12 @@ export const MetricsViewDropdownContainerFixed = styled.div<{
   }
 `;
 
+export const MetricsViewDropdownContainer = styled(
+  MetricsViewDropdownContainerFixed
+)`
+  position: unset;
+`;
+
 export const MetricsViewDropdownLabel = styled.div`
   width: 100%;
   display: flex;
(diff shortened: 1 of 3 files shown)


mxosman:
Hi @jwhrec / @forrestkim00 / anyone else who is interested in chiming in:

Would love your thoughts and improvements on the copy and the placement of the copy (design-wise):

**Copy**: 
If you would like to view data for the agencies you manage, please switch to specific agency you would like to view data for.

**Couple of coloring options - orange/blue - red felt a bit too error-y (open to other colors/styles as well)**:

Screenshot 2023-08-21 at 10 02 50 AM

---

image

publisher/src/components/DataViz/MetricsDataChart.styled.tsx
```diff
@@ -161,8 +161,14 @@ export const DisclaimerTitle = styled.div`
...
 
-export const DisclaimerText = styled.div`
+export const DisclaimerText = styled.div<{
+  textColor?: string;
+  topSpacing?: boolean;
+}>`
   ${typography.sizeCSS.normal}
+  ${({ topSpacing }) => topSpacing && `padding-top: 24px;`};
+  color: ${({ textColor }) =>
+    textColor === "orange" ? palette.solid.orange : palette.solid.darkgrey};
```
alfonsogoberjr, line 171: I want to resurface my comment from an earlier PR of yours where I proposed creating css-aligned styled-component prop utils. Not needed for this PR, but I'm thinking something like

in utils:
```
const styledColor = ({ color }) => `color: ${color};`
const styledPaddingTop = ({ paddingTop }) => `padding-top: ${paddingTop}`
```
and so on.

in styled components:
```
export const DisclaimerText = styled.div<{
  textColor?: string;
  topSpacing?: boolean;
}>`
${styledColor}
`
```

and move this: 

```
textColor === "orange" ? palette.solid.orange : palette.solid.darkgrey};
```
into the inline render

alfonsogoberjr (approved):

publisher/src/components/DataViz/MetricsDataChart.styled.tsx
mxosman: Ah, yes. I remember now - thanks for resurfacing. QQ about this:

> in utils:
> 
> ```
> const styledColor = ({ color }) => `color: ${color};`
> const styledPaddingTop = ({ paddingTop }) => `padding-top: ${paddingTop}`
> ```

That makes sense! 

> and so on.
> 
> in styled components:
> 
> ```
> export const DisclaimerText = styled.div<{
>   textColor?: string;
>   topSpacing?: boolean;
> }>`
> ${styledColor}
> `
> ```

In order to use `${styledColor}` like this, the `styledColor` function parameters would have to match the parameters used here, right? (e.g. change `textColor` to `color` or the above function from `color` to `textColor`)

> and move this:
> 
> ```
> textColor === "orange" ? palette.solid.orange : palette.solid.darkgrey};
> ```
> 
> into the inline render

Sorry, can you clarify where you are referring to? Like the styled component in the render? e.g. something like this?

```
<DisclaimerText textColor={textColor === "orange" ? palette.solid.orange : palette.solid.darkgrey} />
```

forrestkim00:
How do we feel about temporary UI for this? Maybe like a tooltip when the user enters this screen for the first time. Once users understand they have to toggle to access different agency data, they wouldn't need the copy anymore. Does that feels like a mechanism they can remember?

jwhrec:
Any time we have a temporary UI, we hit the issues of multiple users for
one agency and needing to remember who has seen what. I think having a
permanent solution would reduce complexity.

mxosman:
> What prompted us to include this copy? Did we get feedback that users didn't know where to find super-agency/child agency data?

The superagency system hasn't been rolled out yet, so it's not coming from feedback per se - I think this is just hedging for potential confusion when a superagency logs in and goes to "Explore Data" - they might assume that they will be able to view data viz for all of their child agencies from this page, which they can't. They'll have to go to that specific child agencies' Explore Data page.

forrestkim00:
Got it thanks for the explanation! 

I feel like there are only so many places we could place the copy without changing other parts of the UI and turning this into a larger ticket. I wonder if users would be able to find what they're looking for without the copy. If they're used to the concept of toggling between agencies that wouldn't seem like too big of a jump? I guess I'm not too big of a fan of the copy to start because whatever the copy explains should just be made intuitive with good UX (to a certain extent). 


For the short term, I experimented with a couple of options. Placing it to the side, below the header. But these option felt very missable, and I feel like users would probably just end up clicking around trying to find the data rather than just reading the text. But what about a banner? (btw I can make it prettier if we go with this option)
image
image
image

mxosman:
> Got it thanks for the explanation!
> 
> I feel like there are only so many places we could place the copy without changing other parts of the UI and turning this into a larger ticket. I wonder if users would be able to find what they're looking for without the copy. If they're used to the concept of toggling between agencies that wouldn't seem like too big of a jump? I guess I'm not too big of a fan of the copy to start because whatever the copy explains should just be made intuitive with good UX (to a certain extent). Also, a little bit worried about creating design debt.

Totally hear that - I'm fine moving away from adding copy if it feels unnecessary! 

> If we want to keep the copy, I experimented with a couple of options. Placing it to the side, below the header. But these option felt very missable, and I feel like users would probably just end up clicking around trying to find the data rather than just reading the text. But what about a banner? (btw I can make it prettier if we go with this option)

> image

Thank you so much for taking the time to explore these options, Forrest! If we're keeping the copy - of all of the options, I do agree the banner seems to be the one that doesn't get lost visually. It does feel a bit loud for the message it's trying to convey - but I can't think of a different spot that feels and reads right (it might honestly just be the blue - but regardless, of all the placement options - this is definitely the best IMO).

lilidworkin (approved):
I really like the banner @mxosman ! But will obvs let you / @forrestkim00 / @jwhrec make the final call. I don't feel strongly and don't think we need to overthink it

forrestkim00:
I'm good with banner!

mxosman:
Sounds great - I like the banner as well! Forrest, should I use the version in the screenshot above?

forrestkim00:
Could you see if it matches this [link](https://www.figma.com/file/0UZOuDCrwGqWLKOhAyvGbc/Forrest's-First-Figma-File-(FFFF)?type=design&node-id=2319%3A3407&mode=design&t=Xu6KQ84XktJgCln8-1). Just want to make sure it doesn't look super big true to size.

mxosman:
> Could you see if it matches this [link](https://www.figma.com/file/0UZOuDCrwGqWLKOhAyvGbc/Forrest's-First-Figma-File-(FFFF)?type=design&node-id=2319%3A3407&mode=design&t=Xu6KQ84XktJgCln8-1). Just want to make sure it doesn't look super big true to size.

Sorry Forrest - can you clarify what you mean? If the screenshot w/ the banner above matches the Figma node you linked?

I was wondering if I should implement this blue banner design/screenshot or if you had a different design in mind:
image

forrestkim00:
Sorry for the confusion! I think the first SC I just kind of quickly threw together the banner, but the Figma link should be more pixel perfect.

mxosman:
> Sorry for the confusion! I think the first SC I just kind of quickly threw together the banner, but the Figma link should be more pixel perfect.

Oh sorry! Nice - amazing! Thank you so much for the quick Figma mock - will refer to that one. 🙏

mxosman:
Thank you all so much for your help with this one! Sorry I just kinda threw these out there without any good context or warm up into what was going on!

PR #880, opened 2023-08-21
[#22950] [Agency Dashboard] breakdown chart updates

alfonsogoberjr:
## Description of the change

- [x] Order of breakdown legend should be highest to lowest based on most recent observation (e.g. State Appropriations, Contract Beds, Grants, etc.)
- [x] Order of breakdown within the hover tooltip should be highest to lowest based on most recent observation (e.g. State Appropriations, Contract Beds, Grants, etc.)
- [x] Hover tooltip breakdown values should be comma-separated and have a dollar sign for funding and expenses
- [x] Tooltip for breakdown line chart has only year whereas tooltip for bar chart has start month/year - end month/year. Add the start month/year - end month/year to the tooltip on the breakdown line chart to match the tooltip for the bar chart.

## Type of change

Feature

## Test Plan

Navigate to https://alfonsotest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/agency/department-of-corrections/capacity-and-costs

Observe the breakdown values in order matching the magnitudes in the chart. 
Observe the updated formatting of values

## Related issues

Closes #22950

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/CategoryOverview/CategoryOverview.styled.tsx
+++ b/agency-dashboard/src/CategoryOverview/CategoryOverview.styled.tsx
@@ -36,7 +36,7 @@ export const Container = styled.div`
   display: flex;
   flex-direction: column;
   align-items: start;
-  gap: 50px;
+  gap: 32px;
   width: 100%;
   padding: 0 96px;
 `;
@@ -46,7 +46,7 @@ export const TopBlock = styled.div`
   flex-direction: column;
   align-items: start;
   width: ${MIN_METRIC_BOX_WIDTH}px;
-  padding-top: 40px;
+  padding-top: 16px;
 
   & > div > div {
     min-width: unset;
@@ -60,7 +60,7 @@ export const CategoryTitle = styled.div`
 
 export const CategoryDescription = styled.div`
   ${typography.sizeCSS.normal};
-  margin-bottom: 40px;
+  margin-bottom: 16px;
 `;
 
 export const TopBlockControls = styled.div`
@@ -98,7 +98,7 @@ export const MetricsFilters = styled.div`
 export const MetricsFilterButton = styled.div<{ active: boolean }>`
   ${typography.sizeCSS.normal};
   color: ${({ active }) =>
-    active ? palette.solid.blue : palette.solid.darkgrey};
+    active ? palette.solid.blue : palette.highlight.grey9};
   cursor: pointer;
 `;
 
@@ -117,10 +117,14 @@ export const MetricsWrapper = styled.div`
 export const MetricBox = styled.div`
   width: 50%;
   min-width: ${MIN_METRIC_BOX_WIDTH}px;
-  padding: 56px 96px 56px 0;
+  padding-top: 16px;
   display: flex;
   flex-direction: column;
   align-items: start;
+
+  &:not(:first-child) {
+    padding: 100px 96px 56px 0;
+  }
 
   @media only screen and (max-width: 1220px) {
     width: 100%;
@@ -137,15 +141,13 @@ export const MetricName = styled.div`
 
 export const MetricDescription = styled.div`
   ${typography.sizeCSS.normal};
-  margin-bottom: 40px;
+  margin-bottom: 24px;
 `;
 
 export const MetricDataVizContainer = styled.div`
   display: flex;
-  align-items: center;
   justify-content: space-around;
   flex-direction: row;
-  width: 90vw;
   height: 647px;
 
   .recharts-bar-rectangle {
@@ -153,9 +155,14 @@ export const MetricDataVizContainer = styled.div`
   }
 `;
 
+export const MetricDescriptionBarChartWrapper = styled.div`
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+`;
+
 export const BreakdownsTitle = styled.h3`
-  font-size: 18px;
-  line-height: 24px;
-  font-weight: 500;
+  ${typography.sizeCSS.medium};
   margin-left: 50px;
 `;
--- a/common/components/DataViz/CategoryOverviewBreakdown.tsx
+++ b/common/components/DataViz/CategoryOverviewBreakdown.tsx
@@ -31,7 +31,7 @@ import React, { FunctionComponent } from "react";
 
 // eslint-disable-next-line no-restricted-imports
 import { Datapoint } from "../../types";
-import { printDateAsYear } from "../../utils";
+import { printDateAsShortMonthYear } from "../../utils";
 import { palette } from "../GlobalStyles";
 import {
   Container,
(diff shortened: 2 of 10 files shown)


alfonsogoberjr:
cc @jwhrec @lilidworkin @mxosman

common/components/DataViz/CategoryOverviewBreakdown.tsx
```diff
@@ -39,38 +49,46 @@ import {
...
+                    acc + value,
+                  0
+                )
+              )(data) as number,
+              isFundingOrExpenses
+            )}
+          </LegendValue>
+        </LegendItem>
+      ))
+    )(dimensions)}
```
alfonsogoberjr, line 92: Let me explain this, since there have been issues regarding Ramda and readability:

`sort()` and `descend()` are needed to sort the values by magnitude in descending order. I was asked before soomething to the effect of "why not use the native Array prototype methods?" and it boils down to this:

Array prototype methods  like .sort() mutate the array without returning it, which means you can't chain your transformations. It's a known limitation with the Javascript runtime: native String and Array methods basically mangle the object which is referenced in memory. This would be good enough if we were _only_ mutating one array, but in this case we are iterating over `dimeinsions` and actually sorting by a derived value based on `data`. Would be impossible to cleanly write something like this without a chainable sort method that returns a modified copy of the original.

lilidworkin:
Can you deploy this to playtesting @alfonsogoberjr ? It should be working again now!

common/components/DataViz/CategoryOverviewBreakdown.tsx
alfonsogoberjr: tl;dr readability in Ramda is an issue coming from a non-functional paradigm, however once you grasp that `pipe()` and `compose()` are simply higher-order functions for chaining transform functions like `map()`, `filter()` etc, it actually becomes easier to read than the alternative.

alfonsogoberjr:
Deployed to playtesting

common/components/DataViz/CategoryOverviewBreakdown.tsx
mxosman: Thank you for clarifying the sort and descend here! Totally hear your position on using Ramda - personally, I need to get used to it more for its syntax to read clear to me - I'm still not 100% there yet! (And, I wish it didn't force you to cast typings). 

But sorry, I'm confused about the `sort` comment. The native `.sort` method does return the mutated array. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

mxosman (commented):
Hi Alfonso! Quick question about this one:
> * Order of breakdown within the hover tooltip should be highest to lowest based on most recent observation (e.g. State Appropriations, Contract Beds, Grants, etc.)

I don't see a tooltip with breakdowns when I hover - is that what the last checklist item is referring to or are they separate things?
> [ ] Tooltip for breakdown line chart has only year whereas tooltip for bar chart has start month/year - end month/year.

alfonsogoberjr:
@mxosman whoops I was wrong about the return value, I was thinking of methods like push() and pop() which don't return the array. However it does mutate the original which is still problematic from an idempotency perspective

common/components/DataViz/CategoryOverviewBreakdown.tsx
alfonsogoberjr: Good catch, I was wrong about the return value, I was thinking of methods like push() and pop() which don't return the array. However it does mutate the original which is still problematic from an idempotency perspective.

alfonsogoberjr:
> Hi Alfonso! Quick question about this one:
> 
> > * Order of breakdown within the hover tooltip should be highest to lowest based on most recent observation (e.g. State Appropriations, Contract Beds, Grants, etc.)
> 
> I don't see a tooltip with breakdowns when I hover - is that what the last checklist item is referring to or are they separate things?
> 
> > [ ] Tooltip for breakdown line chart has only year whereas tooltip for bar chart has start month/year - end month/year.

Tooltip here refers to the breakdowns themselves. Ivan removed the actual tooltips in response to an earlier product request.

alfonsogoberjr:
@lilidworkin @mxosman playtesting has been updated

agency-dashboard/src/CategoryOverview/CategoryOverview.styled.tsx
```diff
@@ -137,25 +141,28 @@ export const MetricName = styled.div`
...
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+`;
+
 export const BreakdownsTitle = styled.h3`
-  font-size: 18px;
-  line-height: 24px;
-  font-weight: 500;
+  ${typography.sizeCSS.medium};
```
mxosman, line 166: A smattering of styling adjustments to get it to better match the Figma.

common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -87,51 +138,48 @@ export function CategoryOverviewLineChart({
...
+                )}
               />
             )
           )
         )(data)}
         <XAxis
-          dataKey={(datapoint) => printDateAsYear(datapoint.start_date)}
+          dataKey={(datapoint) =>
+            printDateAsShortMonthYear(
+              new Date(datapoint.start_date).getUTCMonth() + 1,
```
mxosman, line 176: This is not ideal - but for now syncs the dates on breakdown line chart as they appear on the bar charts (they were a month/year behind - e.g. Dec 2017 vs Jan 2018). We should have a consistent way of getting these dates all throughout but it's hard to do so in this PR without a slightly more involved refactor.

common/utils/dateUtils.ts
```diff
@@ -153,3 +153,26 @@ export const printDateRangeFromMonthYear = (
     month === 1 ? year : year + 1
   }`;
 };
+
+export const convertShortDateToUTCDateString = (shortDate: string) => {
```
mxosman, line 157: Will add a comment here

mxosman:
This should be good to go as well!

common/components/DataViz/CategoryOverviewLineChart.tsx
lilidworkin: Ah, interesting! Do you know what the underlying cause is here? Agreed that this is totally fine for now, thanks for noticing that they were out of sync and fixing!

lilidworkin (approved):
Checked out the playtesting link, looks beautiful!!

common/components/DataViz/CategoryOverviewLineChart.tsx
mxosman: It's hard to tell - I would think that `datapoint.start_date` is consistent and getting `.getUTCMonth()` should be consistent, but for some reason they're not throughout these components - or how we're parsing them is not consistent or the transformation of them (if they are being transformed?) is not consistent? 

In a refactor, we can probably just do this date parsing once and use those dates throughout these charts instead of creating new Date objects each time. I'll have to trace through how this `data` is being fed to get a better idea, it might be something minor - still puzzling to me.

mxosman: It definitely smells like an OBOE somewhere with these months (e.g. one thing treats months with 0-index Jan-0, Feb-1 and the other with 1 - Jan-1, Feb-2, etc.)

PR #881, opened 2023-08-21
[Bug] Fix issue where reports publish old values

mxosman:
## Description of the change

**Issue:**
CSG noticed that when they update a value in the data entry form and go to publish, the previous value is what is actually saved and published.

**Resolution:**
Remove running the`formStore.validatePreviouslySavedInputs(reportID)` method in the DataEntryReview because this will load the FormStore with the previously saved inputs. This caused an issue because the FormStore which normally has the latest values, gets reset by this method (which I naively thought would be safe to use to get the metric validation when you go from Task Card -> DataEntryReview). 

Will work on a better solution and write up a task, but this should resolve the current issue at hand.

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/DataEntryReview.tsx
+++ b/publisher/src/components/Reports/DataEntryReview.tsx
@@ -86,7 +86,6 @@ const DataEntryReview = () => {
 
   useEffect(() => {
     const initialize = async () => {
-      formStore.validatePreviouslySavedInputs(reportID);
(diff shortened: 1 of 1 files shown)


morden35 (approved):
fantastic debugging!

mxosman:
Thank you for the quick stamp, Michelle!

PR #882, opened 2023-08-21
update allowlist

alfonsogoberjr:
## Description of the change

Update allowlist from [here](https://docs.google.com/document/d/1AMq36vUvUIj_yEjNtjLbGFXjU8ttNk8u3MSqLRnHPoE/edit)

## Type of change

Feature

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/utils/allowlist.ts
+++ b/agency-dashboard/src/utils/allowlist.ts
@@ -19,6 +19,13 @@ import { UserAgency } from "@justice-counts/common/types";
 
 export const ALLOW_LIST: Record<string, boolean> = {
   "Department of Corrections [CLA DEMO]": true,
+  "Puerto Rico Department of Corrections": true,
+  "Louisiana Department Of Public Safety & Corrections": true,
+  "Nebraska Department Of Correctional Services": true,
+  "New Hampshire Department Of Corrections": true,
+  "Massachusetts Department of Correction": true,
+  "New York State Department of Corrections And Community Supervision": true,
+  "Wisconsin Department of Corrections": true,
 };
 
 export const isAllowListed = ({ name }: UserAgency): boolean | undefined =>
--- a/agency-dashboard/src/utils/allowlist.ts
+++ b/agency-dashboard/src/utils/allowlist.ts
@@ -18,7 +18,6 @@
 import { UserAgency } from "@justice-counts/common/types";
 
 export const ALLOW_LIST: Record<string, boolean> = {
-  "Department of Corrections [CLA DEMO]": true,
   "Puerto Rico Department of Corrections": true,
   "Louisiana Department Of Public Safety & Corrections": true,
   "Nebraska Department Of Correctional Services": true,
--- a/agency-dashboard/src/DashboardView/DashboardView.tsx
+++ b/agency-dashboard/src/DashboardView/DashboardView.tsx
@@ -167,7 +167,7 @@ export const DashboardView = observer(() => {
 
   useEffect(() => {
     if (
-      (api.environment === "production" || api.environment === "staging") &&
+      api.environment === "production" &&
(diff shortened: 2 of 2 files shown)


agency-dashboard/src/utils/allowlist.ts
```diff
@@ -19,6 +19,13 @@ import { UserAgency } from "@justice-counts/common/types";
 
 export const ALLOW_LIST: Record<string, boolean> = {
   "Department of Corrections [CLA DEMO]": true,
+  "Puerto Rico Department of Corrections": true,
```
lilidworkin, line 22: Thanks! Couple quick requests:

1) Let's remove "Department of Corrections [CLA DEMO]" from this list -- we decided to create this agency in Staging rather than Prod, and the allowlist shouldn't apply in Staging
2) Let's actually change the logic so the allowlist indeed isn't applied in Staging (change this line [here]))
2) I think currently, the allowlist isn't actually working -- e.g. "Law Enforcement [DEMO]" is not allowlisted, but I can still access it in staging. I think `!isAllowListed(agencyDataStore.agency)` needs to be `      !isAllowListed(agencyDataStore.agency.name)`

lilidworkin: @mxosman do you have thoughts on why the allowlist isn't working? Logic is [here](https://github.com/Recidiviz/justice-counts/blob/f240183112e7cf1a90679734bac6c5a59fdaf470/agency-dashboard/src/DashboardView/DashboardView.tsx#L168). Do we need to take this out of the `useEffect`, since `api` is a mobX store, so is there some weird interaction going on there?

lilidworkin: Actually another thing I'm confused about -- @alfonsogoberjr does the allowlist run on the Agency page or the Capacity and Cost page? It should apply to the Agency page -- at least that was my original intent. Now that I think about it, I guess it should run on both, actually.

mxosman: Ah - just reading this and noticed you had it on #3. Hmm...

I think one thing we could try is to remove that check and redirect from the `DashboardView` component and pass it up a level in the component tree. We'll leave the `DashboardView`'s concerns isolated to whatever it needs to render and have a separate component (e.g. `<Protected>`) to handle the concern of what is allowed to be viewed and either render or redirect. I think we're fighting an uphill battle doing the checking and redirect after the DashboardView has rendered - there's some wonkyness going on in those useEffects that could be removed and the concern itself passed on to another component to handle.

Let me try something to illustrate this more.

lilidworkin: Amazing, thanks for looking into this. Yea, if you don't mind putting up a PR, that would be great. It's actually not too hard to repro locally, you can set your .env to point to staging.

alfonsogoberjr: @mxosman pointed out something that I've wanted to address but always felt time-pressured to ignore, by the way

alfonsogoberjr: He can do his PR I suppose, no sense in duplicating the effort. Unless you are okay with me adding it to this PR @mxosman

mxosman: OK - let me know your thoughts: https://github.com/Recidiviz/justice-counts/pull/883

Wrote some inline explanations. Hopefully this makes sense and works on your ends.

alfonsogoberjr: Approved

alfonsogoberjr: Points 1 and 2 are also addressed here

lilidworkin (approved):
Great, this looks good to go! We can merge this separately from @mxosman 's.

PR #883, opened 2023-08-21
[Proposal][Agency Dashboards] Move allowlist logic up a level to a "wrapper" component

mxosman:
## Description of the change

Moves the allowlist logic to a `<Protected>` component that will determine whether or not an agency's overview and dashboards are available to be viewed. Proposed option for https://github.com/Recidiviz/justice-counts/pull/882.

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/AgencyOverview/AgencyOverview.test.tsx
+++ b/agency-dashboard/src/AgencyOverview/AgencyOverview.test.tsx
@@ -95,11 +95,11 @@ test("renders list of metrics", async () => {
       </MemoryRouter>
     </StoreProvider>
   );
-  await waitFor(async () => {
+  waitFor(async () => {
     const textElementFunding = await screen.findByText(/FUNDING/i);
     expect(textElementFunding).toBeInTheDocument();
   });
-  await waitFor(async () => {
+  waitFor(async () => {
     const textElementExpenses = await screen.findByText(/EXPENSES/i);
     expect(textElementExpenses).toBeInTheDocument();
   });
--- a/agency-dashboard/src/AgencyOverview/AgencyOverview.test.tsx
+++ b/agency-dashboard/src/AgencyOverview/AgencyOverview.test.tsx
@@ -14,15 +14,23 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
-import { render, screen, waitFor } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
+import { runInAction } from "mobx";
 import React from "react";
 import { MemoryRouter } from "react-router-dom";
 
-import { StoreProvider } from "../stores";
+import { rootStore, StoreProvider } from "../stores";
 import { AgencyOverview } from ".";
+
+jest.mock("../utils", () => ({
+  isAllowListed: () => true,
+}));
 
 beforeEach(() => {
   fetchMock.resetMocks();
+  runInAction(() => {
+    rootStore.agencyDataStore.loading = false;
+  });
 });
 
 test("renders list of metrics", async () => {
@@ -95,12 +103,14 @@ test("renders list of metrics", async () => {
       </MemoryRouter>
     </StoreProvider>
   );
-  waitFor(async () => {
-    const textElementFunding = await screen.findByText(/FUNDING/i);
-    expect(textElementFunding).toBeInTheDocument();
-  });
-  waitFor(async () => {
-    const textElementExpenses = await screen.findByText(/EXPENSES/i);
-    expect(textElementExpenses).toBeInTheDocument();
-  });
+
+  // TODO(#884) Update test accounting for the new allowlist logic
+  // const textElementFunding = await screen.findByText(/FUNDING/i);
+  // expect(textElementFunding).toBeInTheDocument();
+
+  // const textElementExpenses = await screen.findByText(/EXPENSES/i);
+  // expect(textElementExpenses).toBeInTheDocument();
+
+  const textElementJusticeCounts = await screen.findByText(/Justice Counts/i);
+  expect(textElementJusticeCounts).toBeInTheDocument();
 });
--- a/agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
+++ b/agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
@@ -17,7 +17,6 @@
 
 import MiniBarChart from "@justice-counts/common/components/DataViz/MiniBarChart";
 import { transformDataForBarChart } from "@justice-counts/common/components/DataViz/utils";
-import { showToast } from "@justice-counts/common/components/Toast";
 import {
   AgencySystems,
   DataVizAggregateName,
@@ -25,9 +24,8 @@ import {
 } from "@justice-counts/common/types";
 import { printDateAsShortMonthYear } from "@justice-counts/common/utils";
 import { observer } from "mobx-react-lite";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useNavigate, useParams } from "react-router-dom";
-import useAsyncEffect from "use-async-effect";
 
 import { Footer } from "../Footer";
 import { HeaderBar } from "../Header";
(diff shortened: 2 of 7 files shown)


agency-dashboard/src/Protected/Protected.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export const Protected: React.FC<PropsWithChildren> = observer(
+  ({ children }) => {
+    const { agencyDataStore } = useStore();
+    const { slug } = useParams();
+    const [loading, setLoading] = useState(true);
+
+    useAsyncEffect(async () => {
+      try {
+        await agencyDataStore.fetchAgencyData(slug as string);
```
mxosman, line 37: We need to first fetch the list of agencies to determine their name and conclude whether or not they are allowed in. This effectively moves this fetch into this wrapper and out of the components, since after this fetch, the store should be hydrated and those components should always have the latest values to render. No net change in the number of calls to fetch data - this still runs every time you access an agency as before - it's just up a level on the component tree.

agency-dashboard/src/AgencyOverview/AgencyOverview.tsx
```diff
@@ -77,18 +75,9 @@ export const AgencyOverview = observer(() => {
...
-      showToast({
-        message: "Error fetching data.",
-        color: "red",
-        timeout: 4000,
-      });
-    }
-  }, []);
+  useEffect(() => {
+    setCurrentSystem(agencyDataStore.agency?.systems[0]);
+  }, [agencyDataStore]);
```
mxosman, line 80: After we've gone past the `Protected` component - the agencyStore should have the latest systems value and this just needs to update the local `currentSystem` state on load w/ the default first agency.

agency-dashboard/src/Protected/Protected.tsx
alfonsogoberjr: this is so much better, thank you!

alfonsogoberjr (approved):

alfonsogoberjr:
Just needs to pass tests and can be merged

agency-dashboard/src/Protected/Protected.tsx
```diff
@@ -0,0 +1,57 @@
...
+          color: "red",
+          timeout: 4000,
+        });
+      }
+    }, [slug]);
+
+    if (loading) {
+      return <Loading />;
+    }
+    return agencyDataStore.agency && isAllowListed(agencyDataStore.agency) ? (
```
lilidworkin, line 51: Can we have the allowlist only apply in production?

agency-dashboard/src/utils/allowlist.ts
```diff
@@ -19,6 +19,7 @@ import { UserAgency } from "@justice-counts/common/types";
 
 export const ALLOW_LIST: Record<string, boolean> = {
   "Department of Corrections [CLA DEMO]": true,
+  "Law Enforcement": true,
```
lilidworkin, line 22: If we only run it in production, then I think we can use the list from @alfonsogoberjr 's PR here.

lilidworkin (commented):
Awesome, thank so much for this @mxosman ! Just a couple quick requests inline.

lilidworkin:
@mxosman were you able to test locally by just hacking the `env` from the backend?

agency-dashboard/src/utils/allowlist.ts
mxosman: When I rebase I'll make sure those changes are used in this file.

mxosman:
> @mxosman were you able to test locally by just hacking the `env` from the backend?

Yes - 100%. All tested against the staging env.

agency-dashboard/src/Protected/Protected.tsx
mxosman: Oh yeah! Yes, we can - let me make the update.

agency-dashboard/src/AgencyOverview/AgencyOverview.test.tsx
```diff
@@ -95,12 +103,14 @@ test("renders list of metrics", async () => {
...
-    const textElementExpenses = await screen.findByText(/EXPENSES/i);
-    expect(textElementExpenses).toBeInTheDocument();
-  });
+
+  // TODO(#884) Update test accounting for the new allowlist logic
+  // const textElementFunding = await screen.findByText(/FUNDING/i);
+  // expect(textElementFunding).toBeInTheDocument();
+
+  // const textElementExpenses = await screen.findByText(/EXPENSES/i);
+  // expect(textElementExpenses).toBeInTheDocument();
```
mxosman, line 112: I couldn't wrangle this test quick enough - it'll be a bit more involved to update - so I'll just add a todo here to not block.

agency-dashboard/src/Protected/Protected.tsx
mxosman: Update made! As a test, switching the conditional to check for Staging env worked for me (it triggered the allowlist - switching it back away from staging rendered everything).

One thing to note - there's another thing that will trigger not showing the agency overview which is the `metricsByAvailableCategoriesWithData` - just for our reference when we're trying to test this - if you are hit with a NotFound it could be allowlist or `metricsByAvailableCategoriesWithData`.

agency-dashboard/src/AgencyOverview/AgencyOverview.test.tsx
lilidworkin: No worries at all -- tysm for the fast work on this

agency-dashboard/src/DashboardView/DashboardView.tsx
```diff
@@ -153,28 +150,6 @@ export const DashboardView = observer(() => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  useAsyncEffect(async () => {
-    try {
-      await agencyDataStore.fetchAgencyData(slug as string);
```
lilidworkin, line 158: Ooh were we fetching agency data more times than needed?? I thought I saw 2 calls in the terminal!

lilidworkin: Oh nvm, saw your comment below

mxosman:
Oh shoot - sorry I merged before double checking - happy to address any feedback in a follow up!

agency-dashboard/src/Protected/Protected.tsx
lilidworkin: Totally -- I noticed that too. I think that's expected because we don't want to show empty dashboards. But agreed it may cause issues when testing/debugging

lilidworkin (approved):
Looks great, tysm!

agency-dashboard/src/DashboardView/DashboardView.tsx
mxosman: Yeah, I think so. This might have been when DashboardView and AgencyOverview were separate? I'm not 100% sure. But, now the call is happening earlier before you get to these components so they can both be removed.

mxosman: Actually, I incorrectly assumed that I could get away with just wrapping the `Protected` component on the first route `/agency/:slug` - but you can technically just go to `.../agency/law-enforcement/capacity-and-costs` and the Dashboard/Category View will load. Quickest solution is to wrap all of them in the Protected component - but will need to think of a solution to only call the fetch if we're going to the components directly/first-time to prevent duplicate calls. Will follow up.

mxosman: Sorry - I definitely didn't explain that clearly! 

I thought I could get away with the below (just wrap <Protected> around the first path `/agency/:slug` element which is `<AgencyOverview />` assuming that if you go to `/agency/:slug/:category` or `/agency/:slug/:dashboard` it will go through the `/agency/:slug` first before loading those path's (category and dashboard) components):

```
      <Route
        path="/agency/:slug"
        element={
          <Protected>
            <AgencyOverview />
          </Protected>
        }
      />
      <Route
        path="/agency/:slug/:category"
        element={<CategoryOverview />}
      />
      <Route
        path="/agency/:slug/dashboard"
        element={<DashboardView />}
      />
```

But, since you can directly go to `/agency/:slug/:category` or `/agency/:slug/dashboard`, we need to wrap the elements around `<Protected>` to enforce the allowlist:

```
      <Route
        path="/agency/:slug"
        element={
          <Protected>
            <AgencyOverview />
          </Protected>
        }
      />
      <Route
        path="/agency/:slug/:category"
        element={
          <Protected>
            <CategoryOverview />
          </Protected>
        }
      />
      <Route
        path="/agency/:slug/dashboard"
        element={
          <Protected>
            <DashboardView />
          </Protected>
        }
      />
```

With this set up - if you go to `/agency/:slug` (AgencyOverview) and from there click on the link that takes you to `/agency/:slug/dashboard` (DashboardView) - the `Protected` fetch & logic runs again. It's not ideal that it runs again, but it protects all those routes (we can add/adjust logic to help prevent that extra call - will just need to think about it for a little bit).

LMK if this is still confusing - I can illustrate it another way!

PR #912, opened 2023-08-30
[Test] Add FormStore test to cover buggy line addition in DataEntryReview

mxosman:
## Description of the change

As a follow up to the buggy line that was introduced that caused previously saved report values to overwrite recently updated values - this test should provide coverage to ensure that between the DataEntryForm and DataEntryReview components that the data in the FormStore (which is what ultimately gets sent to the BE to be saved) remains consistent. 

In the case where the bug was introduced, the values in the FormStore between DataEntryForm and DataEntryReview were different, and Publishing is what officially saved the values.

Test in action (without and with the buggy line):

https://github.com/Recidiviz/justice-counts/assets/59492998/c076b02c-ee62-4b6a-9454-97c663d84ced

## Related issues

Closes #895 

### Development

**This box MUST be checked by the submitter prior to merging**:
- [x] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [x] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #913, opened 2023-08-30
Fix metrics list validation in Publish Review page when coming from a publish task card in the Home page

### LINKED ISSUE #896: [Frontend][Publisher] DataEntryReview: Revisit validation on metrics when coming from task cards

mxosman:
## Description of the change

Implements an alternative solution to validating metrics when going from a publish task card in the Home page directly to the Publish Review page. 

When a user goes to the `ReportDataEntry` > `DataEntryForm` component via opening a record, the previously saved values are loaded and validated in the FormStore so that they appear in the form, and thus, the review page will have validated values it can check to show the appropriate checkmark for each metric. 

Currently, if a user is going directly to the `DataEntryReview` page without having opened up the record in question, the FormStore will have no values loaded for that record - thus, the validation checkmark logic in the review page will have nothing to validate. 

This solution updates the logic in the publish task card and checks to see if the FormStore has loaded & validated values for the record:
* if so, no action needed - loading the review page should be no problem and the validation checkmarks will appear as usual
* if not, we can assume the user has not opened the report or made any changes and we can safely load, validate and reference the previously saved values in the review page

Demo (also sanity checks manual entry to review page is not buggy):

https://github.com/Recidiviz/justice-counts/assets/59492998/4332f978-74f9-4e16-baaa-b8ebc7fb2a5c

As an additional sanity check, I locally rebased this on top of the test added in #912 and tests passed.

## Related issues

Closes #896 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/TaskCard.tsx
+++ b/publisher/src/components/Home/TaskCard.tsx
@@ -48,7 +48,7 @@ export const TaskCard: React.FC<{
       {actionLinks && (
         <Styled.TaskCardActionLinksWrapper>
           {actionLinks.map((action) => {
-            // Exclude "Upload Data" action link from Superagency data entry metric task cards
+            /** Exclude "Upload Data" action link from Superagency data entry metric task cards */
             if (
               isSuperagency &&
               action.label === taskCardLabelsActionLinks.uploadData.label
@@ -95,7 +95,7 @@ export const TaskCard: React.FC<{
                        * If there are no values loaded in the FormStore for metrics and breakdowns, then we can
                        * assume the user has not gone through the DataEntryForm and made updates, and is going directly
                        * from the Task Card to the Publish Review page. If that is the case, the previously saved values
-                       * will be the latest values - and we will neecd to load and validate them in the FormStore in order
+                       * will be the latest values - and we will need to load and validate them in the FormStore in order
                        * to render the validation checkmarks in the metrics list in the review page.
                        */
                       formStore.validatePreviouslySavedInputs(recordID);
--- a/publisher/src/components/Home/TaskCard.tsx
+++ b/publisher/src/components/Home/TaskCard.tsx
@@ -16,14 +16,14 @@
 // =============================================================================
 
 import { Tooltip } from "@justice-counts/common/components/Tooltip";
+import { observer } from "mobx-react-lite";
 import React from "react";
 import { useNavigate } from "react-router-dom";
 
+import { useStore } from "../../stores";
 import HomeStore from "../../stores/HomeStore";
 import { taskCardLabelsActionLinks, TaskCardMetadata } from ".";
 import * as Styled from "./Home.styled";
-import { useStore } from "../../stores";
-import { observer } from "mobx-react-lite";
 
 export const TaskCard: React.FC<{
   metadata: TaskCardMetadata;
--- a/publisher/src/components/Home/TaskCard.tsx
+++ b/publisher/src/components/Home/TaskCard.tsx
@@ -22,11 +22,14 @@ import { useNavigate } from "react-router-dom";
 import HomeStore from "../../stores/HomeStore";
 import { taskCardLabelsActionLinks, TaskCardMetadata } from ".";
 import * as Styled from "./Home.styled";
+import { useStore } from "../../stores";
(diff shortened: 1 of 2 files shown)


publisher/src/components/Home/TaskCard.tsx
```diff
@@ -16,17 +16,20 @@
...
 import HomeStore from "../../stores/HomeStore";
 import { taskCardLabelsActionLinks, TaskCardMetadata } from ".";
 import * as Styled from "./Home.styled";
 
 export const TaskCard: React.FC<{
   metadata: TaskCardMetadata;
   isSuperagency?: boolean;
-}> = ({ metadata, isSuperagency }) => {
+}> = observer(({ metadata, isSuperagency }) => {
+  const { formStore } = useStore();
```
nichelle-hall, line 32: What does `observer` do? Is that tracking any changes to the `metadata` or `isSuperagency` variables?

publisher/src/components/Home/TaskCard.tsx
```diff
@@ -86,7 +89,17 @@ export const TaskCard: React.FC<{
...
-                  if (isPublishAction) {
+                  if (isPublishAction && recordID) {
+                    if (!formStore.hasFormStoreValuesLoaded(recordID)) {
+                      /**
+                       * If there are no values loaded in the FormStore for metrics and breakdowns, then we can
+                       * assume the user has not gone through the DataEntryForm and made updates, and is going directly
+                       * from the Task Card to the Publish Review page. If that is the case, the previously saved values
+                       * will be the latest values - and we will neecd to load and validate them in the FormStore in order
+                       * to render the validation checkmarks in the metrics list in the review page.
+                       */
```
nichelle-hall, line 100: great comment!

publisher/src/components/Home/TaskCard.tsx
```diff
@@ -86,7 +89,17 @@ export const TaskCard: React.FC<{
                       }
                     );
                   }
-                  if (isPublishAction) {
+                  if (isPublishAction && recordID) {
+                    if (!formStore.hasFormStoreValuesLoaded(recordID)) {
+                      /**
+                       * If there are no values loaded in the FormStore for metrics and breakdowns, then we can
+                       * assume the user has not gone through the DataEntryForm and made updates, and is going directly
+                       * from the Task Card to the Publish Review page. If that is the case, the previously saved values
```
nichelle-hall, line 97: So why would that happen? From Task Card -> Publish Review page? They uploaded via bulk upload instead of the form and now they can review the data?

publisher/src/components/Home/TaskCard.tsx
mxosman: Ooo - that's a great question! `observer` is a part of `mobx` (the library we use to manage state throughout the app). Since `mobx` can work with just plain JS, we need to use this `observer` wrapper to get it to work with React so it can handle observing changes in values its tracking and trigger a re-render when those values change.

```
const ReactComponent = observer((props) => {
  const { formStore } = useStore();
  const loading = formStore.loading; 
    // because we have the observer around this component and we're accessing 
    // loading (via `formStore.loading` - emphasis on the `.` property access b/c 
    // this is what the `observer` listens to) - this component will re-render whenever 
    // `loading` changes in the `formStore`

});
```

It's not entirely necessary here since we're just using a method from the `formStore` and that method will likely never be re-defined. But, since we're tapping into the `formStore` I wrapped the component in the `observer` just incase we'll need to access more stores/store values in the future - it would be easy to assume since we're accessing the `formStore` that this component is being observed by mobX when that's not the case (in the past this has been a source of out-of-sync issues). 

It's not tracking changes to `metadata` or `isSuperagency` - just values from our store that we access.

Does that make sense? It's not terribly intuitive at first glance and difficult to explain entirely with just words - happy to illustrate in another way!

nichelle-hall: Yes! That makes sense. Thanks for the explanation.

publisher/src/components/Home/TaskCard.tsx
nichelle-hall: This makes sense! Thanks for the explanation and the video :)

nichelle-hall (approved):

PR #914, opened 2023-08-31
Add typeahead search in agency dropdown menu

### LINKED ISSUE #877: [Frontend][Publisher] Superagencies: Add a typehead to the agency dropdown menu

mxosman:
## Description of the change

Adds a typeahead search input in the agency dropdown menu to help filter agency options.

https://github.com/Recidiviz/justice-counts/assets/59492998/17ec0d07-9d94-4afa-980b-538329526908

Playtesting Link: https://mahmoudtest---justice-counts-web-qqec6jbn6a-uc.a.run.app/agency/161/

## Related issues

Closes #877

### Development

### Code review

### DIFF
--- a/common/components/Dropdown/Dropdown.styled.tsx
+++ b/common/components/Dropdown/Dropdown.styled.tsx
@@ -22,7 +22,7 @@ import {
   DropdownToggle,
 } from "@recidiviz/design-system";
 // eslint-disable-next-line no-restricted-imports
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 
 import { palette, typography } from "../GlobalStyles";
 import { ToggleHover, ToggleSize } from "./types";
@@ -94,12 +94,7 @@ export const CustomDropdownMenu = styled(DropdownMenu)<{
     menuFullWidth ? "width: 100%;" : "min-width: 293px;"}
 `;
 
-export const CustomDropdownMenuItem = styled(DropdownMenuItem)<{
-  color?: "green" | "red";
-  disabled?: boolean;
-  highlight?: boolean;
-  noHover?: boolean;
-}>`
+const customDropdownMenuItemBaseCSS = css`
   margin: 0 !important;
   width: 100%;
   min-width: 100%;
@@ -111,6 +106,15 @@ export const CustomDropdownMenuItem = styled(DropdownMenuItem)<{
   border-radius: 3px;
   text-transform: capitalize;
   ${typography.sizeCSS.normal};
+`;
+
+export const CustomDropdownMenuItem = styled(DropdownMenuItem)<{
+  color?: "green" | "red";
+  disabled?: boolean;
+  highlight?: boolean;
+  noHover?: boolean;
+}>`
+  ${customDropdownMenuItemBaseCSS}
   color: ${({ color, highlight }) => {
     if (color === "green") {
       return palette.solid.green;
@@ -173,4 +177,32 @@ export const OptionLabelWrapper = styled.div<{
   gap: 16px;
   color: inherit !important;
   ${({ highlightIcon }) => highlightIcon && "padding-left: 8px;"}
+`;
+
+export const CustomInputWrapper = styled.div`
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 20px 16px 8px 16px;
+`;
+
+export const CustomInput = styled.input`
+  ${typography.sizeCSS.normal}
+  width: 100%;
+  padding: 4px 8px 4px 8px;
+  border: 1px solid ${palette.solid.lightgrey3};
+  color: ${palette.highlight.grey9};
+  background: ${palette.solid.lightgrey2};
+  border-radius: 2px;
+
+  &::placeholder {
+    color: ${palette.highlight.grey5};
+  }
+`;
+
+export const NoResultsFoundWrapper = styled.div`
+  ${customDropdownMenuItemBaseCSS}
+  color: ${palette.solid.darkgrey};
+  padding-left: 24px;
 `;
--- a/common/components/Dropdown/Dropdown.tsx
+++ b/common/components/Dropdown/Dropdown.tsx
@@ -15,7 +15,7 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import React from "react";
+import React, { useState } from "react";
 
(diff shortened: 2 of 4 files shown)


common/components/Dropdown/Dropdown.styled.tsx
```diff
@@ -111,6 +106,15 @@ export const CustomDropdownMenuItem = styled(DropdownMenuItem)<{
...
   ${typography.sizeCSS.normal};
+`;
+
+export const CustomDropdownMenuItem = styled(DropdownMenuItem)<{
+  color?: "green" | "red";
+  disabled?: boolean;
+  highlight?: boolean;
+  noHover?: boolean;
+}>`
+  ${customDropdownMenuItemBaseCSS}
```
mxosman, line 117: No change here except for extracting similar styles into a `customDropdownMenuItemBaseCSS` variable

common/components/Dropdown/Dropdown.tsx
```diff
@@ -92,41 +110,66 @@ export function Dropdown({
...
-              )
+                  {/**
+                   * icon: a label icon that's always visible in the dropdown menu
+                   * highlightIcon: an icon that indicates (and is only visible on) the menu option that is currently active
+                   */}
+                  {icon}
+                  {optionLabel}
+                  {highlight && highlightIcon}
+                </Styled.OptionLabelWrapper>
+              </Styled.CustomDropdownMenuItem>
```
mxosman, line 161: L131 is the only change within this block

common/components/Dropdown/Dropdown.tsx
```diff
@@ -61,7 +62,24 @@ export function Dropdown({
...
+    const regex = new RegExp(`${val}`, `i`);
+    setFilteredOptions(() =>
+      options.filter((option) => regex.test(option.label as string))
+    );
+  };
+
+  useEffect(() => {
+    const timeout = setTimeout(() => inputRef.current?.focus(), 0);
+    return () => clearTimeout(timeout);
+  }, [inputValue]);
```
mxosman, line 81: So, there was a funny thing happening with the input focus on re-renders using the `design-system`'s Dropdown (which controls focus in its own way). When you type, there are moments when the dropdown list would re-render and the focus on the input would reset/go away causing you to have to click back into the input to finish. This is because we're adding an input _within_ the dropdown that's changing/re-rendering as the user types.

This is the best solution I could come up with that plays nicely and maintains the focus between re-renders. It's only necessary to do here because we're including the search box within the dropdown (if it lived outside of the dropdown, it would maintain its focus just fine). LMK if any of this is confusing or if anyone would like to talk through this solution. The `timeout` is only being called when the input values change and is cleaned up each time.

common/components/Dropdown/Dropdown.tsx
```diff
@@ -61,7 +62,29 @@ export function Dropdown({
...
 }: DropdownProps) {
+  const [filteredOptions, setFilteredOptions] = useState(options);
+  const [inputValue, setInputValue] = useState("");
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const updateFilteredOptions = (val: string) => {
+    const regex = new RegExp(`${val}`, `i`);
+    setFilteredOptions(() =>
+      options.filter(
+        (option) => typeof option.label === "string" && regex.test(option.label)
```
lilidworkin, line 75: Just OOC, can typeof option.label ever not be string, or is this just to satisy type checker?

common/components/Dropdown/Dropdown.tsx
lilidworkin: Thanks for the explanation! Makes sense to me

common/components/Dropdown/Dropdown.tsx
```diff
@@ -92,41 +115,66 @@ export function Dropdown({
...
-                  highlight={highlight && !highlightIcon}
+        <>
+          {withTypeaheadSearch && (
+            <Styled.CustomInputWrapper>
+              <Styled.CustomInput
+                ref={inputRef}
+                id="dropdown-typeahead"
+                name="dropdown-typeahead"
+                type="search"
+                placeholder="Search for Agency"
```
lilidworkin, line 126: Technically, maybe this text should be a property on Dropdown, like `withTypeaheadSearch` is, to make it fully generic. Otherwise we can't really re-use the Dropdown component with `withTypeaheadSearch = true` anywhere else. But that's probably a premature optimization, and I'm also wary of the number of properties getting out of hand. Not entirely sure what is best practice here

lilidworkin: Maybe best practice is to have `withTypeaheadSearch` not be a boolean, but rather a dictionary of sub-properties needed to customize the typeahead search?

Totally not merge blocking, just thinking out loud

lilidworkin (approved):
Looks SO good! Love it!

common/components/Dropdown/Dropdown.tsx
mxosman: Great question! Yeah - `option.label` is flexible to where it can take a ReactNode or a string. In the agency dropdown, it will always be a string list, so this is mostly to satisfy the type-checker, and since we can only `regex.test` on a string, this skips over any ReactNode so we don't run into issues.

common/components/Dropdown/Dropdown.tsx
mxosman: Ah - great idea & catch! I didn't think about making the `placeholder` flexible too - of course! If it was just a matter of showing the search without any placeholder text - then I think `boolean` is fine - but if we're customizing other properties, then most certainly a dictionary of props is ideal. Thank you for calling this out!

mxosman: What do you think of this? [573425a](https://github.com/Recidiviz/justice-counts/pull/914/commits/573425a05cef43ab4d2164bcca52c84a3325d3a8)

Mostly curious if it's clear that adding the `typeaheadSearch` prop and setting it to a dictionary alone means you want a typeahead, and just have the dictionary include the customizations. Or, I can include an explicit boolean prop within the `typeaheadSearch` dictionary - e.g. `{ active: <boolean>, placeholder: <string>, ...other customizations }`

lilidworkin: I love this! Yea I think it's perfectly clear and fine to have it as you do now! Thanks for making this change

lilidworkin (approved):

mxosman:
Thank you so much for your eyes!

PR #918, opened 2023-09-01
Agency Dashboard Refactor (Ramda + Chart Components)

mxosman:
## Description of the change

Last wave of refactoring & clean up for the Agency Dashboard. Mostly replacing Ramda with native JS, cleaning up the chart hooks and components, and deprecating the `DashboardView` components.

Playtesting Link: https://mahmoudtest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/

### Development

### Code review


agency-dashboard/src/Footer/Footer.tsx
```diff
@@ -23,35 +23,36 @@ import React from "react";
...
+          endorse, this Web site (including, without limitation, its content,
+          technical infrastructure, and policies, and any services or tools
+          provided).
+        </Styled.FooterBottomInfo>
+        <Styled.FooterCopyrights>
+          © 2023 The Council of State Governments. All Rights Reserved.
+        </Styled.FooterCopyrights>
+      </Styled.FooterContent>
+    </Styled.FooterWrapper>
+  );
```
mxosman, line 57: We can keep this as is w/o the `return null` and just remove the Footer from each component that has it until we need it again.

lilidworkin: Much better, thank you!

common/components/DataViz/CategoryOverviewBreakdown.tsx
```diff
@@ -41,61 +27,44 @@ import {
...
+  const sortedDimensions = dimensions.sort(
+    (dimA: string, dimB: string) =>
+      Number(data[dimB].value) - Number(data[dimA].value)
+  );
+
+  return (
+    <Container>
+      <LegendTitle>
+        {data.start_date &&
+          (hoveredDate ? displayDate : `Recent (${displayDate})`)}
```
lilidworkin, line 52: So much cleaner!

common/components/DataViz/CategoryOverviewLineChart.tsx
```diff
@@ -110,43 +106,32 @@ export function CategoryOverviewLineChart({
...
-          stroke={colorDict[dimension]}
-          type="monotone"
-        />
-      );
-      lineDefinitions = [newLine, ...lineDefinitions];
-    });
-    return lineDefinitions;
-  };
-  const { legendData, referenceLineHeight } = useLineChartLegend(
+  /** Creates a { [dimension]: colorFromDataVizPalette } map */
```
lilidworkin, line 109: love these comments

common/hooks/useLineChart.ts
```diff
@@ -39,45 +35,41 @@ export type LineChartHookProps = Partial<UserAgency> & {
...
-      }
-      return [];
-    },
-    [datapointsByMetric]
-  );
+  const getLineChartDataFromMetric = (metric: Metric) => {
+    if (datapointsByMetric) {
+      /**
+       * Gets the first set of disaggregated datapoints.
+       * NOTE: This assumes there's just one breakdown per metric. We will need to adjust this
```
lilidworkin, line 47: Thanks for calling this assumption out!

common/hooks/useLineChartLegend.ts
```diff
@@ -15,93 +15,58 @@
...
-      )(datapoints) as Datapoint;
+    return { ...datapoint, ...dimensionsValueFill } as LegendData;
+  };
 
-    const getHoveredDatapoint = (datapoints: Datapoint[]): Datapoint =>
-      pipe(
-        filter<Datapoint>(propEq(hoveredDate, "start_date")),
-        head
-      )(datapoints) as Datapoint;
+  const getLatestDatapoint = (dps: Datapoint[]): Datapoint =>
```
lilidworkin, line 39: It seems like in the older version of this function, we checked if the datapoint had any dimensions first? Just confirming that that is actually _not_ what we want to do

lilidworkin: Also is it safe to assume that datapoints will be sorted properly before being passed to this function? Might be worth adding a comment

lilidworkin (approved):
Awesome, thank you so much for taking the initiative to do this cleanup! It looks SO MUCH better! I didn't do the most thorough of reviews, but basically just sanity checked that things look good at a high level, and then spot checked a few things in more depth. If you think it would be helpful, we could do a playtesting of the dashboards -- by now we might have enough "real" dashboards with data (i.e. from bottom section here) so we could check that all of those look reasonable https://docs.google.com/document/d/1AMq36vUvUIj_yEjNtjLbGFXjU8ttNk8u3MSqLRnHPoE/edit

common/hooks/useLineChartLegend.ts
mxosman: > It seems like in the older version of this function, we checked if the datapoint had any dimensions first? Just confirming that that is actually _not_ what we want to do

Yeah - this one was throwing me a little bit. To be honest, I wasn't entirely sure what its original purpose was or if it was even necessary to do? I think the thing that uses this determines whether or not the user is hovered over a datapoint - if they've loaded the page and haven't hovered over a datapoint in the breakdown chart, it will display the latest datapoint's breakdowns (if that makes sense).

From my understanding, the older version was taking in a `datapoints` array, reversing the array, then filtering out datapoints that have dimensions w/ no number values and then getting the first element in this now reversed and filtered array? I don't know why, but in my head it feels like this might create a weird scenario where you have a list of datapoints `[x,y,z]`, and if `z` doesn't have dimensions with number values but `y` does, then you'll get `y` as the latest datapoint - which doesn't make a lot of sense to me, esp. since we're not filtering out datapoints like this elsewhere. Maybe there's something I'm not thinking about here...

> Also is it safe to assume that datapoints will be sorted properly before being passed to this function? Might be worth adding a comment

Good question! I viewed this as a helper function that's not necessarily concerned with ordering the datapoints, but will just give you the datapoint at the end of the list (we don't even really need this function since we can just do `datapoints[datapoints.length - 1]` directly, but I felt like this abstracted that and self-commented nicely). I was trying to figure out where the sorting of the datapoints was happening and I don't see any sorting the FE is doing w/ datapoints. I'll take a look and see where that might be happening tomorrow. Would it feel better to adjust this to do a sort?

mxosman:
> Awesome, thank you so much for taking the initiative to do this cleanup! It looks SO MUCH better! I didn't do the most thorough of reviews, but basically just sanity checked that things look good at a high level, and then spot checked a few things in more depth. 

Thank you so much, Lili! I always appreciate your time and review (regardless of depth)!

>If you think it would be helpful, we could do a playtesting of the dashboards -- by now we might have enough "real" dashboards with data (i.e. from bottom section here) so we could check that all of those look reasonable https://docs.google.com/document/d/1AMq36vUvUIj_yEjNtjLbGFXjU8ttNk8u3MSqLRnHPoE/edit

Totally - I think it would be super helpful. I can set something up next week.

common/hooks/useLineChartLegend.ts
lilidworkin: > Yeah - this one was throwing me a little bit. To be honest, I wasn't entirely sure what its original purpose was or if it was even necessary to do?

Yea I'm also confused by it. I imagine of course that you tested that the breakdown chart seems to work properly after this refactor? I'm not too worried then.

> I viewed this as a helper function that's not necessarily concerned with ordering the datapoints, but will just give you the datapoint at the end of the list

Got it! I think the BE orders the datapoints, so I think we're fine. nit, but I might name this function `getLastDatapoint` then though, since it's not guaranteed to get the latest one, depending on what array is passed in?

lilidworkin (approved):

common/hooks/useLineChartLegend.ts
mxosman: > Yea I'm also confused by it. I imagine of course that you tested that the breakdown chart seems to work properly after this refactor? I'm not too worried then.

For sure! The charts are working properly. Now that I'm thinking about this, I think I might have an idea of the intent - but LMK if you think this is worth filtering for. So, when you load the charts for the first time, the default set of breakdowns that will appear will be the latest datapoints (in the below example Jan 2022's breakdowns) -- I think the filtering was done incase the latest datapoints had no breakdown values and if so, it'll show (for example) Jan 2021's breakdowns as the default on page load (or whatever the latest datapoint is with at least one breakdown value).

Screenshot 2023-09-08 at 6 01 54 PM

> > I viewed this as a helper function that's not necessarily concerned with ordering the datapoints, but will just give you the datapoint at the end of the list
> 
> Got it! I think the BE orders the datapoints, so I think we're fine. nit, but I might name this function `getLastDatapoint` then though, since it's not guaranteed to get the latest one, depending on what array is passed in?

Agreed! Love that name - thank you for this!

lilidworkin: @mxosman 

> So, when you load the charts for the first time, the default set of breakdowns that will appear will be the latest datapoints (in the below example Jan 2022's breakdowns) -- I think the filtering was done incase the latest datapoints had no breakdown values and if so, it'll show (for example) Jan 2021's breakdowns as the default on page load (or whatever the latest datapoint is with at least one breakdown value).

Ah okay this makes sense! That might be worth filtering for. I think we can merge this PR in as-is, but can you spin up a thread with Jessica/Forrest asking them what you think the right behavior is there? We can address in a followup PR!

I seem to remember there was a discussion with Matt around this, like what to do if we have different time range of data available for the aggregate data vs. breakdown data, but I don't remember what the conclusion is. Could be worth asking his opinion too

mxosman: Thanks, Lili! 

>can you spin up a thread with Jessica/Forrest asking them what you think the right behavior is there? We can address in a followup PR!

Do you mind if we discuss during playtesting this week? I can include this in the playtesting doc.

>I seem to remember there was a discussion with Matt around this, like what to do if we have different time range of data available for the aggregate data vs. breakdown data, but I don't remember what the conclusion is. Could be worth asking his opinion too

I'll look back at the Slack conversations and see what I can find.

lilidworkin: Yea discussing during playtesting is a great idea!

PR #925, opened 2023-09-07
[Bug] Fix out-of-sync behavior in agency dropdown menu 

mxosman:
## Description of the change

Matt (CSG) surfaced an issue with the agency dropdown causing weird behavior with navigation as described below:

"If you log in and then select a new agency from the drop down in the upper left menu, then navigate to any of the menu items using the upper left menu dropdown, it returns you to the agency you originally logged in as, rather than the agency that you've selected"

There was another issue with the agency dropdown where if you select an agency, the highlight icon doesn't update (e.g. you're in Agency 1 when you first login, then switch to Agency 2 - the dropdown highlight icon is still on Agency 1). These two issues are surprisingly related - and all to do with us not having a fresh copy of the dropdown `options` from the parent `Menu` component.

I made an incorrect assumption by initializing the below `useState` with the `options` that's passed into the `Dropdown` component from the parent `Menu` component.

```
...

const [filteredOptions, setFilteredOptions] = useState<DropdownOption[]>(options);

...

  { filteredOptions.map(...
```

The issue with this is that it persists the original `options` between re-renders which is not what we want when the parent component updates the `options` array (which it does for example when you click on another agency - the `highlight` boolean gets updated in the list and a freshly updated `options` list is passed to the `Dropdown` component). 

Instead of initializing the `useState` with `options`, we can create a `useEffect` that will update the `filteredOptions` appropriately when we get a new list of `options`. 

```
...

const [filteredOptions, setFilteredOptions] = useState<DropdownOption[]>();

...
  useEffect(() => {
    setFilteredOptions(options);
    setInputValue("");
  }, [options]);
...

  { filteredOptions.map(...
```

This way, when a user clicks on an item in the menu, the `Dropdown` component will always have the latest `options` list that will have the up-to-date navigation paths and highlight boolean.

Demo (please feel free to sanity check): https://mahmoudtest---justice-counts-web-qqec6jbn6a-uc.a.run.app

### Development

### Code review

### DIFF
--- a/common/components/Dropdown/Dropdown.tsx
+++ b/common/components/Dropdown/Dropdown.tsx
@@ -82,7 +82,7 @@ export function Dropdown({
 
   useEffect(() => {
     setFilteredOptions(options);
-    if (inputValue) setInputValue("");
+    setInputValue("");
   }, [options]);
 
   useEffect(() => {
--- a/common/components/Dropdown/Dropdown.tsx
+++ b/common/components/Dropdown/Dropdown.tsx
@@ -82,6 +82,7 @@ export function Dropdown({
 
   useEffect(() => {
     setFilteredOptions(options);
+    if (inputValue) setInputValue("");
   }, [options]);
 
   useEffect(() => {
--- a/common/components/Dropdown/Dropdown.tsx
+++ b/common/components/Dropdown/Dropdown.tsx
@@ -67,7 +67,7 @@ export function Dropdown({
   highlightIcon,
   typeaheadSearch,
 }: DropdownProps) {
-  const [filteredOptions, setFilteredOptions] = useState(options);
+  const [filteredOptions, setFilteredOptions] = useState<DropdownOption[]>();
   const [inputValue, setInputValue] = useState("");
   const inputRef = useRef<HTMLInputElement>(null);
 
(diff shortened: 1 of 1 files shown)


common/components/Dropdown/Dropdown.tsx
```diff
@@ -80,6 +80,11 @@ export function Dropdown({
     );
   };
 
+  useEffect(() => {
+    setFilteredOptions(options);
+    if (inputValue) setInputValue("");
```
mxosman, line 85: This little addition clears what's in the search box when we get a new updated list of dropdown options. Otherwise, when you type "abc" and click on an agency (trigger an update in the `options` list), the search box will still show "abc".

common/components/Dropdown/Dropdown.tsx
```diff
@@ -80,6 +80,11 @@ export function Dropdown({
     );
   };
 
+  useEffect(() => {
+    setFilteredOptions(options);
+    setInputValue("");
```
mxosman, line 85: This little addition clears what's in the search box when we get a new updated list of dropdown options. Otherwise, when you type "abc" and click on an agency (trigger an update in the options list), the search box will still show "abc".

lilidworkin (approved):
Thanks for the quick fix @mxosman !! I can deploy this to prod tomorrow morning :)

mxosman:
> Thanks for the quick fix @mxosman !! I can deploy this to prod tomorrow morning :)

Thank you so much, Lili! I don't mind doing a staging + prod deploy after this is merged in if you're OK with that. One less thing for you to do tomorrow!

PR #926, opened 2023-09-08
add spinner while downloading custom spreadsheet

nichelle-hall:
## Description of the change

https://github.com/Recidiviz/justice-counts/assets/19961693/ee6fd262-012c-4d73-879f-ddf689bbaa2b

Added a spinner to the template buttons! I also deleted the static templates since we will be generating all templates to match a agency's metric config. 

### Development

### Code review


publisher/src/components/DataUpload/DataUpload.styles.tsx
```diff
@@ -479,3 +479,10 @@ export const ListOfModifiedRecordsContainer = styled.div`
 `;
 
 export const ModifiedRecordTitle = styled.div``;
+
+export const MiniLoaderWrapper = styled.div`
+  position: absolute;
+  z-index: 3;
+  margin-left: 15px;
+  margin-top: 8px;
```
nichelle-hall, line 487: Is it okay that these are hard coded? I wanted them to be on a certain part of the excel symbol.

publisher/src/components/DataUpload/InstructionsTemplate.tsx
```diff
@@ -180,6 +182,10 @@ export const systemToDetails: { [system: string]: SystemDetails } = {
 export const GeneralInstructions: React.FC<
   GeneralInstructionsTemplateParams
 > = ({ agencyId, systems, downloadTemplate, isDownloading }) => {
+  const [showSinglePageTemplateSpinner, setShowSinglePageTemplateSpinner] =
+    React.useState(false);
+  const [showMultiPageTemplateSpinner, setShowMultiPageTemplateSpinner] =
+    React.useState(false);
```
nichelle-hall, line 188: Created new state variables instead of using `isDownloading` because I only want the spinner to show up on the button that the user clicked. `isDownloading` is too general and I couldn't use it to distinguish what button was clicked.

lilidworkin (commented):
Will let @mxosman review this, but thanks for tackling it @nichelle-hall ! Also exciting to see all the old templates deleted!

mxosman:
Awesome, Nichelle! Will take a closer look next week!

publisher/src/components/DataUpload/DataUpload.styles.tsx
mxosman: Yeah - totally! I usually start putting things in a variable when I realize I'm repeating the same styles across multiple things - but this looks totally fine to me and what I would do.

publisher/src/components/DataUpload/InstructionsTemplate.tsx
```diff
@@ -218,16 +224,25 @@ export const GeneralInstructions: React.FC<
           return (
             <DownloadAutogeneratedTemplateBox
               key={system}
-              onClick={() =>
-                downloadTemplate(system, /* isSinglePageTemplate */ false)
-              }
+              onClick={async () => {
```
mxosman, line 227: nit: I would define async functions outside of the render and not directly on the `onClick` handler.

publisher/src/components/DataUpload/InstructionsTemplate.tsx
mxosman: That makes sense! Although - I don't think that this will play well with agencies with mutliple systems. I think when you click on either the single or multiple page template, them sharing a single state each will cause either all the single page buttons to show the loader or all the multiple page templates to show the loader (depending on what you click) 

What if there was one state that was an object with: 
```
{ [key that each button uses when rendering]: boolean }
```

The `key` could be something like `system + "_SINGLE"` or `system + "_MULTI"` (e.g. `LAW_ENFORCEMENT_SINGLE`)

Then each button will have its own state and whichever button is clicked will just update that piece of the state to `true` - and each button will show the loader independently.

publisher/src/components/DataUpload/InstructionsTemplate.tsx
```diff
@@ -238,16 +253,23 @@ export const GeneralInstructions: React.FC<
           return (
             <DownloadAutogeneratedTemplateBox
               key={system}
-              onClick={() =>
-                downloadTemplate(system, /* isSinglePageTemplate */ true)
-              }
+              onClick={async () => {
+                setShowSinglePageTemplateSpinner(true);
+                await downloadTemplate(system, /* isSinglePageTemplate */ true);
+                setShowSinglePageTemplateSpinner(false);
```
mxosman, line 259: same nit - we could probably create one function for this

publisher/src/components/DataUpload/DataUpload.styles.tsx
mxosman: Ah sorry - you probably meant is it OK to hard-code the positioning specifically - I think it's totally fine.

publisher/src/components/DataUpload/InstructionsTemplate.tsx
nichelle-hall: oof great point! I'll try this out!

publisher/src/components/DataUpload/InstructionsTemplate.tsx
```diff
@@ -248,6 +282,11 @@ export const GeneralInstructions: React.FC<
                 {systemName} Single Page
                 <span>Download</span>
               </DownloadTemplateSystem>
+              {templateKeyToLoadingStatus[`${system}SINGLE`] === true && (
```
mxosman, line 285: typo - `_SINGLE` otherwise the single page template loaders won't show up

mxosman (approved):
Yay - changes look great, Nichelle! Left one comment to address otherwise this is LGTM!

PR #943, opened 2023-09-14
[Agency Dashboards][Bug] Intermittent crash when hovering on charts 

### LINKED ISSUE #944: [Agency Dashboards][Bug] Intermittent crash when hovering on charts
During a playtesting of Agency 496's dashboard in production, there were times when hovering over a chart crashed the app. This is not easy to reproduce since the charts re-render so often based on mouse movement, but there are times during the transition between Recent and All, where hovering back on a chart after switching the datapoint returned by getHoveredDatapoint comes back as undefined. This happens when there are no datapoints w/ a date that matches the hoveredDate - and this in turn causes a crash within the transformDataForLegend logic when we set the value for dimensionsValueFill.

mxosman:
## Description of the change

During a playtesting of Agency 496's dashboard in production, there were times when hovering over a chart crashed the app. This is not easy to reproduce since the charts re-render so often based on mouse movement, but there are times during the transition between Recent and All, where hovering back on a chart after switching the `datapoint` returned by `getHoveredDatapoint` comes back as `undefined`. This happens when there are no datapoints w/ a date that matches the `hoveredDate` - and this in turn causes a crash within the `transformDataForLegend` logic when we set the value for `dimensionsValueFill`. This at the very least handles the `undefined` cases (and hopefully prevents similar crashes).

I'm still trying to figure out why there's a brief mismatch sometimes and it's a bit of a challenge logging this because it's intermittent and the amount of natural re-renders that happen from moving your mouse over the charts spits out a ton of log info. I was able to see that before the crash, the `hoveredDate` that's causing issues is `Wed, 01 Jan 2020 00:00:00 GMT` in the Funding chart - normally, the `onHoverBar` logic in the charts doesn't pick up on hovers over bars that have no datapoints (the ones that show up in a faint red color w/ `Not Recorded` on the tooltip) - and this one shouldn't have triggered anything because there is no Jan 2020 record. But, for some reason when you switch from All to Recent and hover over the chart quickly (almost like you're doing it in the middle of the transition), it picks up on the `Wed, 01 Jan 2020 00:00:00 GMT` bar and that starts returning undefined throughout. I'm a bit stumped for now but will keep digging into this (time-boxing because I sank a few hours into it and am still coming up puzzled). My latest hunch so far is something funny going on with the `onHoverBar` logic, or something funny going on between hovering while the chart is transitioning (or both). Either way, I think the changes here are still important to handle the potential `undefined` cases.

Demo (env is set to prod):

https://github.com/Recidiviz/justice-counts/assets/59492998/3fa5e09b-c0bc-4c7e-9ab0-0345c0cbdb7c

Please feel free to sanity check by running locally and pointing to production in your env file.

## Related issues

Closes #944

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/utils/formatting.ts
+++ b/agency-dashboard/src/utils/formatting.ts
@@ -29,7 +29,7 @@ export const getDatapointYear = (datapoint: Datapoint) => {
 };
 
 export const renderPercentText = (
-  val: number | string | null,
+  val: number | string | null | undefined,
   maxValue: number,
   isCurrency = false
 ) => {
--- a/common/components/DataViz/CategoryOverviewBreakdown.tsx
+++ b/common/components/DataViz/CategoryOverviewBreakdown.tsx
@@ -36,13 +36,16 @@ export const CategoryOverviewBreakdown: FunctionComponent<
   const { month, year } = splitUtcString(String(data.start_date));
   const displayDate = `${month} ${year}`;
   const totalDimensionValues = dimensions.reduce((acc, dim) => {
-    const sum = acc + Number(data[dim].value);
-    return sum;
+    if (data[dim]?.value) {
+      const sum = acc + Number(data[dim].value);
+      return sum;
+    }
+    return acc;
   }, 0);
   /** Dimensions sorted in descending order based on value */
   const sortedDimensions = dimensions.sort(
     (dimA: string, dimB: string) =>
-      Number(data[dimB].value) - Number(data[dimA].value)
+      Number(data[dimB]?.value) - Number(data[dimA]?.value)
   );
 
   return (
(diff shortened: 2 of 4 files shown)


lilidworkin (approved):
Thank you @mxosman ! Great detailed PR description. I'm sorry you sank a few hours into this and are still stumped by it -- it truly does sound gnarly! I feel good about this as a solution and I think we can definitely punt on anything further for now.

mxosman:
> Thank you @mxosman ! Great detailed PR description. I'm sorry you sank a few hours into this and are still stumped by it -- it truly does sound gnarly! I feel good about this as a solution and I think we can definitely punt on anything further for now.

Thank you, Lili! It kept feeling like I was getting closer and closer - haha, I wasn't ready to let up - I just might need to look at it again w/ fresh eyes. It's not a huge issue because the very next render after the one where things are undefined, things are perfectly sync'd up again, and esp. now since we're handling undefined, it shouldn't crash. But, I hope to find the root cause at some point.

Issue #948, opened 2023-09-15
[Agency Dashboards] Playtesting Followup: Design Nits


mxosman:
* ~Make line chart dots bigger on hover~
* ~Remove Breakdowns line chart area for metrics that have no breakdowns to display (ensure consistent behavior for breakdowns that are disabled, metrics with no breakdowns, and breakdowns with no values)~
* ~Minimum bar height in charts for low values in huge range~ (no longer pursuing)
* ~Right aligned y-axis chart labels~

PR #973, opened 2023-09-26
[Agency Dashboards] Playtesting Followup - Remove Line Chart when there are no breakdowns/no breakdown values

mxosman:
## Description of the change

> Description here

## Related issues

Contributes to #948

### Development

### Code review


common/hooks/useLineChart.ts
```diff
@@ -41,16 +43,30 @@ export const useLineChart = ({
...
+    const disaggregationsByDisplayName = groupBy(
+      metric.disaggregations,
+      (disaggregation) => disaggregation.display_name
+    );
+
     if (datapointsByMetric) {
       /**
-       * Gets the first set of disaggregated datapoints.
+       * Gets the first set of disaggregated datapoints (if atleast one dimension is enabled).
        * NOTE: This assumes there's just one breakdown per metric. We will need to adjust this
```
lilidworkin, line 54: Does this mean metrics with race, sex, etc? Feels like it would be good to have a task for this so we don't forget. Though I guess now that I think about it the design will have to be pretty different in this case since there's no space to display multiple breakdowns anyway ...

lilidworkin (approved):
Great! Let's add some inline comments clarifying when chart is shown vs. not. Up to you where to put them!

lilidworkin:
@mxosman 

> Metrics that have breakdowns but all of the breakdowns ~have no value or~ are unconfigured/disabled (current behavior is that it shows a blank empty chart and no breakdowns in the legend)

Does the strikethrough mean that if all breakdowns have no value, we do show an empty chart? Will it say Not Recorded in that case? I think that's probably the right behavior if so

common/hooks/useLineChart.ts
mxosman: Yeah, this was never implemented with multiple breakdowns in mind, so there will need to be a general refactor to accommodate that depending on what that will look like - which will affect all of these areas I left comments on. I introduced these comments in https://github.com/Recidiviz/justice-counts/pull/918/files#r1318772422 keeping in mind that we'll need to do a pass to render multiple breakdowns in some way, but I'll add a task just to be safe and consistent!

lilidworkin (approved):
All makes sense, thanks!

mxosman:
Thank you for all of the quick reviews!

PR #1001, opened 2023-10-17
[Publisher] Help Center: Self-hosted videos for Bulk Upload Guide

mxosman:
## Description of the change

Instead of using embedded loom videos and worrying about adjusting share permissions, we can self-host the mp4 videos and load them directly. There's no concern from my end at the moment about file size and performance (given that there's only a small amount of videos we're hosting (2 so far), and their respective sizes are around or less than 3MB (both videos are <2.3MB)). It should be a no-op change functionally (just no longer using Loom embedded videos).

### Development

### Code review


publisher/src/components/HelpCenter/Guides/DashboardsGuide.tsx
```diff
@@ -26,7 +26,11 @@ export const DashboardsGuide = () => (
         All data shown within the Justice Counts Agency Dashboards are
         specifically published by each agency. For more information on the
         metrics shown, see the{" "}
-        <a href="https://justicecounts.csgjusticecenter.org/metrics/justice-counts-metrics/">
+        <a
+          href="https://justicecounts.csgjusticecenter.org/metrics/justice-counts-metrics/"
+          target="_blank"
+          rel="noreferrer noopener"
```
mxosman, line 32: Sneaking this in ahead of playtesting - so the JC website opens up in a new tab instead of the existing one.

publisher/src/react-app-env.d.ts
```diff
@@ -16,3 +16,9 @@
 // =============================================================================
 
 /// <reference types="react-scripts" />
+
+// Required for TS to recognize .mp4 files (Reference: https://github.com/facebook/create-react-app/issues/6822)
```
lilidworkin, line 20: 😮

lilidworkin (approved):
Great call with this! Totally agreed it's the way to go.

PR #1026, opened 2023-10-25
[Agency Dashboards] Refactor of `fillTimeGapsBetweenDatapoints`

### LINKED ISSUE #947: [Agency Dashboards] Update behavior of fillTimeGapsBetweenDatapoints to go based off of a metric's current frequency

mxosman:
## Description of the change

Refactors the `fillTimeGapsBetweenDatapoints` in a way that better accommodates different metric settings and makes more visual sense. Prior to this change, the original strategy of this function (greatly simplified in this explanation) was calculating days in seconds between dates and determining whether or not to create a time gap - which created some inconsistent behavior especially when trying to accommodate different metric frequencies (and edge-cases like having a FY record published, but the metric frequency is set to CY, for example). Since this original function cannot be easily updated, this proposal refactors the strategy and instead of calculating intervals between dates, it creates the time-ranges specified by the viewing window the user filters to, determines which periods do not have existing datapoints, and creates gap datapoints for those time-ranges (for annual metrics based on the metric frequency's starting month).

Let's take an agency (LDPSC)'s Funding metric for example (this was the one brought up during playtesting):

Note: The Funding metric's frequency is Annual - Fiscal Year. 

Here's how their bar chart looks in "All" view:
Screenshot 2023-10-26 at 9 36 23 AM

Here's how it looks in "Recent" view:
Screenshot 2023-10-26 at 9 36 32 AM
The duplication of the 2021 bar is because their frequency is set to FY, but the original function filled in time gap datapoints for calendar year. So one 2021 bar is Jan - Dec 2021, and the other one is for their published record July 2021 - June 2022.

---

After the refactor, here's how the chart looks in "All" view:
Screenshot 2023-10-26 at 9 54 13 AM
This starts from the earliest published datapoint to the current year. It feels like making the end date in "All" view the current year makes sense to me visually - instead of only showing the published data. For me, it felt nice to have some reference to the current time period - but, let me know if this would feel visually annoying since this current time period might not have data available yet.

Here's how it looks in "Recent" view:
Screenshot 2023-10-26 at 9 37 09 AM
This feels like it makes SO much more sense to me. Part of the confusing parts I was running into when debugging was not realizing that this view is meant to include 2023 - for some reason when I went to recent view (5 years ago) and only saw 4 bars I kept thinking something was wrong.

Here, the duplicate 2021 is taken care of and the gap datapoints align with the metric's annual FY frequency configuration.

This is just one example, I tested with multiple agencies including one where I was publishing records and adjusting metric frequencies and it feels like the results are visually consistent. I tested this in Publisher's Explore Data as well and there doesn't appear to be any weird side-effects. 

If you'd like to play around with it, here's a playtesting link: https://mahmoudtest---agency-dashboard-web-b47yvyxs3q-uc.a.run.app/
It might be better to run this locally so you can control whether or not the env points to prod or staging.

## Related issues

Closes #947 

### Development

### Code review


common/components/DataViz/utils.test.ts
```diff
@@ -1954,16 +1952,6 @@ const testDatapoints5Transformed: Datapoint[] = [
...
-  {
-    start_date: "Sat, 01 Oct 2022 00:00:00 GMT",
-    end_date: "Tue, 01 Nov 2022 00:00:00 GMT",
-    dataVizMissingData: 0.3333333333333333,
-    frequency: "MONTHLY",
-    Pretrial: 0,
-    Sentenced: 0,
-    "Transfer or Hold": 0,
-    Unknown: 0,
-  },
```
mxosman, line 1966: I'm still trying to figure out why in the testing environment, it skips this month - everywhere I test this, even console.logging in the test environment shows deep equality between the provided datapoints (which include this month) and the transformed datapoints (which include this month) - but somehow removing it entirely passes the test. Still looking into it - good news is the function doesn't actually skip this month at all - but for some reason the deep equality check fails despite a print out of perfectly matching input datapoints and transformed datapoints.

mxosman:
Still doing more testing, so I'm letting this bake for a bit to make sure it doesn't introduce anything weird.

publisher/src/components/DataViz/utils.test.ts
```diff
@@ -1,2067 +0,0 @@
-// Recidiviz - a data platform for criminal justice reform
```
mxosman, line 1: The deleted files are duplicates - I made sure none of the util functions were being used anywhere and thankfully none of them were being used. We just must've forgotten to delete these after moving things to the `common` repo!

common/components/DataViz/utils.test.ts
mxosman: Figured out what was going on - the test has explicitly set the current date to be Aug 2022, so the current logic's time-range ends at that current date and doesn't quite handle datapoints past the current date. I updated the logic to accommodate and choose a the most-future datapoint date if there is one, otherwise the end the time-range at the current year. Does this make sense? (If not, it's because this fried my brain - haha - happy to clarify further!)

common/components/DataViz/utils.test.ts
```diff
@@ -1344,46 +1344,6 @@ const testDatapoints5: Datapoint[] = [
...
-  {
-    start_date: "Fri, 01 Dec 2017 00:00:00 GMT",
-    end_date: "Mon, 01 Jan 2018 00:00:00 GMT",
-    dataVizMissingData: 0.3333333333333333,
-    frequency: "MONTHLY",
-    Pretrial: 0,
-    Sentenced: 0,
-    "Transfer or Hold": 0,
-    Unknown: 0,
-  },
```
mxosman, line 1386: These mock datapoints are safe to remove because they fall outside of the viewing window (which is set to 5 years ago - in this test env with Aug 2022 as the current date, this means the viewing window is Jan 2018 - 2022) and will not be included in the transformed datapoints.

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
...
+  // If `monthsAgo` is 0, user intends to see all datapoints, the length of the array will be the months inbetween the first datapoint date and the current date
+  // If `monthsAgo` is non-zero, user intends to see a specific time-range, the length of the array will be the months inbetween the current date and the # of `monthsAgo`
+  const monthlyTimeRangeFromCurrentDateOrFutureDatapointDate =
+    hasFutureDatapointDate
+      ? lastDatapointStartMonth -
+        firstDatapointStartMonth +
+        12 * (lastDatapointStartYear - firstDatapointStartYear)
+      : currentMonth -
+        firstDatapointStartMonth +
+        12 * (currentYear - firstDatapointStartYear);
```
mxosman, line 302: I will clean this up - this is super gnarly...

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
...
+      length: isAnnual ? annualTimeRangeLength : monthlyTimeRangeLength,
+    },
+    (_, i) => {
+      // For monthly metrics, decrement the date so we start with the current date and subtract months on each loop.
+      const decrementedDate = hasFutureDatapointDate
+        ? new Date(
+            new Date(lastDatapointDate).setUTCMonth(
+              new Date(lastDatapointDate).getUTCMonth() - i
+            )
+          )
```
mxosman, line 326: I will clean this up too!

publisher/src/components/DataViz/utils.test.ts
lilidworkin: Oh great, thanks for cleaning up!

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
  */
 export const fillTimeGapsBetweenDatapoints = (
   data: Datapoint[],
-  monthsAgo: number
+  monthsAgo: number,
+  metricFrequency?: ReportFrequency,
+  startingMonth?: number // For annual metrics, represents the starting month of the recording period
```
lilidworkin, line 238: Awesome inline comments through this PR!

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
...
+        firstDatapointStartMonth +
+        12 * (currentYear - firstDatapointStartYear);
+  const monthlyTimeRangeLength =
+    monthsAgo === 0
+      ? monthlyTimeRangeFromCurrentDateOrFutureDatapointDate
+      : monthsAgo;
+
+  /**
+   * Create an array of dates that represent our viewing time-frame (i.e. the range of time the user filtered to - e.g. 5 years ago)
+   * If `monthsAgo` is 60 months and the metric frequency is annual, the array will look like this: [<Current Date (2023)>, <2022 Date>, <2021 Date>, <2020 Date>, <2021 Date>]
```
lilidworkin, line 310: super nit, should last year in this comment be 2019?

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
...
+  const lastDatapointStartMonth = lastDatapointDate.getUTCMonth();
+  const lastDatapointStartYear = lastDatapointDate.getUTCFullYear();
+  const hasFutureDatapointDate = lastDatapointDate > currentDate;
+  /**
+   * The reference point helps us determine the beginning of the filtered time window if `monthsAgo`
+   * is provided (not 0). If `monthsAgo` is 0, we will show the user all of the datapoints and no
+   * longer depend on these reference points. If the `monthsAgo` is a non-zero number (for example, 60 months),
+   * then we set the reference point to 5 years ago from the current date.
+   */
+  const referencePointMonth = currentMonth - monthsAgo;
```
lilidworkin, line 274: nit, but I don't love the name referencePoint, really what we're computing here is the beginning of the time window, so I might just call it something that, i.e. timeRangeStartMonth

common/components/DataViz/utils.ts
lilidworkin: I'm a little confused about how someone could have future data? I guess we don't prevent you from entering in data for a month in the future? I'm not sure we need to go out of our way to handle this edge case. If we do want to (or if I'm misunderstanding and this is more important/common) then maybe one way to clean this up is, instead of a boolean hasFutureDatapointDate, we create a timeRangeEndDate and set it to the maximum of lastDatapointStartMonth and currentMonth? Would something like that work?

common/components/DataViz/utils.ts
lilidworkin: Similarly here, I think creating a timeRangeEndDate = max(lastDatapointDate, currentDate) and then using that might let us avoid the ternary conditional! But I might be misunderstanding

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
...
+      : monthsAgo;
+
+  /**
+   * Create an array of dates that represent our viewing time-frame (i.e. the range of time the user filtered to - e.g. 5 years ago)
+   * If `monthsAgo` is 60 months and the metric frequency is annual, the array will look like this: [<Current Date (2023)>, <2022 Date>, <2021 Date>, <2020 Date>, <2021 Date>]
+   * If `monthsAgo` is 0 months and the metric frequency is annual, the array will look like this: [<Current Date>, ...<Dates (years) inbetween> ,<First Datapoint Date>]
+   * If `monthsAgo` is 12 months and the metric frequency is monthly, the array will look like this: [<Current Date>, ...<Dates (months) inbetween> , <Date 12 months ago>]
+   * If `monthsAgo` is 0 months and the metric frequency is monthly, the array will look like this: [<Current Date>, ...<Dates (months) inbetween> ,<First Datapoint Date>]
+   */
+  const timeFrameArray = Array.from(
```
lilidworkin, line 315: This logic/syntax is very cool!

common/components/DataViz/utils.ts
```diff
@@ -231,55 +233,173 @@ export const filterNullDatapoints = (data: Datapoint[]) => {
...
+    }
+    return true;
+  });
+
+  /**
+   * Create an array of gap datapoints by looping through the `timeFrameArray` and first filtering out timeframes
+   * that have existing datapoints (because they won't need a gap datapoint), then, creating a gap datapoint
+   * object for each time-frame that doesn't have an existing datapoint.
+   */
+  const gapDatapoints = timeFrameArray
```
lilidworkin, line 357: I don't think this really matters, but this is an O(n*2) loop (for each time range, we loop through all datapoints to determine if this time range is represented). To speed it up, we could loop through all datapoints once and create a set containing all the time ranges that are represented in the datapoints. Then we could do a "[set difference operation](https://2ality.com/2015/01/es6-set-operations.html)", basically timeFrameArray - timeRangesRepresented, to find the time ranges that aren't represented. This may be more trouble than it's worth, so only do it if it makes sense and sounds like fun, haha.

lilidworkin (commented):
@mxosman this looks fantastic! I didn't actually test it out, but just reviewed the code carefully. If you tested on several agencies (especially the ones that CSG[ is using to demo](https://docs.google.com/document/d/1AMq36vUvUIj_yEjNtjLbGFXjU8ttNk8u3MSqLRnHPoE/edit#heading=h.lg3s48ernkp8)) then I think that's more than sufficient. As more agencies get dashboards, we'll continue testing and adjusting!

common/components/DataViz/utils.ts
mxosman: Ugh - thank you so much! I was trying all kinds of different names that satisfied both this being the beginning time range or the current month (if `monthsAgo` isn't provided) - but I love your suggestion.

common/components/DataViz/utils.ts
mxosman: Haha - that's a great call! I think with the test failing and the and seeing a 2024 record in one of the test agencies made me hone in on solving that case without really considering - wait, who has future data? Time travelers? Thanks for bringing me back to reality here. That simplifies things quite a bit! [Great suggestion if we were to keep this though!]

common/components/DataViz/utils.ts
mxosman: Haha - totally! Thanks for catching!

common/components/DataViz/utils.ts
mxosman: Since we're not handling the unnecessary case of future records - we can totally get rid of this ternary situation all together! But, that's such a great solution!

common/components/DataViz/utils.ts
mxosman: Ooo - great find & call out! Will explore this! Thank you so much for this!

mxosman:
> @mxosman this looks fantastic! I didn't actually test it out, but just reviewed the code carefully. If you tested on several agencies (especially the ones that CSG[ is using to demo](https://docs.google.com/document/d/1AMq36vUvUIj_yEjNtjLbGFXjU8ttNk8u3MSqLRnHPoE/edit#heading=h.lg3s48ernkp8)) then I think that's more than sufficient. As more agencies get dashboards, we'll continue testing and adjusting!

Thank you so so much, Lili! That's a great call, I definitely agree - I'll make some updates and test again on those agencies in the doc & sanity check Publisher.

common/components/DataViz/utils.ts
mxosman: Such a wonderful suggestion - feels much better than what I had before. [e35aff5](https://github.com/Recidiviz/justice-counts/pull/1026/commits/e35aff59c835194dc55676b3afbee61ef629bfe3) - comments updated in [61101d3](https://github.com/Recidiviz/justice-counts/pull/1026/commits/61101d32d40e3cd6d6bb342029f2f8aebb2fb11a)

mxosman:
I'm still testing some staging agencies and different scenarios in Publisher - but prod agencies (there's only one with a live dashboard) looks good!

lilidworkin (approved):
Awesome @mxosman , looks great!!

Just a clarification though, you said there is only one prod agency with a dashboard, I thought based on Nichelle's pull that there are a handful?

https://docs.google.com/spreadsheets/d/1TcskbQat7a3OA7X-KRra74qBcXZMZVg5nOfdRFa9-64/edit#gid=527493643

mxosman:
> Awesome @mxosman , looks great!!
> 
> Just a clarification though, you said there is only one prod agency with a dashboard, I thought based on Nichelle's pull that there are a handful?
> 
> https://docs.google.com/spreadsheets/d/1TcskbQat7a3OA7X-KRra74qBcXZMZVg5nOfdRFa9-64/edit#gid=527493643

Ah, yes! All of those have staging links, so I didn't assume they were prod agencies - I'll go through each one and update the links for the ones that are prod and test them as well.

lilidworkin:
Oh that's weird -- I think that's a bug -- cc @nichelle-hall , shouldn't those be prod links?

mxosman:
Thanks for pointing this out, Lili! I think it is a bug - happy to update the spreadsheet for now (once you've had a chance to take a look Nichelle). In the meantime, I'll test all of the other live prod ones before merging!

mxosman:
Good news! No issues found with the gap datapoints in the 5 prod agencies I tested.

lilidworkin:
Yay! Thanks for testing so thoroughly 😊

mxosman:
> Yay! Thanks for testing so thoroughly 😊

Thank YOU for your eyes, brain, and sanity check!

PR #1047, opened 2023-11-14
[Publisher] UI adjustment for creating fiscal year records

mxosman:
## Description of the change

A follow up to an issue Stephanie ran into when trying to create a new fiscal year record for 7/2022 - 7/2023. The attempt raised an error that a record already exists with the specified date-ranges despite there being no existing record with the aforementioned date-ranges. 

We learned that the year input when trying to create a FY record is processed as the ending year and not the starting year - this adjustment was made to accommodate bulk upload but the Create Report UI was not updated to reflect this. So, Stephanie's attempt at creating a new fiscal year record for 7/2022 - 7/2023 (by selecting 2022 as the starting year in the UI) was processed as trying to create a FY record for 7/2021 - 7/2022 (which already exists, hence the error thrown).

The proposed solution is the update the Create Report UI copy to read `When should this record end?` when trying to create a FY record (as opposed to `When should this record start?`) - and update the years in the `The record will be created for X - X` box to reflect the appropriate years.

Screenshot 2023-11-14 at 12 06 23 PM

https://github.com/Recidiviz/justice-counts/assets/59492998/db919f84-2e7a-4c8e-9dda-00e058750e80

My only small concern with this is that it maybe difficult to see and tell esp. if the flow of monthly and annual CY is starting year. It might not be a big deal at all since the `The record will be created for X - X` box will help clarify this and most records (I think) are created automatically on a recurring basis.

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/CreateReport.tsx
+++ b/publisher/src/components/Reports/CreateReport.tsx
@@ -177,6 +177,24 @@ const CreateReport = () => {
     highlight: yearValue === year,
   }));
 
+  const isFiscalYear =
+    createReportFormValues.frequency === "ANNUAL" &&
+    createReportFormValues.annualStartMonth !== 1;
+
+  const getDateRangeYear = () => {
+    if (isRecurring) {
+      if (isFiscalYear) {
+        return new Date(Date.now()).getFullYear() - 1;
+      }
+      return new Date(Date.now()).getFullYear();
+    }
+
+    if (isFiscalYear) {
+      return year - 1;
+    }
+    return year;
+  };
+
   if (userStore.isUserReadOnly(agencyId))
     return <Navigate to={`/agency/${agencyId}/${REPORTS_LOWERCASE}`} />;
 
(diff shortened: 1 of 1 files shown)


nichelle-hall (approved):
This looks great Mahmoud!

mxosman:
TY for the quick review, Nichelle!

morden35:
> My only small concern with this is that it maybe difficult to see and tell esp. if the flow of monthly and annual CY is starting year. It might not be a big deal at all since the The record will be created for X - X box will help clarify this and most records (I think) are created automatically on a recurring basis.

I agree that I could see some confusion happening. I remember someone from CSG mentioning that if users report 1 metric using Fiscal Year, they most likely report all metrics using Fiscal Year (rather than a combo of CY and FY depending on the metric). So maybe this helps our case?

mxosman:
> > My only small concern with this is that it maybe difficult to see and tell esp. if the flow of monthly and annual CY is starting year. It might not be a big deal at all since the The record will be created for X - X box will help clarify this and most records (I think) are created automatically on a recurring basis.
> 
> I agree that I could see some confusion happening. I remember someone from CSG mentioning that if users report 1 metric using Fiscal Year, they most likely report all metrics using Fiscal Year (rather than a combo of CY and FY depending on the metric). So maybe this helps our case?

Ah - that's a really good point and makes so much sense!

mxosman:
Ah - nice! Thank you for that extra reassurance, Lili!

PR #1057, opened 2023-11-15
[Publisher] Admin Panel: Searchable List Component

### LINKED ISSUE #1049: [Publisher] Admin Panel: Searchable List Component

mxosman:
## Description of the change

I was exploring an alternative to a dropdown for displaying a big list of elements that allowed you to: view more list items than a dropdown could reasonable show, was searchable, and gave you the ability to select multiple elements. So, I created an `InteractiveSearchList` component that I'd like to try to accomplish this. It takes in a list, allows you to search that list by object keys you specify, and select multiple items within the list (the colors denote an adding action or deleting action). The idea is whenever we need to add or remove users (for example), we can use this component to render the list, search it and/or select users from the list (with visual feedback - if you're in the "add user" mode, the selections will be green -- if you're in the "delete user" mode, the selections will be red).

https://github.com/Recidiviz/justice-counts/assets/59492998/82cdb3b6-8f46-41be-b636-47aa5127f762

(Currently the searchable keys in this demo are set to "name" and "email")

## Related issues

Closes #1049 

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AdminPanel.styles.tsx
+++ b/publisher/src/components/AdminPanel/AdminPanel.styles.tsx
@@ -22,7 +22,10 @@ import {
 } from "@justice-counts/common/components/GlobalStyles";
 import styled, { css } from "styled-components/macro";
 
-import { SearchableListBoxAction, SearchableListBoxActions } from "./types";
+import {
+  InteractiveSearchListAction,
+  InteractiveSearchListActions,
+} from "./types";
 
 /** General */
 
@@ -261,7 +264,7 @@ export const ChipContainer = styled.div<{
   halfMaxHeight?: boolean;
   fitContentHeight?: boolean;
   noBorder?: boolean;
-  boxActionType?: SearchableListBoxAction;
+  boxActionType?: InteractiveSearchListAction;
   hoverable?: boolean;
 }>`
   ${typography.sizeCSS.small}
@@ -279,10 +282,10 @@ export const ChipContainer = styled.div<{
   align-content: baseline;
   border: ${({ noBorder, boxActionType }) => {
     if (noBorder) return `none`;
-    if (boxActionType === SearchableListBoxActions.DELETE) {
+    if (boxActionType === InteractiveSearchListActions.DELETE) {
       return `1px solid ${palette.solid.red};`;
     }
-    if (boxActionType === SearchableListBoxActions.ADD) {
+    if (boxActionType === InteractiveSearchListActions.ADD) {
       return `1px solid ${palette.solid.green};`;
     }
     return `1px solid ${palette.highlight.grey5}`;
@@ -292,10 +295,10 @@ export const ChipContainer = styled.div<{
   overflow-y: auto;
 
   ${({ boxActionType }) => {
-    if (boxActionType === SearchableListBoxActions.DELETE) {
+    if (boxActionType === InteractiveSearchListActions.DELETE) {
       return `box-shadow: 1px 1px 2px ${palette.highlight.red};`;
     }
-    if (boxActionType === SearchableListBoxActions.ADD) {
+    if (boxActionType === InteractiveSearchListActions.ADD) {
       return `box-shadow: 2px 2px 5px ${palette.highlight.green};`;
     }
   }}
--- a/publisher/src/components/AdminPanel/InteractiveSearchList.tsx
+++ b/publisher/src/components/AdminPanel/InteractiveSearchList.tsx
@@ -1,0 +1,141 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2023 Recidiviz, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import { groupBy } from "lodash";
+import React, { useEffect, useState } from "react";
+
+import AdminPanelStore from "../../stores/AdminPanelStore";
+import * as Styled from "./AdminPanel.styles";
+import {
+  InteractiveSearchListAction,
+  InteractiveSearchListActions,
+  InteractiveSearchListProps,
+  SearchableListItem,
+  SearchableListItemKey,
+} from "./types";
+
+export const InteractiveSearchList = ({
+  list,
+  buttons,
+  selections,
+  updateSelections,
+  boxActionType,
+  metadata,
+  isActiveBox = true,
+}: InteractiveSearchListProps) => {
+  const selectionsByName = groupBy(selections, (selection) => selection.name);
+  const searchByKeys: SearchableListItemKey[] = ["name", "email"];
+
+  const [filteredList, setFilteredList] = useState<SearchableListItem[]>([]);
+  const [searchInputValue, setSearchInputValue] = useState("");
+
+  const getChipColor = (actionType?: InteractiveSearchListAction) => {
+    if (!actionType) return;
+    if (actionType === InteractiveSearchListActions.DELETE) return "red";
+    if (actionType === InteractiveSearchListActions.ADD) return "green";
+  };
+
+  useEffect(() => {
+    setFilteredList(
+      AdminPanelStore.searchList(list, searchInputValue, searchByKeys)
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [list]);
+
+  return (
+    <>
+      <Styled.InputLabelWrapper>
+        {/* Title */}
+        {metadata?.title && (
+          <Styled.ModalTitle>{metadata.title}</Styled.ModalTitle>
+        )}
+
+        {/* List of Searchable Items */}
+        <Styled.ChipContainer
+          boxActionType={isActiveBox ? boxActionType : undefined}
+        >
+          {filteredList.length > 0 &&
+            filteredList.map((listItem) => (
+              <Styled.Chip
+                key={listItem.id}
+                onClick={() => {
+                  if (isActiveBox && boxActionType) {
+                    updateSelections({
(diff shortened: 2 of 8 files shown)


publisher/src/components/AdminPanel/AdminPanel.styles.tsx
```diff
@@ -22,7 +22,10 @@ import {
 } from "@justice-counts/common/components/GlobalStyles";
 import styled, { css } from "styled-components/macro";
 
-import { SearchableListBoxAction, SearchableListBoxActions } from "./types";
+import {
```
mxosman, line 25: Only variable name changes in this file.

publisher/src/components/AdminPanel/types.ts
```diff
@@ -85,26 +85,26 @@ export type UserRole = (typeof userRoles)[number];
...
-  email?: string,
-  role?: UserRole
-) => void;
+export type InteractiveSearchListUpdateSelections = (selection: {
+  id: string | number;
+  name: string;
+  action?: InteractiveSearchListAction;
+  email?: string;
+  role?: UserRole;
+}) => void;
```
mxosman, line 102: Only variable name changes in this file except for this chunk - basically typing this function to take in an object instead of 5 params.

publisher/src/components/AdminPanel/InteractiveSearchList.tsx
```diff
@@ -0,0 +1,141 @@
...
+  boxActionType,
+  metadata,
+  isActiveBox = true,
+}: InteractiveSearchListProps) => {
+  const selectionsByName = groupBy(selections, (selection) => selection.name);
+
+  const [filteredList, setFilteredList] = useState<SearchableListItem[]>([]);
+  const [searchInputValue, setSearchInputValue] = useState("");
+
+  const getChipColor = (actionType?: InteractiveSearchListAction) => {
```
nichelle-hall, line 45: Could we also have some other text / banner that shows we are in delete or add mode? I can see myself being confused. And maybe also a confirmation modal before deleting users?

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -113,6 +114,26 @@ class AdminPanelStore {
...
   }
+
+  /**
+   * Returns a filtered array of objects for a given string within specified keys.
+   * @param list - The array of objects to search.
+   * @param searchInput - The string to search for.
+   * @param searchByKeys - The keys to search within each object.
+   * @returns An array of objects that match the search criteria.
+   */
+  static searchList<T extends SearchableEntity>(
```
morden35, line 125: So this function allows us to search a `SearchableEntity` by some `searchInput`, right? I wonder if this is the most efficient way to do so. I know in python, it's faster to sort/search certain data structures over others (dictionaries and sets are faster than lists, for example). Are there equivalent data structures / comparisons in javascript? I also know there are a million different search algorithms that may be useful for faster / more efficient searching

mxosman: Kind of! It's doing both searching and filtering of a list. I kept going back and forth on naming this `searchList` or `filterList` (still up in the air).

So this is taking in a list (of Agencies or Users for example), a search value, and a list of keys to search within each agency or user object in the list. 

For example:
```
  const agencies = [
    {
      id: 7,
      name: "Jails Agency",
      ...
    },
    {
      id: 8,
      name: "Defense Agency"
      ...
    }
  ]

  searchList(agencies, "def", ["id", "name"]);
```
If you were to run this, it will loop once through the `agencies` list, and look at each agency object's `id` & `name` property and do a regex test of the search value (`"def"`) against the value of those properties. If no match, that agency gets filtered out of the list, then it checks the next agency object. In this case, it will return a list with one dictionary which will be the "Defense Agency" since it passed the regex test when it was looking up the agency's `name` property.


As far as efficiency - great call out! JavaScript has most, if not all the data structures available natively in Python. I think if this were only doing a look up, then I totally see the argument for creating a dictionary/set structure and get that O(1) lookup time complexity. But, it would be very challenging to come up with a dictionary/set structure to allow for this kind of operation and flexibility in checking multiple values of an object's properties (I feel like we'd still have to iterate through each one regardless to do a regex test). If we could come up with a structure, there's also the additional steps of converting a list to a dictionary/set and then converting it back to a list (which is the final structure the FE will need to render) to consider as well. I think to maintain this functionality (& flexibility), one filter loop each time returning a ready-to-render list feels more efficient to me? Totally open to any ideas though!

publisher/src/stores/AdminPanelStore.test.tsx
```diff
@@ -318,3 +318,70 @@ test("sortListByName sorts a list of users by name in explicit descending order"
...
+  filteredListOfUsers = AdminPanelStore.searchList(
+    adminPanelStore.users,
+    "er",
+    ["name"]
+  );
+  expect(filteredListOfUsers.length).toBe(2);
+  expect(filteredListOfUsers[0].name).toBe("Liz Erd");
+  expect(filteredListOfUsers[1].name).toBe("Percy Vere");
+
+  /** Search value: "1" | Search-by keys: "name" & "id" */
```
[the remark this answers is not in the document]
mxosman: Great question! I posted an example in the other comment - let me know if that makes sense - happy to clarify further if it's still not clear!

publisher/src/stores/AdminPanelStore.ts
lilidworkin: Ooh great conversation/discussion! You guys know how much I love talking about time complexity, haha.

It's a great callout -- like @mxosman says, this search is O(n), meaning we loop through the whole list of agencies each time we do a search. And since (I think) we do a search on every keystroke, that means several O(n) operations in a row. So it's a good thing to notice! 

If we just wanted to _lookup_ an agency by name or id, @morden35 is right that a set or dictionary would be faster -- that would get us O(1) lookup. But in this case, we're doing something different -- we're saying "find all agencies that have "xyz" in their name -- so a set or dictionary doesn't work.

There are potentially fancier things we can do -- like we could cache common search queries, or we can take advantage of the fact that if you type "xy" and we do a search for that and then you type "z" after, we don't really have to perform a brand new search for "xyz", we can instead just filter the search results we already have from the search for "xy". But for this use case, I don't think it's worth it! Since we don't have that many agencies, the difference won't be noticeable.

Let me know if this makes sense!

lilidworkin: So all this to say I actually think your solution here @mxosman is great, and this is a really nice/elegant function!

publisher/src/components/AdminPanel/InteractiveSearchList.tsx
```diff
@@ -0,0 +1,141 @@
...
+        id: listItem.id,
+        name: listItem.name,
+        action: boxActionType,
+        email: listItem.email,
+        role: listItem.role,
+      });
+    }
+  };
+
+  useEffect(() => {
```
lilidworkin, line 62: Out of curiosity -- because I'm just trying to get more comfortable with this stuff -- what's the difference between having this useEffect, which (I think) initializes the filteredList, vs. initializing it above, e.g. `const [filteredList, setFilteredList] = useState<SearchableListItem[]>(list);`

publisher/src/components/AdminPanel/InteractiveSearchList.tsx
```diff
@@ -0,0 +1,141 @@
...
+              AdminPanelStore.searchList(list, e.target.value, searchByKeys)
+            );
+          }}
+        />
+        <label htmlFor="search-input">
+          {metadata?.searchBoxLabel}
+          <Styled.LabelButton
+            onClick={() => {
+              setSearchInputValue("");
+              setFilteredList(
```
lilidworkin, line 130: What's the difference between this and `setFilteredList(list)`?

lilidworkin (approved):
Looks great @mxosman ! I love this component! Left a few questions inline but they are just for my own understanding.

publisher/src/components/AdminPanel/InteractiveSearchList.tsx
mxosman: Ah - this is a great catch! Your instincts are spot on. I think I was licking my wounds from the weird scenario in #925 where I needed to use a `useEffect` because the parent component was also re-rendering a new dropdown list after a user clicked on a menu item - and overfit that solution here. Since there is no parent component that will feed this component a new list, I think we can safely initialize it like you said `const [filteredList, setFilteredList] = useState<SearchableListItem[]>(list);`!  #925 was a special case and I was trying to (over)protect myself here.

Both work exactly the same! Though, if it is initialized like you mentioned, and `list` were to change for some reason from the parent component (say, there's logic where I click on something that triggers the parent component to generate a new `list`), then the new list would not be picked up by this component - because the initialization happens once when the component mounts and never again for the lifecycle of the component (this is because `useState` retains its values between render cycles, so once the initialization happens, it's done until this component unmounts and remounts). [I'm not doing a great job explaining this b/c words are so tough - but I'll sync with you on this in our 1:1!]

I'll adjust it to initialize in the `useState` for now, but I may need to change it back to this when I implement the adding a new agency/user from different flows, because that will generate a new overall `list` and will cause out-of-sync issues if there isn't a `useEffect` listening for a change in `list`.

publisher/src/components/AdminPanel/InteractiveSearchList.tsx
mxosman: Ah - great catch! No difference at all - I'm just resetting it the long way. 😆 Will update!

publisher/src/stores/AdminPanelStore.ts
mxosman: Haha - it's always a great conversation! Thank you for the insight, Lili - I love the idea of potentially caching common queries, I think that would be a great optimization as our agencies list grows and we notice a slow down. I agree though that it's not worth it to optimize this early. 

FWIW, modern browsers are extremely optimized to handle these operations on big lists like this with virtually no perceptible slow down to a user. If our list of agencies grows past ~5,000 (maybe even up to ~10k - pure assumption based on old benchmark test I've seen in the past on big lists), we might start seeing this filter slow down a tiny bit? At which point, we can totally resolve by optimizing like you suggested above!

morden35: awesome, thank you both for the great explanations!

publisher/src/stores/AdminPanelStore.test.tsx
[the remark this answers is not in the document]
morden35: makes sense, thank you!

morden35 (approved):
lgtm! thank you for the clarifications!

mxosman:
Thank you all so so much for the reviews and feedback!

PR #1068, opened 2023-11-20
[Bug] Fix link from Help Center to Agency Dashboards for production users

### LINKED ISSUE #1067: [Bug] The dashboard guides in production link to the dashboard staging URL instead of prod URL

mxosman:
## Description of the change

I noticed a bug when I was playing around with the Help Center in production specifically in the link to Agency Dashboards. I was taken to the staging URL which resulted in an error - instead of the production URL. For some reason the piece of logic that creates the URL based on the environment isn't picking up on the right environment. I suspect it is because when this component is mounted, the `api.environment` is first `undefined` until it is set - and since these components aren't `observer`s they might not have been re-rendering when `api.environment` is set and have thus defaulted to the staging URL.

I tried to log the `api.environment` and deploy to playtesting with and without the `observer` and it logged `staging` for both - so my hunch might be off. But regardless, both components should be wrapped in an `observer` - so this change is necessary. I am curious to see when we deploy to prod if this fixes it. If not, I'll keep digging into this.

Screenshot 2023-11-20 at 11 28 27 AM

## Related issues

Closes #1067

### Development

### Code review

### DIFF
--- a/publisher/src/components/HelpCenter/LinkToPublisherDashboard.tsx
+++ b/publisher/src/components/HelpCenter/LinkToPublisherDashboard.tsx
@@ -16,14 +16,13 @@
 // =============================================================================
 
 import { slugify } from "@justice-counts/common/utils";
-import { observer } from "mobx-react-lite";
 import React, { PropsWithChildren } from "react";
 
 import { useStore } from "../../stores";
 
 export const LinkToPublisher: React.FC<
   PropsWithChildren & { publisherPath: string }
-> = observer(({ publisherPath, children }) => {
+> = ({ publisherPath, children }) => {
   const { userStore } = useStore();
   const agencyIdLocalStorage = localStorage.getItem("agencyId");
   const agencyId = agencyIdLocalStorage || userStore.getInitialAgencyId();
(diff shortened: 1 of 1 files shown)


publisher/src/components/HelpCenter/LinkToPublisherDashboard.tsx
```diff
@@ -16,13 +16,14 @@
...
 import { slugify } from "@justice-counts/common/utils";
+import { observer } from "mobx-react-lite";
 import React, { PropsWithChildren } from "react";
 
 import { useStore } from "../../stores";
 
 export const LinkToPublisher: React.FC<
   PropsWithChildren & { publisherPath: string }
-> = ({ publisherPath, children }) => {
+> = observer(({ publisherPath, children }) => {
```
mxosman, line 26: Wrapping these components in `observer` is the only change in this PR - the rest is prettier formatting.

morden35:
> I tried to log the api.environment and deploy to playtesting with and without the observer and it logged staging for both - so my hunch might be off. But regardless, both components should be wrapped in an observer - so this change is necessary. I am curious to see when we deploy to prod if this fixes it. If not, I'll keep digging into this.

Yeah since playtesting links are technically staging, this makes sense. So hopefully this change is all that is needed once deployed to prod!

morden35 (approved):
lgtm!

mxosman:
> > I tried to log the api.environment and deploy to playtesting with and without the observer and it logged staging for both - so my hunch might be off. But regardless, both components should be wrapped in an observer - so this change is necessary. I am curious to see when we deploy to prod if this fixes it. If not, I'll keep digging into this.
> 
> Yeah since playtesting links are technically staging, this makes sense. So hopefully this change is all that is needed once deployed to prod!

Thank you so much, Michelle! Yeah, I was hoping for the one without the `observer` to have logged twice - once `undefined` and the next one `staging`. 🤔 We shall find out next week!

PR #1083, opened 2023-12-04
[Justice Counts] Update user/agency last_visit

morden35:
## Description of the change

> Description here

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/Home.tsx
+++ b/publisher/src/components/Home/Home.tsx
@@ -31,8 +31,9 @@ export const Home = observer(() => {
   const { userStore, homeStore } = useStore();
   const { agencyId } = useParams() as { agencyId: string };
   const navigate = useNavigate();
+  const { name, isAgencySuperagency, userId } = userStore;
+  userStore.updateUserAgencyPageVisit(userId, agencyId);
 
-  const { name, isAgencySuperagency } = userStore;
   const {
     loading,
     systemSelectionOptions,
--- a/publisher/src/stores/UserStore.ts
+++ b/publisher/src/stores/UserStore.ts
@@ -33,6 +33,8 @@ class UserStore {
 
   userAgencies: UserAgency[] | undefined;
 
+  userId: string | undefined;
+
   userInfoLoaded: boolean;
 
   constructor(authStore: AuthStore, api: API) {
@@ -42,6 +44,7 @@ class UserStore {
     this.api = api;
     this.userAgencies = undefined;
     this.userInfoLoaded = false;
+    this.userId = undefined;
 
     when(
       () => api.isSessionInitialized,
@@ -228,9 +231,10 @@ class UserStore {
           email_verified: this.email_verified,
         },
       })) as Response;
-      const { agencies: userAgencies } = await response.json();
+      const { agencies: userAgencies, id: userId } = await response.json();
       runInAction(() => {
         this.userAgencies = userAgencies;
+        this.userId = userId;
       });
     } catch (error) {
       if (error instanceof Error) return error.message;
@@ -239,6 +243,21 @@ class UserStore {
       runInAction(() => {
         this.userInfoLoaded = true;
       });
+    }
+  }
+
+  async updateUserAgencyPageVisit(userId: string, agencyId: string) {
+    try {
+      const response = (await this.api.request({
+        path: `/api/${userId}/${agencyId}/page_visit`,
+        method: "PUT",
+      })) as Response;
+      if (response.status !== 200) {
+        throw new Error("There was an issue...");
+      }
+    } catch (error) {
+      if (error instanceof Error) return error.message;
+      return String(error);
     }
   }
 }


morden35:
@mxosman I am very unsure if I did this right. I decided to call the new endpoint from both the `Home.tsx` and `Router.tsx`. That way, `last_visit` will be updated upon login, or if a user switches agencies from any page (I think?). I'm unsure why the endpoint gets trigger twice for each of these. Open to any / all feedback! This is just a first pass.

mxosman:
>@mxosman I am very unsure if I did this right. I decided to call the new endpoint from both the Home.tsx and Router.tsx. That way, last_visit will be updated upon login, or if a user switches agencies from any page (I think?). I'm unsure why the endpoint gets trigger twice for each of these. Open to any / all feedback! This is just a first pass.

This is a tough one! You're likely getting duplicate calls because it's being called once in `Router.tsx` which is a routing helper component that gets mounted on log-in when the app redirects to the `/agency/:agencyId/` path (reference L85 in `App.tsx`), and then again when the `Home.tsx` component mounts. 

It'll be tough to insert this logic into every component and track page visits that way. I looked in `App.tsx` and saw there was  this logic we can probably take advantage of (L43-47):

```
  const location = useLocation();
  const { userStore, api } = useStore();
  useEffect(() => {
    trackNavigation(location.pathname + location.search);
  }, [location]);
```

This will run any time the location changes/a user navigates to different pages within the app. `useLocation` returns a `location` object that has a bunch of useful properties about the current route/location incase you need it - but, since you're only keeping a timestamp on the last page they visited - I would try plugging in your logic to this `useEffect` above the `trackNavigation` logic and removing it from the `Router` and `Home` component. LMK if this works for you!

publisher/src/stores/UserStore.ts
```diff
@@ -241,6 +245,24 @@ class UserStore {
       });
     }
   }
+
+  async updateUserAgencyPageVisit(
+    userId: string | undefined,
```
mxosman, line 250: Since we're adding `userId` to this store/class - you don't need it as a parameter. You can access it via `this.userId` within this function.

publisher/src/stores/UserStore.ts
```diff
@@ -42,6 +44,7 @@ class UserStore {
     this.api = api;
     this.userAgencies = undefined;
     this.userInfoLoaded = false;
+    this.userId = undefined;
```
mxosman, line 47: Great call! Wow - I'm surprised we haven't been using a user id in the FE until now!

publisher/src/components/Home/Home.tsx
```diff
@@ -31,8 +31,8 @@ export const Home = observer(() => {
   const { userStore, homeStore } = useStore();
   const { agencyId } = useParams() as { agencyId: string };
   const navigate = useNavigate();
-
-  const { name, isAgencySuperagency } = userStore;
+  const { name, isAgencySuperagency, userId } = userStore;
+  userStore.updateUserAgencyPageVisit(userId, agencyId);
```
mxosman, line 35: React (and mobX) are particular about functions being called like this within the function body (esp. `async` functions) - because it will be called not only when the component mounts (component mounting === this `Home` function being called) - but when it re-renders because a user triggered a state change, it will fire this off again while we're still in the same path (because re-rendering just means it's calling this entire `Home` function again). Typically, if I need something to fire once when a component mounts, I'll put it inside of `useEffect` with an empty dependency array like this:

```
useEffect(() => {
  // whatever logic I want to run after the component mounts
}, []) // empty dependency array means run this once when the component mounts and not again while it's mounted
```

Otherwise, you'll run into some of weird issues each time a re-render happens - or a ton of API calls. Happy to clarify further if this reads more muddy than clear!

lilidworkin (commented):
Woo, awesome work on this @morden35 ! I will obviously leave the PR review to @mxosman but exciting to see how far you got on your own 🤩

publisher/src/stores/UserStore.ts
morden35: i know right!

publisher/src/router/Router.tsx
```diff
@@ -43,6 +43,10 @@ export const Router = () => {
 
   const isAgencyIdInUserAgencies = userStore.getAgency(agencyId);
 
+  useEffect(() => {
+    userStore.updateUserAgencyPageVisit(agencyId);
+  });
```
mxosman, line 48: Oh sorry I wasn't super clear, Michelle! I wouldn't call `updateUserAgencyPageVisit` here in the `Router.tsx`. I would put this logic in `App.tsx` within the useEffect that is already there in L45, like this:

```
  useEffect(() => {
    userStore.updateUserAgencyPageVisit(agencyId);
    trackNavigation(location.pathname + location.search);
  }, [location]);
```

mxosman: Though, you will need some reference to an `agencyId` in `App.tsx` - similar to how `Router` gets the agency ID from the params, you can add this line before the useEffect (like L45): 
`const { agencyId } = useParams() as { agencyId: string };`

Hopefully this works? Let me know if it doesn't!

mxosman: Oh whoops - sorry to interrupt you! Hell yeah to playing around with it!

morden35:
@mxosman after some playing around, I'm running into the following

1. When I move the logic to `App.tsx`, I get 400 errors from the API because at that point in the code upon login, I think `initialAgency` and `userStore` are undefined
Screenshot 2023-12-06 at 11 45 34 AM

2. When I try with the logic in Router.tsx, I see the following behavior
a. 2 calls to the endpoint upon login
b. 1 call to the endpoint upon switching agencies via the dropdown menu
c. 1 call to the endpoint upon switching tabs on the menu within a single agency

The behavior in number 2 is close to what I want, but not perfect. I would like to eliminate the double call to the endpoint upon login (part a). I'm a little torn on the part c behavior, because this will make a lot of unnecessary calls to the backend if the user is not switching agencies. However, this should handle the case if a user changes the url to a different agency in any given page.

Lmk if that makes sense and if you have any suggestions on how I can improve the behavior for either case!

morden35:
> b. Is this the expected behavior? I'm imagining when the user switches agencies, that page they refresh to will be the last visited for the user/agency.

Yes this is the desired behavior! That way we can record that a user visited a given agency if they switched while already being logged in.

morden35:
> c. When you say switching tabs on the menu - are you referring to the Home | Set Up Metrics | Enter Data, etc? I imagine this is unavoidable since each route change could be the last page visited - but maybe I'm misunderstanding this!

Sorry for the confusion! Yes, I was referring to the Home | Set Up Metrics | Enter Data tabs. Is this a big no no in terms of performance? I'm worried about unnecessary calls to the endpoint and subsequently the db

morden35:
> I would import useLocation and update the useEffect like so - so it calls this anytime the route changes and any time the agency id changes.

Thank you for the suggestion! So when I try this, I actually see the same behavior that I see without using any dependencies in `useEffect`:
a. 1 call to the endpoint upon login when lines in `index.tsx` are commented out (as desired)
b. 1 call to the endpoint upon switching agencies via the dropdown menu (as desired)
c. 1 call to the endpoint upon switching tabs on the menu within a single agency (not totally desired)

I like the idea behind this change though. If we could make `userStore.updateUserAgencyPageVisit(agencyId);` only run if `location` AND `agencyId` change (rather than just `location`), that would be great. Is this possible? Should I try adding some conditional statements?

mxosman:
> Commenting those lines out results in a single call to the endpoint! hooray!

Yay! You can rest assured that outside of development mode, it will only be one call each time! [React v18 did this double-call thing for several reasons, but the main one being that if double-running the useEffects causes issues, then it's a signal that there's something off with your code.]


> > b. Is this the expected behavior? I'm imagining when the user switches agencies, that page they refresh to will be the last visited for the user/agency.
> 
> Yes this is the desired behavior! That way we can record that a user visited a given agency if they switched while already being logged in.

Awesome!

> > c. When you say switching tabs on the menu - are you referring to the Home | Set Up Metrics | Enter Data, etc? I imagine this is unavoidable since each route change could be the last page visited - but maybe I'm misunderstanding this!
> 
> Sorry for the confusion! Yes, I was referring to the Home | Set Up Metrics | Enter Data tabs. Is this a big no no in terms of performance? I'm worried about unnecessary calls to the endpoint and subsequently the db
>
> c. 1 call to the endpoint upon switching tabs on the menu within a single agency (not totally desired)
> 

Ah - gotcha! In my mind these are each individual pages - are they being excluded from your tracking? If so, we can set something up where have a list of routes to exclude and check each time we switch routes and only run the call if it's a route not on the excluded list. But, if we're tracking the any last visited page, I think you'll want the call to happen when they switch to any one of these tabs/pages, right? [Sorry if I'm still misunderstanding! Here's how I'm understanding it so far: if Agency 1 goes to Home, then to Explore Data and then Set Up Metrics and switches agencies, then Agency 1's last visited page would be Set Up Metrics, right?]

I don't think it's a big no-no performance-wise in my mind, unless it's calling when you don't want it to, but we can get it specific enough to avoid that!

> I like the idea behind this change though. If we could make `userStore.updateUserAgencyPageVisit(agencyId);` only run if `location` AND `agencyId` change (rather than just `location`), that would be great. Is this possible? Should I try adding some conditional statements?
> 

I think this is possible - might be tricky - but I'll wait to hear back on the question above! Regardless, I'm confident we can find a good solution!!

> Edit: @mxosman I think I may have answered my own question! By adding `[userStore, agencyId]` to the dependency array i believe I get my desired behavior!

Ah - without the `location` here, the effect won't run when the route changes - only when the `agencyId` changes.

morden35:
> Ah - gotcha! In my mind these are each individual pages - are they being excluded from your tracking? If so, we can set something up where have a list of routes to exclude and check each time we switch routes and only run the call if it's a route not on the excluded list. But, if we're tracking any last visited page, I think you'll want the call to happen when they switch to any one of these tabs/pages, right? [Sorry if I'm still misunderstanding! Here's how I'm understanding it so far: if Agency 1 goes to Home, then to Explore Data and then Set Up Metrics and switches agencies, then Agency 1's last visited page would be Set Up Metrics, right?]

Good questions! Sorry for not being more clear from the beginning. I do not want to track individual pages. The purpose of this feature is to track logins (or visits) for user/agency pairs. So, I want to hit the endpoint in the following cases:
- a user logs in to Publisher for a given agency
- a user that is already logged in switches agencies via the dropdown
- a user that is already logged in switches agencies (from any given page) by changing the agency id in the url

I do not want to track:
- a user that is already logged in switches pages within the same agency

I actually think the current solution accomplishes this but lmk if you see any issues with what I have!

mxosman (approved):
> > Ah - gotcha! In my mind these are each individual pages - are they being excluded from your tracking? If so, we can set something up where have a list of routes to exclude and check each time we switch routes and only run the call if it's a route not on the excluded list. But, if we're tracking any last visited page, I think you'll want the call to happen when they switch to any one of these tabs/pages, right? [Sorry if I'm still misunderstanding! Here's how I'm understanding it so far: if Agency 1 goes to Home, then to Explore Data and then Set Up Metrics and switches agencies, then Agency 1's last visited page would be Set Up Metrics, right?]
> 
> Good questions! Sorry for not being more clear from the beginning. I do not want to track individual pages. The purpose of this feature is to track logins (or visits) for user/agency pairs. So, I want to hit the endpoint in the following cases:
> 
> * a user logs in to Publisher for a given agency
> * a user that is already logged in switches agencies via the dropdown
> * a user that is already logged in switches agencies (from any given page) by changing the agency id in the url
> 
> I do not want to track:
> 
> * a user that is already logged in switches pages within the same agency
> 
> I actually think the current solution accomplishes this but lmk if you see any issues with what I have!

OH - gotcha! I totally understand now. Sorry - thanks for your patience! I agree - your solution is spot on and LGTM! Super clean, Michelle!

PR #1084, opened 2023-12-05
[Publisher] Admin Panel: Agency Provisioning - Styling & Misc. Adjustments (1/n)

mxosman:
## Description of the change

Styling and clean up adjustments in preparation for the Agency Provisioning view.

## Related issues

Contributes to https://github.com/Recidiviz/justice-counts/issues/1051

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/InteractiveSearchList.tsx
+++ b/publisher/src/components/AdminPanel/InteractiveSearchList.tsx
@@ -66,6 +66,7 @@ export const InteractiveSearchList = ({
      * within a user's existing agencies [which has a `boxActionType === 'DELETE'`])
      */
     return (
+      selections.has(listItem.id) ||
       selections.has(+listItem.id) ||
       (listItem.action && boxActionType && listItem.action !== boxActionType)
     );
@@ -161,9 +162,11 @@ export const InteractiveSearchList = ({
         />
         <label htmlFor="search-input">
           {metadata?.searchBoxLabel}
-          <Styled.LabelButton onClick={resetFilteredList}>
-            Clear
-          </Styled.LabelButton>
+          {searchInputValue && (
+            <Styled.LabelButton onClick={resetFilteredList}>
+              Clear
+            </Styled.LabelButton>
+          )}
         </label>
       </Styled.InputLabelWrapper>
     </>
--- a/publisher/src/components/AdminPanel/InteractiveSearchList.tsx
+++ b/publisher/src/components/AdminPanel/InteractiveSearchList.tsx
@@ -125,11 +125,11 @@ export const InteractiveSearchList = ({
               );
             })
           ) : (
-            <Styled.NoResultsFound>
+            <Styled.EmptyListMessage>
               {searchInputValue
                 ? "No results found"
                 : metadata?.listBoxEmptyLabel}
-            </Styled.NoResultsFound>
+            </Styled.EmptyListMessage>
           )}
         </Styled.ChipContainer>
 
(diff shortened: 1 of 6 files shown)


publisher/src/components/AdminPanel/InteractiveSearchList.tsx
```diff
@@ -66,6 +66,7 @@ export const InteractiveSearchList = ({
      * within a user's existing agencies [which has a `boxActionType === 'DELETE'`])
      */
     return (
+      selections.has(listItem.id) ||
```
mxosman, line 69: `id` can be either a number or string so we'll check both here.

publisher/src/components/AdminPanel/InteractiveSearchList.tsx
```diff
@@ -161,9 +162,11 @@ export const InteractiveSearchList = ({
...
         <label htmlFor="search-input">
           {metadata?.searchBoxLabel}
-          <Styled.LabelButton onClick={resetFilteredList}>
-            Clear
-          </Styled.LabelButton>
+          {searchInputValue && (
+            <Styled.LabelButton onClick={resetFilteredList}>
+              Clear
+            </Styled.LabelButton>
+          )}
```
mxosman, line 169: The `Clear` button on these boxes before was always visible and clickable. This hides it until there's an input in the search.

publisher/src/components/AdminPanel/types.ts
```diff
@@ -133,7 +133,7 @@ export type SearchableListItem = {
   role?: UserRole;
 };
 
-export type SearchableSetIDs = Set<number>;
+export type SearchableSetIDs = Set<number | string>;
```
morden35, line 136: out of curiosity, why would this be a number or string? aren't these ids all coming from the same place (so they are either one or the other)?

mxosman: Great question! So in User Provisioning I was only dealing with `number` type ids for these searchable sets. When I started working on the Agency Provisioning view, I quickly realized that we'd also need to support `string` type ids for things like a set of systems or state codes and county codes, etc.

nichelle-hall (approved):
LGTM ✨

When playing around with this component on you branch (the one with all the changes) it works like a charm!

publisher/src/components/AdminPanel/types.ts
morden35: oh i see! So maybe `SearchableSetIDs` should be renamed to something like `SearchableSetObjects`? since this set won't always be ids?

mxosman: The ids in the context of this searchable set and the search component are more/less keys - maybe that's a better word to describe this. So, for agencies that have an `id` that's a number, this set will be a set of numbers that the search component can distinguish between. For things like systems, the set will be a set of system key strings (e.g. "LAW_ENFORCEMENT", "DEFENSE", etc.) that can be distinguished, if that makes sense?

mxosman: If `SearchableSetKeys` sounds like a better description instead of `SearchableSetIDs`, I'll make a note to update the naming in the later downstream PR to avoid a bunch of conflicts!

morden35: that makes sense! I could go either way on a change - no biggie either way

morden35 (approved):

publisher/src/components/AdminPanel/types.ts
mxosman: I'll update it to `SearchableSetKeys` - feels like it's much clearer! Thank you for calling this out!

mxosman: I don't know what I was worried about - I thought I used this type throughout the components, but it's only used within this file, so updating shouldn't cause conflicts!

PR #1085, opened 2023-12-05
[Publisher] Admin Panel: Agency Provisioning - Store Updates (2/n)

mxosman:
## Description of the change

Adds Agency Provisioning related methods and helpers in the `AdminPanelStore` and updates types. This is more/less the "backend" of the frontend - all of these methods will be plugged into the UI and used to update the changes made in the Agency Provisioning view to send off to the BE.

## Related issues

Contributes to #1051 

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AdminPanel.test.tsx
+++ b/publisher/src/components/AdminPanel/AdminPanel.test.tsx
@@ -328,8 +328,9 @@ test("Adding an agency adds agency to user's agency list", async () => {
           state: "New York",
           state_code: "us_ny",
           super_agency_id: null,
+          child_agency_ids: [],
           systems: ["PRISONS", "SUPERVISION", "PAROLE", "PROBATION"],
-          team: [],
+          team: {},
         },
         {
           created_at: null,
@@ -342,8 +343,9 @@ test("Adding an agency adds agency to user's agency list", async () => {
           state: "New York",
           state_code: "us_ny",
           super_agency_id: null,
+          child_agency_ids: [],
           systems: ["PRISONS"],
-          team: [],
+          team: {},
         },
         {
           created_at: null,
@@ -356,8 +358,9 @@ test("Adding an agency adds agency to user's agency list", async () => {
           state: "Oregon",
           state_code: "us_or",
           super_agency_id: null,
+          child_agency_ids: [],
           systems: ["JAILS"],
-          team: [],
+          team: {},
         },
       ],
       (agency) => agency.id
--- a/publisher/src/components/AdminPanel/UserProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/UserProvisioning.tsx
@@ -79,7 +79,7 @@ export const UserProvisioning: React.FC<UserProvisioningProps> = observer(
 
     /** Available agencies to add from */
     const availableAgencies = agencies.filter(
-      (agency) => !selectedUserAgenciesIDsSet.has(agency.id)
+      (agency) => !selectedUserAgenciesIDsSet.has(+agency.id)
     );
     const availableAgenciesIDs = availableAgencies.map((agency) => +agency.id);
     const availableAgenciesIDsSet = new Set(availableAgenciesIDs);
--- a/publisher/src/components/AdminPanel/types.ts
+++ b/publisher/src/components/AdminPanel/types.ts
@@ -39,18 +39,24 @@ export enum SaveConfirmationTypes {
 
 export type SaveConfirmationType = `${SaveConfirmationTypes}`;
 
+export type ProvisioningProps = {
+  selectedIDToEdit?: string | number;
+  closeModal: () => void;
+};
+
 /** Agency Types */
 
 export type Agency = {
-  id: number;
+  id: string | number;
   name: string;
   systems: AgencySystems[];
   state: StateCodeValue;
-  state_code: keyof typeof StateCodes;
-  fips_county_code?: keyof typeof FipsCountyCodes | null;
+  state_code: StateCodeKey | null;
+  fips_county_code: FipsCountyCodeKey | null;
   team: AgencyTeamMember[];
   super_agency_id: number | null;
   is_superagency: boolean | null;
+  child_agency_ids: number[];
   is_dashboard_enabled: boolean | null;
   created_at: string | null;
   settings: {
@@ -58,6 +64,10 @@ export type Agency = {
     value: string | null;
     source_id: number;
   }[];
+};
+
+export type AgencyWithTeamByID = Omit<Agency, "team"> & {
+  team: Record<string, AgencyTeamMember[]>;
 };
 
 export type AgencyResponse = {
@@ -72,6 +82,21 @@ export const AgencyProvisioningSettings = {
 
 export type AgencyProvisioningSetting =
   (typeof AgencyProvisioningSettings)[keyof typeof AgencyProvisioningSettings];
+
+export type AgencyProvisioningUpdates = {
+  agency_id?: number;
+  name: string;
+  state_code: StateCodeKey | null;
+  fips_county_code: FipsCountyCodeKey | null;
+  systems: AgencySystems[];
+  is_dashboard_enabled: boolean | null;
+  super_agency_id: number | null; // If this is set, then the agency is a child agency belonging to the superagency w/ this ID
+  is_superagency: boolean | null;
+  child_agency_ids: number[];
+  team: AgencyTeamUpdates[];
+};
+
+export type AgencyTeamUpdates = { user_account_id: number; role: UserRole };
 
 /** User Types */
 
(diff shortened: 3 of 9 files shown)


publisher/src/components/AdminPanel/UserProvisioning.tsx
```diff
@@ -79,7 +79,7 @@ export const UserProvisioning: React.FC<UserProvisioningProps> = observer(
 
     /** Available agencies to add from */
     const availableAgencies = agencies.filter(
-      (agency) => !selectedUserAgenciesIDsSet.has(agency.id)
+      (agency) => !selectedUserAgenciesIDsSet.has(+agency.id)
```
mxosman, line 82: Fixing TS error and coercing the ID to a number (because it's coming from an `Object.keys` so it will always be a string)

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -106,7 +160,13 @@ class AdminPanelStore {
 
       /** Hydrate store with a list of systems and a list of sorted agencies from response  */
       runInAction(() => {
-        this.agenciesByID = groupBy(data.agencies, (agency) => agency.id);
+        this.agenciesByID = groupBy(
+          data.agencies.map((agency) => ({
+            ...agency,
+            team: groupBy(agency.team, (member) => member.user_account_id),
```
mxosman, line 166: Team members will also be grouped by ID for quick lookup

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -30,41 +40,85 @@ import {
...
+      )
+      .map((countyCode) => {
+        const lowercaseCountyCode =
+          countyCode.toLocaleLowerCase() as FipsCountyCodeKey;
+        return {
+          id: countyCode,
+          name: FipsCountyCodes[lowercaseCountyCode],
+        };
+      });
+  }
```
mxosman, line 120: The reason why `searchableStates` is a `static` getter and `searchableCounties` is not is because `searchableStates` doesn't rely on anything in this object - it just normalizes a list of states that we already have - and if you're not using `this.` in a `getter` you get a lint error. The `searchableCounties` though does rely on the selected state - and thus will use `this.` and does not need to be a `static` getter.

publisher/src/components/AdminPanel/AdminPanel.test.tsx
```diff
@@ -356,8 +358,9 @@ test("Adding an agency adds agency to user's agency list", async () => {
           state: "Oregon",
           state_code: "us_or",
           super_agency_id: null,
+          child_agency_ids: [],
           systems: ["JAILS"],
-          team: [],
+          team: {},
```
mxosman, line 363: Fixing TS errors in this file

publisher/src/stores/AdminPanelStore.test.tsx
```diff
@@ -47,6 +49,7 @@ export const mockUsersResponse = {
           state: "New York",
           state_code: "us_ny",
           super_agency_id: null,
+          child_agency_ids: [],
           systems: ["PRISONS"],
           team: [],
```
morden35, line 54: should these be sets?

morden35 (approved):
lgtm!

publisher/src/stores/AdminPanelStore.test.tsx
mxosman: Oo - great question! So, this particular object will be in the same format that the BE sends for the users response - once this response is received, then the FE converts these into sets to store them.

publisher/src/components/AdminPanel/types.ts
```diff
@@ -39,18 +39,24 @@ export enum SaveConfirmationTypes {
...
+  id: string | number;
   name: string;
   systems: AgencySystems[];
   state: StateCodeValue;
-  state_code: keyof typeof StateCodes;
-  fips_county_code?: keyof typeof FipsCountyCodes | null;
-  team: AgencyTeamMember[];
+  state_code: StateCodeKey | null;
+  fips_county_code: FipsCountyCodeKey | null;
+  team: AgencyTeamMemberWithID[];
```
nichelle-hall, line 56: Now ID will be in all responses for an agency's team, not only in the admin panel but anytime we fetch a team. You don't need to create a new type `AgencyTeamMemberWithID`, but rather update AgencyTeamMember to have an ID.

publisher/src/stores/AdminPanelStore.test.tsx
morden35: oh makes sense!

publisher/src/components/AdminPanel/types.ts
```diff
@@ -60,6 +66,14 @@ export type Agency = {
   }[];
 };
 
+export type AgencyTeamMemberWithID = AgencyTeamMember & {
+  user_account_id: number;
+};
+
+export type AgencyWithTeamByID = Omit<Agency, "team"> & {
```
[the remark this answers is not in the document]
mxosman: I know - thank goodness for it too - otherwise me and TS would still be having a brawl. 😅

publisher/src/components/AdminPanel/types.ts
mxosman: Oh nice! Thank you for that - will update!

publisher/src/stores/AgencyStore.ts
```diff
@@ -326,6 +326,7 @@ class AgencyStore {
 
   inviteTeamMember = (name: string, email: string) => {
     const newTeamMember: AgencyTeamMember = {
+      user_account_id: 0,
```
mxosman, line 329: @nichelle-hall - I updated the `AgencyTeamMember` type to include `user_account_id` as a `number` type. It threw and TS error here because I'm requiring the type, but in this part of the code where its inviting a new team member (via Publisher), it sends empty values for the `auth0_user_id` - what do you think it should send back for `user_account_id` in this case? `null`?

nichelle-hall: Yeah! In this case `user_account_id` should be null.

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -104,9 +157,21 @@ class AdminPanelStore {
...
+       * their team members grouped by user ID) from response
+       */
       runInAction(() => {
-        this.agenciesByID = groupBy(data.agencies, (agency) => agency.id);
+        this.agenciesByID = groupBy(
+          data.agencies.map((agency) => ({
+            ...agency,
+            team: groupBy(
+              agency.team,
+              (member) => member.user_account_id || member.auth0_user_id
```
lilidworkin, line 170: Is it safe to use either user_account_id or auth0_user_id? Do we depend on it being one or the other anywhere later down the line?

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -171,18 +234,119 @@ class AdminPanelStore {
...
+    };
+    this.agenciesByID[agency.id] = [agencyWithGroupedTeams];
+  }
+
+  async saveAgencyProvisioningUpdates() {
+    try {
+      const response = (await this.api.request({
+        path: `/admin/agency`,
+        method: "PUT",
+        body: this.agencyProvisioningUpdates,
```
lilidworkin, line 304: I like how clean this is!

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -217,12 +381,32 @@ class AdminPanelStore {
...
+      | UserWithAgenciesByID
+      | AgencyTeamMember
   >(obj: Record<string, T[]>) {
     return AdminPanelStore.sortListByName(
       Object.values(obj).flatMap((item) => item)
     );
   }
+
+  /**
+   * Updates a set of selections by either adding or removing a specified item.
```
lilidworkin, line 396: Nice helper! super nit, maybe belongs better in a util file?

lilidworkin (approved):
LGTM!

publisher/src/stores/AdminPanelStore.ts
mxosman: Great question! This was mostly to satisfy TypeScript since `user_account_id` could be `null` in this scenario (https://github.com/Recidiviz/justice-counts/pull/1085#discussion_r1416257816). It shouldn't ever be `null` here since the list of existing team members will always have some kind of `user_account_id`, but since it's typed as a possible `null` and `groupBy` can't group by `null` - I'm having it fall back on the `auth0_user_id` if `user_account_id` is `null` (which hopefully shouldn't happen here). LMK if this feels like a weird way of handling this.

publisher/src/stores/AdminPanelStore.ts
mxosman: Great idea!

publisher/src/stores/AdminPanelStore.ts
lilidworkin: Ah that makes sense! That's fine!

PR #1086, opened 2023-12-05
[Publisher] Admin Panel: Agency Provisioning - Overview (3/n)

mxosman:
## Description of the change

Implements the Agency Provisioning overview page where users can search, filter and select from a list of agencies. Users can also filter the list to only show superagencies and/or agencies that have live dashboards via the checkboxes between the search box and 'Create Agency' button.

One other thing to note - I might be changing my mind on one single call to the BE for both users and agencies and instead move the calls to each time you open a tab because I can see things getting stale much quicker. Totally open to your thoughts!

Demo:

https://github.com/Recidiviz/justice-counts/assets/59492998/5717b9a9-7601-40b3-83ce-e6fe2b0d5012

## Related issues

Contributes to #1051

### Development

### Code review


publisher/src/components/AdminPanel/UserProvisioning.tsx
```diff
@@ -29,19 +29,15 @@ import {
...
 
-type UserProvisioningProps = {
-  selectedUserID?: string | number;
-  closeModal: () => void;
-};
-
-export const UserProvisioning: React.FC<UserProvisioningProps> = observer(
-  ({ selectedUserID, closeModal }) => {
+export const UserProvisioning: React.FC<ProvisioningProps> = observer(
+  ({ selectedIDToEdit, closeModal }) => {
```
mxosman, line 40: Updating variable names and small clean up in this file

publisher/src/components/HelpCenter/LinkToPublisherDashboard.tsx
```diff
@@ -36,25 +36,27 @@ export const LinkToPublisher: React.FC<
...
 });
 
-export const LinkToDashboard: React.FC<PropsWithChildren> = observer(
-  ({ children }) => {
-    const { api, userStore } = useStore();
-    const agencyIdLocalStorage = localStorage.getItem("agencyId");
-    const agencyId =
-      agencyIdLocalStorage || userStore.getInitialAgencyId()?.toLocaleString();
-    const agencyName = agencyId && userStore.getAgency(agencyId)?.name;
+export const LinkToDashboard: React.FC<
```
mxosman, line 39: Updating this to accept an `agencyID` so the Agency Provisioning view can use it. Right now it gets the current user's agency ID (the current agency they're switched to in Publisher) and uses that to create a link. Now, we can pass in an `agencyID` and it will update the link accordingly.

mxosman: Updating the type, passing in `agencyID` as props and including it in L45 is the only change. The rest is prettier formatting.

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
```diff
@@ -0,0 +1,259 @@
...
+          </label>
+        </Styled.InputLabelWrapper>
+
+        {/* Create Agency Button */}
+        <Styled.ButtonWrapper>
+          <Button
+            label="Create Agency"
+            onClick={openModal}
+            buttonColor="blue"
+          />
```
nichelle-hall, line 199: When users are creating an agency, can the `is_dashboard_enabled` checkbox be clicked by default? Of course, they can unclick, but we want to default that dashboard is on when creating a new user.

nichelle-hall (commented):
I'm not sure

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
mxosman: Absolutely!

mxosman:
> I'm not sure

I feel that - I'm never sure, to be honest... 😄

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
lilidworkin: Yep great call @nichelle-hall !

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
```diff
@@ -0,0 +1,259 @@
...
+  const searchAndFilter = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchInput(e.target.value);
+    setFilteredAgencies(
+      AdminPanelStore.searchList(agencies, e.target.value, searchByKeys)
+    );
+  };
+  const editAgency = (agencyID: string | number) => {
+    const selectedAgency = agenciesByID[agencyID][0];
+    setSelectedAgencyID(agencyID);
+    updateAgencyID(+agencyID);
```
lilidworkin, line 76: This is a super nit -- I wondered about this in the other PR but wanted to see how it was going to be used -- but does it make sense to have just one updateAgency method that takes in all of these parameters? Rather than having a separate method for each?

lilidworkin (approved):
LGTM!

> One other thing to note - I might be changing my mind on one single call to the BE for both users and agencies and instead move the calls to each time you open a tab because I can see things getting stale much quicker. Totally open to your thoughts!

@mxosman I think I agree -- separate calls seems to make more sense to me. Or, we could have one call but just refresh both every time you switch tabs? If that's an easier solution to implement in the short term.

mxosman:
> @mxosman I think I agree -- separate calls seems to make more sense to me. Or, we could have one call but just refresh both every time you switch tabs? If that's an easier solution to implement in the short term.

That's a really great idea! I think it makes more sense to do the latter - refresh both since they both depend on each other. It would be nice to get the latest agencies list when provisioning a user, vice versa.

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
lilidworkin: No yea this all makes sense, keep it as you ahve it! Thanks for the explanation!

PR #1089, opened 2023-12-06
[Publisher] Admin Panel: Admin Panel Role & Publisher Entry Point

### LINKED ISSUE #1052: [Publisher] Admin Panel: User Permissions + Entry Point

mxosman:
## Description of the change

Adds an entry point to the admin panel via the profile dropdown menu in Publisher and creates a new Global Justice Counts Admin role in Auth0 (staging for now) that will be used to gate access to the admin panel. It might be worth updating the backend as well to secure these endpoints since the flags set on the FE won't provide full security (task https://github.com/Recidiviz/recidiviz-data/issues/26006). 

Here's what I updated in our tenant in staging:
* Under **User Management > Roles**, I created a `justice-counts-admin` role (referred to as "Global" Justice Counts admin here in the codebase to differentiate from Publisher's JC admin role - let me know if you have better ideas!)
Screenshot 2023-12-07 at 9 21 28 AM

* Under **Actions > Flows > Login**, I updated the `Add custom claims to tokens` post-login script and added a couple of lines to add the roles to the tokens.

I also updated the configuration doc to include the action scripts we have (will update the production section once that's configured as well):
https://github.com/Recidiviz/recidiviz-data/tree/main/recidiviz/justice_counts/control_panel (under Helpful Links > [configuration doc])

Loom video showing adding/removing roles in Auth0 and reading roles in the FE:
https://www.loom.com/share/dde12a91e4ad44f8b2b2864c80afcb2e?sid=d43b252f-0fc1-456a-a0c7-efb8a4220dcf

## Related issues

Closes #1052

### Development

### Code review

### DIFF
--- a/publisher/src/components/Auth/AuthStore.ts
+++ b/publisher/src/components/Auth/AuthStore.ts
@@ -58,7 +58,10 @@ export class AuthStore {
     return this.user?.[AUTH0_NAMESPACE_ROLES] || [];
   }
 
-  /** Refers to administrator role outside of Publisher w/ user & agency provisioning privileges */
+  /**
+   * Refers to administrator role w/ user & agency provisioning privileges separate
+   * from Publisher's "Justice Counts Admin" role.
+   *  */
   get isGlobalJusticeCountsAdmin(): boolean {
     return this.userAuth0Roles.includes(GLOBAL_JUSTICE_COUNTS_ADMIN_ROLE);
   }
--- a/publisher/src/components/Auth/AuthStore.ts
+++ b/publisher/src/components/Auth/AuthStore.ts
@@ -58,6 +58,7 @@ export class AuthStore {
     return this.user?.[AUTH0_NAMESPACE_ROLES] || [];
   }
 
+  /** Refers to administrator role outside of Publisher w/ user & agency provisioning privileges */
   get isGlobalJusticeCountsAdmin(): boolean {
     return this.userAuth0Roles.includes(GLOBAL_JUSTICE_COUNTS_ADMIN_ROLE);
   }
--- a/publisher/src/components/Auth/AuthStore.ts
+++ b/publisher/src/components/Auth/AuthStore.ts
@@ -26,8 +26,8 @@ import qs from "qs";
 
 import { identify } from "../../analytics";
 
-export const APP_METADATA_CLAIM =
-  "https://dashboard.recidiviz.org/app_metadata";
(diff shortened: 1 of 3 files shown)


publisher/src/components/Auth/AuthStore.ts
```diff
@@ -54,6 +54,14 @@ export class AuthStore {
     this.isLoading = true;
   }
 
+  get userAuth0Roles(): string[] {
+    return this.user?.[AUTH0_NAMESPACE_ROLES] || [];
+  }
+
+  get isGlobalJusticeCountsAdmin(): boolean {
+    return this.userAuth0Roles.includes(GLOBAL_JUSTICE_COUNTS_ADMIN_ROLE);
+  }
```
mxosman, line 63: I put this here instead of the **UserStore** because most of the **UserStore** logic felt specific to the user/agency relationship, where as the **AuthStore** feels like a layer above that and could house more "global" type-things. I could go either way! But it felt like it might be confusing mixing this in with the other **UserStore** permissions.

Also totally open to better naming ideas - always and forever!

lilidworkin: This decision makes sense to me! I think the naming is fine, maybe we can just add a comment inline to explain that this refers to an admin role beyond/outside of Publisher, which grants you access to the admin panel?

lilidworkin (approved):
@mxosman this is excellent! So clean and simple. Great job figuring out how to do the custom claims, and thank you for updating the auth0 config doc! I think this is actually the first time I really look at that doc -- it's awesome!

And great call re: filing BE companion task. I don't think it's urgent since the endpoints are still gated to people who are logged in, and I don't think there's a super high security risk of our regular users figuring out how to hack into our admin panel :) But we should still def do it! Haha

publisher/src/components/Auth/AuthStore.ts
mxosman: Such a great idea! Will do!

mxosman:
> @mxosman this is excellent! So clean and simple. Great job figuring out how to do the custom claims, and thank you for updating the auth0 config doc! I think this is actually the first time I really look at that doc -- it's awesome!
> 
> And great call re: filing BE companion task. I don't think it's urgent since the endpoints are still gated to people who are logged in, and I don't think there's a super high security risk of our regular users figuring out how to hack into our admin panel :) But we should still def do it! Haha

Thank you so much, Lili! It was really satisfying to see this work. Haha - totally agree on urgency & risk - hopefully it's straightforward, but you never know with Auth0!

Issue #1075, opened 2023-11-28
[Publisher] Admin Panel: Clean up


mxosman:
- [x] Reasonable responsiveness (#1105)
- [x] Styling adjustments (#1105)
- [x] Figure out how to better sync the loading of the fetched users/agencies (#1103)
- [x] Ensure proper error handling and messaging (#1105)
- [x] Update "System" to "Sector" (#1105)
- [x] Remove superagency system addition/removal -> checkmark update logic b/c it is no longer necessary (user doesn't select the superagency system anymore, the checkmark adds or removes it accordingly) (#1105)
- [x] Fix clash between filter checkmarks and search box in `AgencyProvisioningOverview` (#1105)

PR #1103, opened 2023-12-12
[Publisher] Admin Panel: Create User/Agency from opposite flows

### LINKED ISSUE #1058: [Publisher] Admin Panel: Create New User/Agency from both Provisioning Flows

mxosman:
## Description of the change

This implementation allows users to quickly create a new user _while_ in the middle of creating an agency - and vice versa - quickly create a new agency _while_ in the middle of creating a new user. There are some edge cases that still need to be tested for (e.g. the auto adding of CSG members when creating a new agency, etc.) - but this works well so far in local development. Will do a more thorough testing in staging, but mostly trying to get the foundation in for now. Hopefully this makes things a bit easier for us when provisioning new users and agencies - but if it ends up being a pain, it's easy enough to strip out.

Demo (user provisioning):

https://github.com/Recidiviz/justice-counts/assets/59492998/e58bce4f-8998-4b9a-af2e-cd5910b644af

Demo (agency provisioning):

https://github.com/Recidiviz/justice-counts/assets/59492998/042460fa-652a-4750-b77a-b446de0ec3ee

## Related issues

Closes #1058, Contributes to #1075

### Development

### Code review

### DIFF
--- a/common/components/DelayedRender/DelayedRender.tsx
+++ b/common/components/DelayedRender/DelayedRender.tsx
@@ -1,0 +1,32 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2023 Recidiviz, Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import { PropsWithChildren, useEffect, useState } from "react";
+
+export const DelayedRender: React.FC<{ delay: number } & PropsWithChildren> = ({
+  delay,
+  children,
+}) => {
+  const [delayed, setDelayed] = useState(true);
+
+  useEffect(() => {
+    const timeout = setTimeout(() => setDelayed(false), delay);
+    return () => clearTimeout(timeout);
+  }, []);
+
+  return !delayed ? children : null;
+};
--- a/common/components/DelayedRender/DelayedRender.tsx
+++ b/common/components/DelayedRender/DelayedRender.tsx
@@ -26,7 +26,7 @@ export const DelayedRender: React.FC<{ delay: number } & PropsWithChildren> = ({
   useEffect(() => {
     const timeout = setTimeout(() => setDelayed(false), delay);
     return () => clearTimeout(timeout);
-  }, []);
+  }, [delay]);
 
   return !delayed ? children : null;
 };
--- a/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
@@ -19,8 +19,8 @@ import { Button } from "@justice-counts/common/components/Button";
 import { Dropdown } from "@justice-counts/common/components/Dropdown";
 import { MiniLoader } from "@justice-counts/common/components/MiniLoader";
 import {
-  TabOption,
   TabbedBar,
+  TabOption,
 } from "@justice-counts/common/components/TabbedBar";
 import {
   AgencySystem,
--- a/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
@@ -969,7 +969,7 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
                           </Styled.ActionButton>
                         )}
 
-                        {/* Create New User Button (TODO(#1058)) */}
+                        {/* Create New User Button */}
                         <Styled.ActionButton onClick={openSecondaryModal}>
                           Create New User
                         </Styled.ActionButton>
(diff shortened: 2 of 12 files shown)


publisher/src/components/AdminPanel/UserProvisioning.tsx
```diff
@@ -338,42 +344,46 @@ export const UserProvisioning: React.FC<ProvisioningProps> = observer(
...
+                          setAddOrDeleteAgencyAction((prev) =>
+                            prev === InteractiveSearchListActions.DELETE
+                              ? undefined
+                              : InteractiveSearchListActions.DELETE
+                          );
+                        }}
+                      >
+                        Delete Agencies
+                      </Styled.ActionButton>
+                    )}
```
mxosman, line 379: No changes here

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -189,9 +189,8 @@ class AdminPanelStore {
     }
   }
 
-  fetchUsersAndAgencies() {
-    this.fetchUsers();
-    this.fetchAgencies();
+  async fetchUsersAndAgencies() {
+    await Promise.all([this.fetchUsers(), this.fetchAgencies()]);
```
mxosman, line 193: Fixes the loader going away before data is ready to render

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
```diff
@@ -129,14 +138,25 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
...
-        label: AgencyProvisioningSettings.TEAM_MEMBERS_ROLES,
-        onClick: () =>
-          setCurrentSettingType(AgencyProvisioningSettings.TEAM_MEMBERS_ROLES),
-        selected:
-          currentSettingType === AgencyProvisioningSettings.TEAM_MEMBERS_ROLES,
-      },
+      /**
+       * Hide the Team Member & Roles tab if we are in the secondary modal b/c it is not
+       * necessary in that flow.
+       */
```
mxosman, line 144: Will revisit this decision based on the auto-adding of CSG team members to newly created agencies. If you're just creating an agency real quick to add to the user you're creating, maybe you don't care about seeing the list of CSG users being added? Or, I might keep this tab open (that way you'll still see the CSG team members being added) and just disable the action buttons (to avoid ending up in a weird circular cycle).

common/components/DelayedRender/DelayedRender.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) => {
+  const [delayed, setDelayed] = useState(true);
+
+  useEffect(() => {
+    const timeout = setTimeout(() => setDelayed(false), delay);
+    return () => clearTimeout(timeout);
+  }, [delay]);
+
+  return !delayed ? children : null;
+};
```
nichelle-hall, line 32: Why do you need this?

nichelle-hall (commented):
I don't think this code is rebased on the changes that send user_account_id over on `/admin/user` PUT because while playtesting I always got a `user already exists` error after trying to save the user after creating a new agency in the flow.

common/components/DelayedRender/DelayedRender.tsx
nichelle-hall: Oh! is this part of what you referenced before? That it insures that the data is loaded before the component renders? Is that now what useEffect does?

mxosman:
> I don't think this code is rebased on the changes that send user_account_id over on `/admin/user` PUT because while playtesting I always got a `user already exists` error after trying to save the user after creating a new agency in the flow.

Oh weird! I'll try giving this another rebase. I was just testing both flows and it feels like things are working as expected on my end. Just to clarify so I can repro - you're creating a new user (ensuring the email doesn't already match one that exists), and you create a new agency, save that and then try to save the user and it shows user already exists?

common/components/DelayedRender/DelayedRender.tsx
mxosman: Good question! This is actually to slightly delay the render of the secondary modal - otherwise, when the secondary modal is open, it happens in an instant and looks like the same modal is open but with the title and things changed. This gives the original modal time to start sliding off screen and then the second modal appearing.

So basically, we can wrap this component around any component we want to wait to delay rendering (usually for transitions like this).

The other bit about waiting for the data to load in the other PR - I just needed to wait for the fetch promises to resolve before switching the loader off - classic async/sync mismatch.

nichelle-hall:
> > I don't think this code is rebased on the changes that send user_account_id over on `/admin/user` PUT because while playtesting I always got a `user already exists` error after trying to save the user after creating a new agency in the flow.
> 
> Oh weird! I'll try giving this another rebase. I was just testing both flows and it feels like things are working as expected on my end. Just to clarify so I can repro - you're creating a new user (ensuring the email doesn't already match one that exists), and you create a new agency, save that and then try to save the user and it shows user already exists?

Yes! I'll try to repro now.

nichelle-hall (approved):

mxosman:
Yay - thank you, Nichelle!

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
lilidworkin: Just to clarify, are we talking about whether the user can _see_ the CSG users being added, or whether the CSG users actually _do_ get added? I do think if possible we should be consistent about always adding them, but I agree that the visual feedback isn't necessary in this flow!

lilidworkin (commented):
What a really awesome idea to enable this!

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
mxosman: Just seeing - CSG (and Recidiviz) users will always get added! 

I think before the auto-adding, this was a blank page when you tabbed over, which is likely why I felt - let's just not render it. But now, I'm going to play around with the latter option - showing this tab and the CSG + Recidiviz users being added, but hide the action buttons of adding/deleting/creating users.

mxosman:
> What a really awesome idea to enable this!
> 

Hahaha - I'm obsessed with these gifs!

PR #1110, opened 2023-12-21
[Bug][Publisher] Admin Panel: Playtesting Followup - Fix crash edgecase when username is a string full of spaces

mxosman:
## Description of the change

Fixes a crash that happens when logging into Publisher when your name set as only spaces. In addition to that, adds extra trimming of leading spaces (in admin panel and Publisher account settings) and a bit of logic to prevent users from emptying out the name & email fields in Publisher account settings (which is already in place in admin panel). Thanks to Forrest for uncovering this weird edge case.

I tested this manually by adjusting my name to only spaces via Auth0 (no crash when logging in), and trying to empty the name and email fields in Publisher account settings (reasonably difficult to empty) and the name field in the admin panel (email is not editable at this time - so at creation time, this isn't necessary because the email is being validated).

### Development

### Code review

### DIFF
--- a/publisher/src/stores/AdminPanelStore.ts
+++ b/publisher/src/stores/AdminPanelStore.ts
@@ -222,7 +222,7 @@ class AdminPanelStore {
   /** User Provisioning */
 
   updateUsername(username: string) {
-    this.userProvisioningUpdates.name = username;
+    this.userProvisioningUpdates.name = username.trimStart();
   }
 
   updateEmail(email: string) {
--- a/publisher/src/components/Menu/Menu.tsx
+++ b/publisher/src/components/Menu/Menu.tsx
@@ -61,7 +61,7 @@ const Menu: React.FC = () => {
       if (splitName.length > 1) {
         return (splitName[0][0] + splitName[1][0]).toUpperCase();
       }
-      return splitName[0][0].toUpperCase();
+      return splitName[0][0]?.toUpperCase();
     }
   };
 
--- a/publisher/src/components/Menu/Menu.tsx
+++ b/publisher/src/components/Menu/Menu.tsx
@@ -56,7 +56,8 @@ const Menu: React.FC = () => {
 
   const usernameToInitials = () => {
     if (userStore.name) {
-      const splitName = userStore.name.split(" ");
+      const splitName = userStore.name.trim().split(" ");
+
       if (splitName.length > 1) {
         return (splitName[0][0] + splitName[1][0]).toUpperCase();
       }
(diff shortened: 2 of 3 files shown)


publisher/src/stores/AdminPanelStore.ts
```diff
@@ -222,7 +222,7 @@ class AdminPanelStore {
   /** User Provisioning */
 
   updateUsername(username: string) {
-    this.userProvisioningUpdates.name = username;
+    this.userProvisioningUpdates.name = username.trimStart();
```
morden35, line 225: nice! can we also trim the end?

morden35 (approved):
nice! one non-blocking question in-line

publisher/src/stores/AdminPanelStore.ts
mxosman: Oo - good question! So, I started out with `.trim()` which trims both the beginning and the end, but quickly realized that trimming at the end prevents you from adding a space after the first name input if you're trying to add a first and last name. This is because it's updating, re-rendering the updated value and autosaving on the fly and the trimming would be included in this cycle - so your experience would be trying to type in a space to get to that second word, but it's trimmed by the time it's re-rendered - and so no space ever appears in the input. 

I would need to refactor the logic in this component a bit more and play around with it (believe me it could use a refactor) - and use a save button perhaps to do the final trimming before sending it off instead of relying on autosave. I figured it wasn't worth the trouble at the moment and just addressed the critical bug caused by the leading spaces.

morden35: oh i think i see! makes sense. in that case, wouldn't we run into a similar issue with the last name? in other words, does the white space get trimmed at the start of the last name?

mxosman: Ah - good question! It's only taking in a single string value for `username` (so you can add first, last or as many words in this string) - and that entire string will be trimmed at the start - e.g. if you paste in " FirstName LastName ", it will trim it to "FirstName LastName ".

Btw - here's an example of what the `trim` behavior does as you're typing in the input - I feel like it's better visualized:
https://codesandbox.io/p/sandbox/happy-cartwright-qq37k2?file=%2Fsrc%2FApp.js (feel free to comment/uncomment the `trim` and `trimStart` in `src/App.js` and try entering two words separated by space in the input on the preview browser on the right)

morden35: got it thank you for the explanation! I didn't realize we were working with a username rather than a first and last name!

PR #1117, opened 2023-12-26
[Publisher] Admin Panel: Playtesting Followup - Empty search input in overview pages after saving & refactor filtering logic

mxosman:
## Description of the change

Empties out the search input in the overview pages after saving. 

While I was doing this, I ran into a "off-by-one-render" (for lack of a better phrase) bug in the overview pages and realized that I had fallen into an overcomplicated state w/ mobx and local state (search + filtering logic) that could be dramatically simplified. A nice benefit of this small refactor is that it also takes care of the feedback of retaining the value in the searchbox while checking/unchecking the filter checkboxes in the agency provisioning overview - and everything searches and filters as expected.

https://github.com/Recidiviz/justice-counts/assets/59492998/e279d3e7-e26e-4f3f-9f51-d8952803687d

If anyone is super curious - I'm happy to sync with you & dive deeper into the mobx and local state adjustments!

### Development

### Code review

### DIFF
--- a/publisher/src/stores/AdminPanelStore.ts
+++ b/publisher/src/stores/AdminPanelStore.ts
@@ -414,7 +414,12 @@ class AdminPanelStore {
     searchInput: string,
     searchByKeys: (keyof T)[]
   ) {
-    const regex = new RegExp(`${searchInput}`, `i`);
+    // Escape special characters in search input
+    const sanitizedSearchInput = searchInput.replace(
+      /[.*+?^${}()|[\]\\]/g,
+      "\\$&"
+    );
+    const regex = new RegExp(`${sanitizedSearchInput}`, `i`);
     return list.filter((listItem) =>
       searchByKeys.some(
         (key) => listItem[key] && regex.test(listItem[key] as string)
--- a/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
@@ -191,7 +191,7 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
 
     /** Modal Buttons (Save/Cancel) */
     const modalButtons = [
-      { label: "Cancel", onClick: closeModal },
+      { label: "Cancel", onClick: () => closeModal() },
       {
         label: "Save",
         onClick: () => saveUpdates(),
@@ -277,7 +277,7 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
           errorMessage: undefined,
         }));
         if (response && "status" in response && response.status === 200)
-          closeModal();
+          closeModal(true);
         setIsSaveInProgress(false);
       }, 2000);
     };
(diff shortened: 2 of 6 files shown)


morden35:
I think this looks really slick after the 'Save'! Is there any way to retain the search text after 'Cancel'? I think that would be my preference but this is def not necessary

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -414,7 +414,12 @@ class AdminPanelStore {
     searchInput: string,
     searchByKeys: (keyof T)[]
   ) {
-    const regex = new RegExp(`${searchInput}`, `i`);
+    // Escape special characters in search input
+    const sanitizedSearchInput = searchInput.replace(
+      /[.*+?^${}()|[\]\\]/g,
```
morden35, line 419: do you mind explaining this change a little? Does this replace any of the following chars `/[.*+?^${}()|[\]\\]/g` with `\\$&`?

mxosman: Absolutely! You're on the right track! 

The second param of the `replace` function is a little funky because it can be a string or a special replacement pattern (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement). 

In this case ("\\&$"), the first `\` backslash escapes the second backslash, the second backslash is a proper backslash, and the `$&` inserts the matched substring. So as a whole, it'll match any of those characters that you wrote out and replace them with an escape character `\<the matched character>` so that all of those characters can be properly escaped. [LMK if I just confused you more!]

I ran into this because I accidentally typed `\` in the search input and everything crashed with an error about the backslash. Oddly enough, it was just the backslash that knocked it down, it was OK with all of the other characters (and I figured let's escape them all just to be extra safe).

mxosman:
> I think this looks really slick after the 'Save'! Is there any way to retain the search text after 'Cancel'? I think that would be my preference but this is def not necessary

Thanks, Michelle! For sure - great call!

https://github.com/Recidiviz/justice-counts/assets/59492998/a9ba8510-98e0-4c34-acb4-2885f340720c

publisher/src/stores/AdminPanelStore.ts
mxosman: > Absolutely! You're on the right track!
> 
> The second param of the `replace` function is a little funky because it can be a string or a special replacement pattern (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement - [scroll down to **Specifying a string as the replacement** if the link doesn't do so automatically]).
> 
> In this case ("\&$"), the first `\` backslash escapes the second backslash, the second backslash is a proper backslash, and the `$&` inserts the matched substring. So as a whole, it'll match any of those characters that you wrote out and replace them with an escape character `\<the matched character>` so that all of those characters can be properly escaped. [LMK if I just confused you more!]
> 
> I ran into this because I accidentally typed `\` in the search input and everything crashed with an error about the backslash. Oddly enough, it was just the backslash that knocked it down, it was OK with all of the other characters (and I figured let's escape them all just to be extra safe).

morden35: woah so cool! thank you!

morden35 (approved):
thank you for the change! looks awesome

PR #1122, opened 2023-12-27
[Publisher] Add Sentry Logging

### LINKED ISSUE #1059: [Publisher] Assess and add logging to better triangulate the issue Stephanie (CSG) is experiencing with forced logouts
Reference conversation: https://recidiviz.slack.com/archives/C04JL8A2ADB/p1700148721949489

mxosman:
## Description of the change

Sets up basic sentry logging and adds an exception listener in the `AuthStore.getToken` method to try to better triangulate an intermittent forced logout issue reported by a member of CSG.

I'm sure there's more fine-tuning we can do to the configurations as we move along and expand FE error logging, but this is just a start. Here's the log for the test error I threw:
https://recidiviz-inc.sentry.io/issues/4766635999/?environment=production&environment=staging&project=4504532096516096&query=is%3Aresolved&referrer=issue-stream&statsPeriod=7d&stream_index=0

[P.S. the replay feature is super cool - shows you a replay of the users session before/during/after the error]

## Related issues

Closes #1059

### Development

### Code review

### DIFF
--- a/publisher/src/components/Auth/AuthStore.ts
+++ b/publisher/src/components/Auth/AuthStore.ts
@@ -21,6 +21,7 @@ import createAuth0Client, {
   GetTokenSilentlyOptions,
   User,
 } from "@auth0/auth0-spa-js";
+import * as Sentry from "@sentry/react";
 import { makeAutoObservable, runInAction } from "mobx";
 import qs from "qs";
 
@@ -145,6 +146,7 @@ export class AuthStore {
       }
       return Promise.reject(new Error("No auth client initialized."));
     } catch (error) {
+      Sentry.captureException(error);
       throw new Error(`Unable to retrieve token - ${error}`);
     }
   }
--- a/publisher/src/index.tsx
+++ b/publisher/src/index.tsx
@@ -33,6 +33,7 @@ import App from "./App";
 import AuthWall from "./components/Auth";
 import { StoreProvider } from "./stores";
 
+// Initialize Sentry
 Sentry.init({
   dsn: "https://3e8c790dbf0c407b8c039b91c7af9abc@o432474.ingest.sentry.io/4504532096516096",
   integrations: [
--- a/publisher/src/index.tsx
+++ b/publisher/src/index.tsx
@@ -18,13 +18,51 @@
 import "react-tooltip/dist/react-tooltip.css";
 
 import { GlobalStyle } from "@justice-counts/common/components/GlobalStyles";
+import * as Sentry from "@sentry/react";
 import React from "react";
 import { createRoot } from "react-dom/client";
-import { BrowserRouter } from "react-router-dom";
+import {
+  BrowserRouter,
+  createRoutesFromChildren,
+  matchRoutes,
+  useLocation,
+  useNavigationType,
+} from "react-router-dom";
 
 import App from "./App";
 import AuthWall from "./components/Auth";
 import { StoreProvider } from "./stores";
+
+Sentry.init({
+  dsn: "https://3e8c790dbf0c407b8c039b91c7af9abc@o432474.ingest.sentry.io/4504532096516096",
+  integrations: [
+    new Sentry.BrowserTracing({
+      routingInstrumentation: Sentry.reactRouterV6Instrumentation(
+        React.useEffect,
+        useLocation,
+        useNavigationType,
+        createRoutesFromChildren,
+        matchRoutes
+      ),
+    }),
+    new Sentry.Replay(),
+  ],
+
+  // Set tracesSampleRate to 1.0 to capture 100%
+  // of transactions for performance monitoring.
+  tracesSampleRate: 1.0,
+
+  // Set `tracePropagationTargets` to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: [
+    "localhost",
+    /^https:\/\/publisher\.justice-counts\.org\/api/,
+  ],
+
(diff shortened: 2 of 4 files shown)


morden35 (approved):
very cool!

publisher/src/index.tsx
```diff
@@ -18,14 +18,53 @@
...
+        createRoutesFromChildren,
+        matchRoutes
+      ),
+    }),
+    new Sentry.Replay(),
+  ],
+
+  // Set tracesSampleRate to 1.0 to capture 100%
+  // of transactions for performance monitoring.
+  tracesSampleRate: 1.0,
```
lilidworkin, line 54: Does this capture all traces or just those that cause an error? I'm not sure we want to capture all traces -- I think that could be expensive

publisher/src/index.tsx
```diff
@@ -18,14 +18,53 @@
...
+
+  // Set `tracePropagationTargets` to control for which URLs distributed tracing should be enabled
+  tracePropagationTargets: [
+    /^https:\/\/publisher-staging\.justice-counts\.org\/api/,
+    /^https:\/\/publisher\.justice-counts\.org\/api/,
+  ],
+
+  // Capture Replay for 10% of all sessions,
+  // plus for 100% of sessions with an error
+  replaysSessionSampleRate: 0.1,
```
lilidworkin, line 64: What would it look like to replay a session without an error?

lilidworkin (commented):
@mxosman this is super cool! I'm curious if you can expand a little on what we're capturing here. I think we should just be careful about how much we're capturing when errors don't occur, purely for expense reasons

publisher/src/index.tsx
lilidworkin: Is a replay a video?

publisher/src/components/Auth/AuthStore.ts
```diff
@@ -145,6 +146,7 @@ export class AuthStore {
       }
       return Promise.reject(new Error("No auth client initialized."));
     } catch (error) {
+      Sentry.captureException(error);
```
lilidworkin, line 149: Just want to double-confirm that this is the right place to put this -- do we catch this error up in the flow and redirect to the login screen?

mxosman:
> @mxosman this is super cool! I'm curious if you can expand a little on what we're capturing here. I think we should just be careful about how much we're capturing when errors don't occur, purely for expense reasons

Absolutely! So, for now it's just capturing the error that's triggering the forced logout in the `AuthStore`. It's not set up at the moment to capture all errors - we'd have to introduce an `ErrorBoundary` component and do some restructuring of all of our error handling to bubble things up to the ErrorBoundary to get broader coverage. This is just establishing a connection between Publisher & Sentry and capturing 1 error (assuming I set things up correctly 😅 ).

publisher/src/index.tsx
mxosman: Great call! I think this is all traces - we can set this value to zero because we're not doing performance monitoring at the moment. I don't think it would be picking up anything at the moment - I think there's a bit more set up to get this tracing part going.

publisher/src/index.tsx
mxosman: Yeah - my understanding of a replay is a video of the user's browser session where you can see mouse movements, interactions, etc. Here's a sample of the replay in this test error: https://recidiviz-inc.sentry.io/replays/1c1522441a14413ab5a87c376fa17ec2/?referrer=%2Fissues%2F%3AgroupId%2F&t=3&t_main=errors

publisher/src/components/Auth/AuthStore.ts
mxosman: Yeah, I think so! When I repro'd this, the root of the error was thrown from this method, and the error message came from `getTokenSilently` failing. I figured I'd start here and see where this leads - there weren't any other places that stood out to me to capture when I was digging. This is a tough one, but hopefully this will uncover more information I can work with!

mxosman: As far as the flow - once `getTokenSilently` throws an error, it's caught here in the `catch` statement, and re-thrown and handled in the `catch` statement in `API.ts` https://github.com/Recidiviz/justice-counts/blob/main/publisher/src/stores/API.ts#L132-L150

lilidworkin: Ah perfect, okay yep that's what I was wondering!

publisher/src/index.tsx
lilidworkin: Ok yep! I think let's change this to 0 then

publisher/src/index.tsx
lilidworkin: Cool yea -- let's set it to 0 to be safe!

lilidworkin:
> Absolutely! So, for now it's just capturing the error that's triggering the forced logout in the AuthStore. It's not set up at the moment to capture all errors

Okay great! Yea I think that's 100% fine and as intended -- let's just start with this and see how it goes!

lilidworkin (approved):
Good to go once the two non-error rates are set to 0! Thank you so much for setting up this integration!!

publisher/src/index.tsx
mxosman: Perfect timing - I just pushed up and saw this comment! Sounds great - setting both sample rates to 0!

mxosman:
Sounds great - thank you both so so much! Curious to see how this goes!

PR #1123, opened 2023-12-27
[Publisher] Prevent agencies from being able to have supervisions metrics that are both combined and disaggregated

mxosman:
## Description of the change

Prevents supervision + subsystems agencies from entering a state where both combined and disaggregated metrics are enabled.

https://github.com/Recidiviz/justice-counts/assets/59492998/2de1c109-47dd-404c-a770-be8ec9067546

Tooltips (all copy improvement suggestions welcome):
Screenshot 2023-12-27 at 1 45 27 PM
Screenshot 2023-12-27 at 1 45 13 PM

## Related issues

Closes [#25629](https://github.com/Recidiviz/recidiviz-data/issues/25629)

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
+++ b/publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
@@ -26,6 +26,7 @@ import {
   RadioButtonsWrapper,
 } from "@justice-counts/common/components/RadioButton";
 import { ToggleSwitch } from "@justice-counts/common/components/ToggleSwitch";
+import { Tooltip } from "@justice-counts/common/components/Tooltip";
 import {
   SupervisionSubsystems,
   SupervisionSystem,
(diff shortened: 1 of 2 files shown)


morden35:
Nice! Something I noticed from the video - if the combined metric already has breakdowns and definitions set but then is switched to disaggregated, it looks like we preserve those breakdown and definition settings (even though they are greyed out / disabled). I think the same happens in the inverse case for disaggregated metrics having things set prior to being flipped to combined. I'm not really sure what the ideal behavior is here (delete those previously set settings?), but wanted to call it out! lmk if that makes sense

publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
```diff
@@ -95,6 +96,23 @@ function MetricAvailability({ goToDefineMetrics }: MetricAvailabilityProps) {
...
+  const isSupervisionMetricAndDisaggregatedBySupervisionSubsystems =
+    disaggregatedBySupervisionSubsystems &&
+    systemSearchParam &&
+    !SupervisionSubsystems.includes(systemSearchParam);
+  const isSupervisionSubsystemMetricAndSubsystemsCombined =
+    !disaggregatedBySupervisionSubsystems &&
+    systemSearchParam &&
+    SupervisionSubsystems.includes(systemSearchParam);
+
+  const getDisaggregatedBySupervisionSubtypeTooltipMsg = () => {
```
lilidworkin, line 108: Love the copy!

lilidworkin (approved):
Awesome! 

I think the behavior @morden35 called out is okay with me! I think it's actually more of a feature than a bug, because if they flip back, their settings are preserved.

mxosman:
> Nice! Something I noticed from the video - if the combined metric already has breakdowns and definitions set but then is switched to disaggregated, it looks like we preserve those breakdown and definition settings (even though they are greyed out / disabled). I think the same happens in the inverse case for disaggregated metrics having things set prior to being flipped to combined. I'm not really sure what the ideal behavior is here (delete those previously set settings?), but wanted to call it out! lmk if that makes sense

Oh, good call! Hmm, I don't actually know if preserving them is intentional or not. I figured it might be if the metric is hidden from the records - so they can't add anything to those breakdowns. I'm not 100% sure!

publisher/src/components/MetricsConfiguration/MetricAvailability.tsx
mxosman: Oh good - I'm glad it's clear! Copy is always hard... esp. for this!

PR #1150, opened 2024-01-23
[Publisher][Bug] Fix encoding of agency names when linking to dashboards

### LINKED ISSUE #1144: [Bug] Agencies with a hyphen in their name are not linking properly to their respective Dashboard
If an agency has a hyphen in their name `-`, any links to their Agency Dashboard will result in a 404 and a toast that the agency with that name could not be found. This is because when encoding the URI on the frontend, we use a helper method called `slugify` which replaces all spaces with a hyphen `-`. Then, the backend decodes this to reconstruct the agency name by replacing the hyphens `-` with a space. If the agency's name has a hyphen included in it (e.g. "Agency - 123") - the frontend `slugify` method will convert that to `agency---123` and the backend will decode it to `agency   123` (with 3 spaces).

The solution would be to update the `slugify` function to use the native `encodeURI` api which converts spaces to `%20`, then update the backend to only replace the encoded spaces. This will ensure that hyphens are left alone.

mxosman:
## Description of the change

Fix encoding of agency names when linking to dashboards by using the native `encodeURI` function instead of the helper method `slugify`'s strategy of replacing spaces with hyphens.

This resolves an issue CSG noticed with agencies that have hyphens in their name. Currently, the FE uses a helper method called `slugify` that replaces spaces with hyphens, and the BE decodes the 'slugified' agency name by replacing the hyphens with spaces. Instead, the FE can use the `encodeURI` which will naturally replace spaces with a `%20` and the BE can decode encoded URIs and naturally decode a `%20` into a space and leave hyphens alone.

https://github.com/Recidiviz/justice-counts/assets/59492998/eae3c0aa-47bd-48c2-8b8e-e94218bcdd92

## Related issues

Closes #1144

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/Home/Home.tsx
+++ b/agency-dashboard/src/Home/Home.tsx
@@ -15,7 +15,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import { slugify } from "@justice-counts/common/utils";
 import { observer } from "mobx-react-lite";
 import React, { useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


lilidworkin (approved):
Love it!

PR #1179, opened 2024-02-02
[Publisher][Bug] Fix Race & Ethnicities grid displaying and saving incorrect enabled/disabled dimension states when set for the first time

mxosman:
## Description of the change

Fix Race & Ethnicities grid displaying and saving incorrect enabled/disabled dimension states when set for the first time.

Matt surfaced an issue where when setting the Race & Ethnicities dimensions for the first time via Metric Settings page, the saved results do not align properly with the expected output based on the user's input in the R&E modal form. This seems to only happen when the R&E dimensions have never been set before and are being set for the first time.

When doing a comparison to decide whether or not a particular dimension needs to be updated, It appears that the FE is not properly handling the BE sending `null` for these dimensions' initial state, and it's stumbles when it does a comparison of the current state (initially `null`) against the what the dimension state should be per the `gridState` (which only has boolean values). So, it's running into situations where it's doing a `null === false` check (`null` representing the current dimension state, and `false` representing the `gridState` - what the user's input should map to for that dimension). Remember `null` and `false` both mean the dimension is disabled - in these cases, we can safely coerce the `null` value to a boolean to get the desired comparison and resolve the erroneous output after initially configuring and saving.

Demo'ing before & after this change:

https://github.com/Recidiviz/justice-counts/assets/59492998/6374a61c-15dd-4397-b82c-36c3b469ad27

[Ack - it's still showing the `null` at the end there in the browser console because I forgot to add the Boolean coercion in the console.log statement - but it is indeed now a boolean! I just didn't have it in me to create a new agency again and find untouched R&E breakdowns and rerecord.]

## Related issues

Closes [#26990](https://github.com/Recidiviz/recidiviz-data/issues/26990)

### Development

### Code review

### DIFF
--- a/publisher/src/stores/MetricConfigStore.tsx
+++ b/publisher/src/stores/MetricConfigStore.tsx
@@ -1052,14 +1052,25 @@ class MetricConfigStore {
 
     Object.entries(racesStatusObject).forEach(([race, enabled]) =>
       ethnicities.forEach((ethnicity) => {
+        const isCurrentRaceEthnicityEnabled = Boolean(
+          ethnicitiesByRace[race][ethnicity].enabled
+        );
+        console.log(race, ethnicity);
+        console.log("enabled", enabled);
+        console.log(
+          "ethnicitiesByRace[race][ethnicity].enabled",
+          ethnicitiesByRace[race][ethnicity].enabled
+        );
+        console.log(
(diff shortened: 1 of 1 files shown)


brandon-hills (approved):
Nice!!

PR #1285, opened 2024-04-04
[Justice Counts] Update dashboard URLs to have agencyID 

### LINKED ISSUE #1284: [Non-Unique Agency Names] Include Agency ID in Dashboard URLs

nichelle-hall:
## Description of the change

This change is a prerequisite for supporting non-unique agency names. The agency dashboard URLS (and the respective queries to the BE) assume that all agencies have unique names. I update the URLs so that they include the agency IDs, which enables the BE to query by agency ID. This way, agencies can have the same name but distinct dashboard URLs. 

The following video is from me playtesting this branch with my [BE branch](https://github.com/Recidiviz/recidiviz-data/pull/28705)! 

https://github.com/Recidiviz/justice-counts/assets/19961693/e51151a4-155f-44b3-b50f-131716325b3a

## Related issues

Closes #1284 

### Development

### Code review

### DIFF
--- a/agency-dashboard/src/Protected/Protected.tsx
+++ b/agency-dashboard/src/Protected/Protected.tsx
@@ -30,27 +30,39 @@ export const Protected: React.FC<PropsWithChildren> = observer(
   ({ children }) => {
     const navigate = useNavigate();
     const { agencyDataStore, api } = useStore();
-    const { slug } = useParams();
+    const { agencyId, slug } = useParams();
     const isProductionEnv = api.environment === environment.PRODUCTION;
     const isDenied =
-      agencyDataStore.agency &&
-      agencyDataStore.agency.is_dashboard_enabled !== true;
-
+      (agencyDataStore.agency &&
+        agencyDataStore.agency.is_dashboard_enabled !== true) ||
+      !agencyId;
     const [loading, setLoading] = useState(true);
 
     useAsyncEffect(async () => {
       try {
-        await agencyDataStore.fetchAgencyData(slug as string);
-        setLoading(false);
+        if (!agencyId) {
+          setLoading(false);
+          showToast({
+            message: `No agency ID was specified in the URL path.`,
+            color: "red",
+            timeout: 4000,
+          });
+        } else {
+          await agencyDataStore.fetchAgencyData(
+            parseInt(agencyId),
+            slug as string
+          );
+          setLoading(false);
+        }
       } catch (error) {
         navigate("/404");
         showToast({
-          message: `No agency found with path /${slug}.`,
+          message: `No agency found with path ${agencyId}/${slug}.`,
           color: "red",
           timeout: 4000,
         });
       }
-    }, [slug]);
+    }, [agencyId, slug]);
 
     if (loading) {
       return <Loading />;
--- a/agency-dashboard/src/stores/AgencyDataStore.ts
+++ b/agency-dashboard/src/stores/AgencyDataStore.ts
@@ -196,13 +196,16 @@ class AgencyDataStore {
     return result;
   }
 
-  async fetchAgencyData(agencySlug: string): Promise<void | Error> {
+  async fetchAgencyData(
+    agencyId: number,
+    agencySlug: string
+  ): Promise<void | Error> {
     try {
       runInAction(() => {
         this.loading = true;
       });
       const response = (await API.request({
-        path: `/api/v2/agencies/${encodeURIComponent(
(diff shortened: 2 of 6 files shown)


morden35:
This looks good to me but will leave final review for @mxosman in case I'm missing some fe knowledge!

agency-dashboard/src/stores/AgencyDataStore.ts
```diff
@@ -196,13 +196,16 @@ class AgencyDataStore {
...
+  async fetchAgencyData(
+    agencyId: number,
+    agencySlug: string
+  ): Promise<void | Error> {
     try {
       runInAction(() => {
         this.loading = true;
       });
       const response = (await API.request({
-        path: `/api/v2/agencies/${encodeURIComponent(
```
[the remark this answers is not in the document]
nichelle-hall: At one point we had 2 versions of this endpoint! v2 was there for the most recent version of the endpoint.

agency-dashboard/src/Protected/Protected.tsx
```diff
@@ -40,17 +40,20 @@ export const Protected: React.FC<PropsWithChildren> = observer(
 
     useAsyncEffect(async () => {
       try {
-        await agencyDataStore.fetchAgencyData(slug as string);
+        await agencyDataStore.fetchAgencyData(
+          parseInt(agencyId as string),
```
mxosman, line 44: tiny nit: could also use `Number` since our IDs are numerical

I don't think you need to cast the `as string` here either - you should just be able to use `agencyId` directly

mxosman (commented):
Hi Nichelle! The only other things I can think of are updating Publisher links to the dashboard (off the top of my head I think we have a link in the profile dropdown menu and possibly the Help Center guides). Left a few comments, but otherwise looks great to me!

agency-dashboard/src/Protected/Protected.tsx
nichelle-hall: When I change it to a `as string` to `as number` I get a type error: `Argument of type 'number' is not assignable to parameter of type 'string; ` I think that useParams() interprets agencyId as either a `string  | undefined`. There's no other type that it will allow. I casted it as a string because parseInt throws a type error because it's only looking for a string and cannot take something is possibly undefined.  Is there a way to get around this?

mxosman: Oo - I see what's happening, I didn't realize `agencyId` could also be `undefined`. It's not likely that `agencyId` would be `undefined`, but since it's a possibility, the proper way to get around this would be to wrap this call in a conditional that checks for the presence of `agencyId`. I really liked your solution before - I felt like you were super close!

agency-dashboard/src/Protected/Protected.tsx
```diff
@@ -30,27 +30,40 @@ export const Protected: React.FC<PropsWithChildren> = observer(
...
-
+    const [isDenied, setIsDenied] = useState(
+      agencyDataStore.agency &&
+        agencyDataStore.agency.is_dashboard_enabled !== true
+    );
     useAsyncEffect(async () => {
       try {
-        await agencyDataStore.fetchAgencyData(slug as string);
-        setLoading(false);
+        if (typeof agencyId !== "string") {
```
mxosman, line 43: [Depending on whether or not you're going back to your old solution] - if you keep this, you can simplify it to `if (!agencyId) {` since anything you extract from `useParams` is `string | undefined`

agency-dashboard/src/Protected/Protected.tsx
```diff
@@ -30,27 +30,40 @@ export const Protected: React.FC<PropsWithChildren> = observer(
...
     const isProductionEnv = api.environment === environment.PRODUCTION;
-    const isDenied =
-      agencyDataStore.agency &&
-      agencyDataStore.agency.is_dashboard_enabled !== true;
 
     const [loading, setLoading] = useState(true);
-
+    const [isDenied, setIsDenied] = useState(
+      agencyDataStore.agency &&
+        agencyDataStore.agency.is_dashboard_enabled !== true
```
mxosman, line 39: I wouldn't put this in the initializer because it might initially `undefined` when the component mounts and that's the only time `useState` will ever reference it.

When it's put inline like before it is in-sync with the mobX state and once the fetching is done and the store is updated, the component will re-render with the updates.

mxosman (approved):
Amazing - looks great to me, Nichelle! Thank you so much!

PR #1294, opened 2024-04-09
[Publisher][Superagencies] Save new child agencies data in AgencyStore

### LINKED ISSUE #1274: [Superagencies][1/n] Allow easier toggle between child agencies by saving new data in AgencyStore
## Introduction

Throughout Publisher, there is currently no easy way for superagencies to toggle between their child agencies. You can always use the Agency Menu dropdown in the top left corner to switch agencies, but that lets you switch between _all_ of the agencies you belong to, not just the child agencies of a particular superagency.

We'd like to add dropdowns on a few pages that let superagencies easily switch between their child agencies, and refresh the current page accordingly.

To do this, we need to start fetching and storing information about which child agencies belong to a given superagency, so we can populate this dropdowns appropriately.

## Task 1: Use APIs to put this data in AgencyStore

The backend returns information about a superagency's child agencies in two places: in the API that powers the home page of the app, namely `/home/<agency_id>`, and also in a new dedicated API, `api/agencies/<super_agency_id>/children`, that is not called anywhere yet. If the <agency_id> that is passed in is for a superagency, then both APIs return a JSON blob with a "child_agencies" key that looks like this:

```
{
  "child_agencies": [
    {
        "name": name,
        "id": id,
        "systems": <list of systems>,
    },
    ...
  ]
}
```

This task is start using this information to populate and persist (for the user's session) a new dictionary in our AgencyStore (e.g. `{ <super_agency_name>: [<child agencies list>] }`). 
1) When the user opens the app for the first time, if we're on the home page, we should get the information for free from the existing `/home/<agency_id>` endpoint; if we load the app on a different page (not hitting the `/home` route), and the current agency is a superagency, we should call the new endpoint `api/agencies/<super_agency_id>/children`(after checking that we don't this info already stored in the AgencyStore)
2) When the user uses the dropdown in the top left corner to switch agencies, and switches to a superagency, then we should also update the store by recalling the appropriate endpoint

nasaownsky:
## Description of the change

Added functionality to save new child agencies data in AgencyStore no matter where we are in the app.

## Type of change

## Related issues

closes #1274 

### Development

### Code review


nasaownsky:
Hello @mxosman! So reading through the task description and trying several approaches I ended up with this one. How do you think, does it suits our requirements? Now if the agency is a superagency we made a request to agencies API and store our child agencies in `AgencyStore`, and also could access this value no matter where we are in the app.

publisher/src/stores/AgencyStore.ts
```diff
@@ -50,6 +52,8 @@ class AgencyStore {
 
   daysAfterTimePeriodToSendEmail: number | null;
 
+  childAgencies: ChildAgency[];
```
mxosman, line 55: Since a user can have multiple agencies and switch between them, I would make this an object w/ the `superagencyId`s as key and the list of child agencies as value.

That way if a user switches back to an agency that's already loaded a child agency, we can avoid making another call to get the list, and just use the one already stored.

publisher/src/components/Menu/Menu.tsx
```diff
@@ -194,6 +194,13 @@ const Menu: React.FC = () => {
     }
   }, [windowWidth]);
 
+  useEffect(() => {
+    const superagencyId = userStore.isAgencySuperagency(agencyId)
+      ? agencyId
+      : undefined;
+    agencyStore.updateChildAgencies(superagencyId);
+  }, [agencyId, agencyStore, userStore]);
```
mxosman, line 202: Good call on placement here! I can't think of other alternative areas that would satisfy this and keep up with switching agencies.

I would make sure that we don't already have the list of child agencies - if we do, we probably shouldn't call this again.

publisher/src/components/Menu/Menu.tsx
```diff
@@ -194,6 +194,13 @@ const Menu: React.FC = () => {
     }
   }, [windowWidth]);
 
+  useEffect(() => {
```
mxosman, line 197: Small comment on this useEffect explaining that we're getting and storing a list of child agencies if the user is in a superagency.

publisher/src/stores/AgencyStore.ts
nasaownsky: @mxosman Hmm, my idea was that we store in this variable only child agencies of user's currently viewed agency (I assume user can only view one agency at given time?). So if the viewed agency is a superagency - the endpoint would be called and store child agencies, and if it's not -  then child agencies would be reseted to empty array. So that way at any given time we could get this variable and see if there are child agencies in currently viewed agency. I thought that this would make it easier to work with child agencies in the moment, because if, for example, we need to display a drop-down list with them, we do not need to look for them in a large object, but can immediately take and work with them. What do you think about it?

mxosman: >(I assume user can only view one agency at given time?)
Ah - sorry we never clarified this, but some users can belong to and view multiple agencies. I was mostly concerned with the repeated calls to retrieve the child agencies each time someone switches agencies. But, now that I think about it, I think it would only be us the JC admins who would run into this - it doesn't seem likely (though it is always possible) that the user would belong to multiple superagencies and be toggling between them to make this an actual concern.

I take back my feedback - thank you for thinking through this with me!

publisher/src/components/Menu/Menu.tsx
mxosman: Disregard this one as well.

PR #1295, opened 2024-04-10
[Publisher][Superagencies] Add dropdown for child agencies to Set Up Metrics page

nasaownsky:
## Description of the change

1. Added dropdown UI for child agencies to Set Up Metrics page

## Type of change

> All pull requests must have at least one of the following labels applied (otherwise the PR will fail):

| Label                       	| Description                                                                                               	|
|-----------------------------	|-----------------------------------------------------------------------------------------------------------	|
| Type: Bug                   	| non-breaking change that fixes an issue                                                                   	|
| Type: Feature               	| non-breaking change that adds functionality                                                               	|
| Type: Breaking Change       	| fix or feature that would cause existing functionality to not work as expected                            	|
| Type: Non-breaking refactor 	| change addresses some tech debt item or prepares for a later change, but does not change functionality    	|
| Type: Configuration Change  	| adjusts configuration to achieve some end related to functionality, development, performance, or security 	|
| Type: Dependency Upgrade      | upgrades a project dependency - these changes are not included in release notes                             |

## Related issues

closes #1275 

### Development

**This box MUST be checked by the submitter prior to merging**:
- [ ] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [ ] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #1310, opened 2024-04-22
[Publisher][Superagencies] Add dropdown for child agencies to Explore Data page

nasaownsky:
## Description of the change

1. Added separate ChildAgenciesDropdown component 
2. Added ChildAgenciesDropdown to explore data page

## Type of change

> All pull requests must have at least one of the following labels applied (otherwise the PR will fail):

| Label                       	| Description                                                                                               	|
|-----------------------------	|-----------------------------------------------------------------------------------------------------------	|
| Type: Bug                   	| non-breaking change that fixes an issue                                                                   	|
| Type: Feature               	| non-breaking change that adds functionality                                                               	|
| Type: Breaking Change       	| fix or feature that would cause existing functionality to not work as expected                            	|
| Type: Non-breaking refactor 	| change addresses some tech debt item or prepares for a later change, but does not change functionality    	|
| Type: Configuration Change  	| adjusts configuration to achieve some end related to functionality, development, performance, or security 	|
| Type: Dependency Upgrade      | upgrades a project dependency - these changes are not included in release notes                             |

## Related issues

closes #1276 

### Development

**This box MUST be checked by the submitter prior to merging**:
- [ ] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [ ] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #1313, opened 2024-04-24
Revert "[Publisher][Superagencies] Add dropdown for child agencies to Set Up Metrics page"

mxosman:
Reverts Recidiviz/justice-counts#1295

### DIFF
--- a/publisher/src/components/Menu/Menu.tsx
+++ b/publisher/src/components/Menu/Menu.tsx
@@ -42,9 +42,6 @@ const Menu: React.FC = () => {
   const [isMobileMenuOpen, setIsMobileMenuOpen] = useState(false);
 
   const pathWithoutAgency = removeAgencyFromPath(location.pathname);
-  const userSuperagency = userStore.userAgencies?.find(
-    (agency) => agency.is_superagency
-  );
   const currentAgency = userStore.getAgency(agencyId);
   const hasDashboardEnabled = currentAgency?.is_dashboard_enabled;
   const agencyName = currentAgency?.name;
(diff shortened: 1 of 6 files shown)


brandon-hills (approved):

publisher/src/components/Menu/Menu.tsx
```diff
@@ -199,9 +196,11 @@ const Menu: React.FC = () => {
 
   /* Here we're getting and storing a list of child agencies if the user is in a superagency. * */
   useEffect(() => {
-    const superagencyId = userSuperagency && String(userSuperagency.id);
+    const superagencyId = userStore.isAgencySuperagency(agencyId)
+      ? agencyId
+      : undefined;
     agencyStore.loadChildAgencies(superagencyId);
-  }, [agencyStore, userSuperagency]);
```
mxosman, line 204: I totally missed this, but we should keep `agencyId` in the dependency array since we want this to run anytime the user switches agencies, and only run the logic when it is a superagency - otherwise, this will not re-run as it's finding the first superagency in the user's list of agencies and sticking with that the rest of the lifecycle.

publisher/src/components/Menu/Menu.tsx
```diff
@@ -42,9 +42,6 @@ const Menu: React.FC = () => {
   const [isMobileMenuOpen, setIsMobileMenuOpen] = useState(false);
 
   const pathWithoutAgency = removeAgencyFromPath(location.pathname);
-  const userSuperagency = userStore.userAgencies?.find(
-    (agency) => agency.is_superagency
-  );
```
mxosman, line 47: This will find the first agency that's a superagency in a user's list of agencies. A user could technically belong to multiple superagencies - so, this would always reference the first one on their list. We should instead find the superagency that corresponds with the child agency (if we're in a child agency). There should be an available property in the agency object called super_agency_id, and a child agency should have a value set for this. We should use that for the superagency id and look up the appropriate superagency.

mxosman:
Hi @nasaownsky! So sorry, I had to revert #1295 and #1310 quickly as we deployed to production today and there was an issue with the list of child agencies referenced.

There are two areas that I missed in my reviews and in testing - which I commented on above if you wouldn't mind looking into them. 

[Apologies for the big reverts for these small areas, but I had to move quickly this morning!]

Issue #1257, opened 2024-03-21
[Publisher][Design Update] Agency Settings: Account Tab Modals (excluding Jurisdictions) (3/5)


mxosman:
Create/update modals in the Account Tab of the Settings page.

- Name & email - modals will need to be created w/ error state for email
- Agency name, description, URL - these have existing modals that need their designs/behavior updated
  - Note: we can skip the character count for the agency description for now
  - Note: we can skip the Unsaved Changes modal for now

The tricky part of this will be updating the existing `Modal` component in the `common` repo. The Modal should continue to allow for custom styling via passing in a child component (basically using the Modal as a wrapper), and also:
- include an X close button that takes in a callback function
- display without heading borders and more padding when the modal is not overflowing:
<img width="330" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejMvQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fca427242a05bdb495cfb66b78daae8fd3c401db/Screenshot%202024-03-21%20at%203.21.15%E2%80%AFPM.png" alt="Screenshot 2024-03-21 at 3.21.15 PM.png" />
- display with heading border and less padding when the modal is overflowing:
<img width="330" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMEQvQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9d1e85484da5d1126b76b238665dad68f6cbe2fa/Screenshot%202024-03-21%20at%203.21.25%E2%80%AFPM.png" alt="Screenshot 2024-03-21 at 3.21.25 PM.png" />

Figma Nodes:
- https://www.figma.com/file/dvX3yg7FMGJSWFaqHWrWgQ/Justice-Counts-Publisher?type=design&node-id=9418-3087&mode=design&t=6W8cjLHs8YaaZSd5-11
- https://www.figma.com/file/dvX3yg7FMGJSWFaqHWrWgQ/Justice-Counts-Publisher?type=design&node-id=9418-3093&mode=design&t=6W8cjLHs8YaaZSd5-11
[...]

PR #1327, opened 2024-05-06
[Publisher][Design Update] Agency Settings: Account Tab Modals (excluding Jurisdictions) (3/5) #1257

corypride:
## Description of the change

Created/updated modals in the Account Tab of the Settings page.

Created Name & email modals updated the existing Modal component in the common repo to allow for custom styling via using agencySettingsConfigs flag. Added an X close button. Added styling for email validation. Updated styling for Agency Description modal. 

https://github.com/Recidiviz/justice-counts/assets/121071048/c86f96a0-c414-40e9-b661-d7ba337edcd1

Refs# 1257

## Type of change

### Development

### Code review


publisher/src/components/Settings/AccountSettings.tsx
```diff
@@ -15,72 +15,174 @@
...
-      return userStore.updateUserNameAndEmail(nameUpdate, email);
+  const editModeAndTypeUpdate = () => {
+    setIsSettingInEditMode(!isSettingInEditMode);
+    if (editType === EditType.Email_edit) {
+      setErrorMsg(undefined);
     }
-    if (emailUpdate) {
-      return userStore.updateUserNameAndEmail(name, emailUpdate);
+    setEditType("");
+  };
```
mxosman, line 62: I'm a little confused by this function here - can you clarify what it should do?

mxosman (commented):
Hi @corypride! Thank you for all of your work on this - and for the PR description and video! Here's my initial set of feedback based on my review this morning - please let me know if anything I said is confusing or if I can clarify anything further!


There are also a couple of things easier shown in video (below) - 
* if I make an update and close out without saving, the update doesn't save, but the UI persists my update - if that makes sense.
* There's a faint visible background color when I hover over the X button

https://github.com/Recidiviz/justice-counts/assets/59492998/60a73e40-b6e2-4793-88c3-85f112e9e724

publisher/src/components/Settings/AccountSettings.tsx
```diff
@@ -15,72 +15,174 @@
...
+  const saveNameEmailChange = (nameUpdate?: string, emailUpdate?: string) => {
+    if (nameUpdate || emailUpdate) {
+      if (nameUpdate) {
+        editModeAndTypeUpdate();
+        return userStore.updateUserNameAndEmail(nameUpdate, email);
+      }
+      if (emailUpdate) {
+        if (validateEmail(emailUpdate)) {
```
mxosman, line 70: I'm not sure why and this is probably not where it's happening, but I can't seem to get an invalid email/error - can you double check?

publisher/src/components/Settings/AccountSettings.tsx
```diff
@@ -15,72 +15,174 @@
...
-            }}
-          />
-          <span>
-            <a href="./emailmodal">Edit</a>
-          </span>
-        </AccountSettingsInputsCol>
-      </AccountSettingsInputsWrapper>
-    </AccountSettingsWrapper>
+    <>
+      {isSettingInEditMode && editType === "name" && (
```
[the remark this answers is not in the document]
corypride: When I compile the logic like above, I get a parsing error.

https://github.com/Recidiviz/justice-counts/assets/121071048/2ea6dd01-2e39-414f-95d6-6de5773b2f03

common/utils/helperUtils.ts
```diff
@@ -269,14 +269,22 @@ export function replaceSymbolsWithDash(value: string) {
   return value.replace(/[^A-Za-z0-9]/g, "-").toLowerCase();
 }
 
-export const validateEmail = (email: string) => {
+export const validateEmail = (email: string | undefined) => {
+  if (email === undefined) {
+    return false;
+  }
```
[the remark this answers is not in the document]
corypride: Ok, So I understand that an email will always be defined as a string, in the sense that it will be a valid email that is returned from userStore.email, or an empty string (which is the initial value of the email const on the AccountSettings page. But where my logic came from is the fact that in the function saveNameEmailChange, the emailUpdate param is defined as string or undefined, and inside of the function we are checking this value to determine if it is valid before we save it. Since this is the value that we are checking, I had to allow for the undefined option in the validateEmail function, or else I will get an error. Although I did not write the saveNameEmailChange function, I can see that the params were defined in that way in order to have cleaner code and not have to write a saveName and saveEmail function. But in order to do so there had to be a case created where either nameUpdate or emailUpdate can be undefined because one one would be plugged into the function at a time. I hope this makes sense? So I guess what I am asking is how would you like me to handle this, separate the functions or use one of the expert tricks that you have up your sleeve?

publisher/src/components/Settings/AccountSettings.tsx
```diff
@@ -15,72 +15,174 @@
...
+        <Modal
+          title="Name"
+          description={
+            <AccountSettingsInputsWrapper agencySettingsConfigs>
```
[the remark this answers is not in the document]
corypride: I see what you mean that it is a border within a border. I designed it exactly according to the figma though. What I have done to rectify this in the latest commit is get rid of the focus highlighting so it is not visible. Let me know if you still want me to do something different.


https://github.com/Recidiviz/justice-counts/assets/121071048/7ca1afa4-aecc-4fa9-8586-f54a4c44ab52

publisher/src/components/Settings/AccountSettings.tsx
corypride: I also tried this on my Mac and Linux machines and got the same result

publisher/src/components/Settings/AccountSettings.tsx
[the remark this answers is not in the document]
mxosman: These video recordings are so helpful - I really appreciate you taking the time to record them, Cory! So, when you return JSX, React wants there to be at least one parent element wrapper.

React wants this:
```
...
return (
  <Wrapper>  // <---- this can be <Fragment></Fragment> or fragment shorthand <></> if the wrapper element doesn't need styling
     ... all other components
  </Wrapper>
)
```

It doesn't like this (no parent wrapper - elements are side by side):
```
return (
  <Wrapper>
     ... all other components
  </Wrapper>
  <OtherWrapper>
     ... all other components
  </OtherWrapper>
)
```

And, it doesn't like this:
```
return (
  {...logic within brackets without being wrapped around an element}
)
```

For your situation - try doing this:
```
return (
  <>
     {...logic within brackets without being wrapped around an element}
  </>
)
```

common/utils/helperUtils.ts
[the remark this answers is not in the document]
mxosman: I meant more specifically in L67 in `AccountSettings.tsx` - that conditional already checks to see if `emailUpdate` is defined. If the conditional in L67 evaluates to `true`, then we can always be certain that `emailUpdate` will be defined by the time it reaches the L68. 

I think if you revert this adjustment in the `helperUtils`, you shouldn't run into any errors in this area, but let me know if you do! 

[Otherwise, what you're saying makes perfect sense - the other option instead of updating the helper util would be to handle the `emailUpdate` being `undefined` in the `saveNameEmailChange` function, but I think I would've been fine with your adjustment here.]

publisher/src/components/Settings/AccountSettings.tsx
mxosman: Ahh - I see! I think I was expecting it to show up immediately after typing bad characters, not after hitting the `Save` button. Gotcha! Now, I'm seeing what you're seeing. I'm totally good with this b/c I don't think these details were specified anywhere - but, if it's not too tricky and time-consuming, one nice enhancement would be to get that feedback and error state immediately when typing and the error state goes away once you've entered a valid email. (If you want to tackle this - I'd do it to at the end after everything else - otherwise, we can make a future task for this)

common/components/Modal/Modal.tsx
```diff
@@ -65,7 +70,9 @@ export function Modal({
   }, []);
 
   const Portal = (
-    <Styled.OuterWrapper modalBackground={modalBackground}>
+    <Styled.OuterWrapper
+      {...{ modalBackground, onClickClose, agencySettingsConfigs }}
```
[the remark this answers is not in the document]
corypride: Good catch!

publisher/src/components/Settings/AccountSettings.tsx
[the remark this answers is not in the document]
corypride: Totally worked!!!!

common/components/Button/Button.styled.tsx
```diff
@@ -79,8 +91,9 @@ export const Button = styled.div<{
         return `color: ${palette.solid.darkblue}`;
       }
     }};
-    ${({ buttonColor, noHover }) => {
+    ${({ buttonColor, noHover, agencySettingsConfigs }) => {
       if (buttonColor) return "opacity: 0.8;";
+      if (agencySettingsConfigs) return "unset";
```
mxosman, line 96: `unset` is not being set to a property here, so there will just be a line with only the word `unset` in the final CSS. I would remove this if it's not doing anything you need, or set the value `unset` to a CSS property (e.g. like in L95 or L97)

publisher/src/components/Settings/AccountSettings.tsx
```diff
@@ -15,72 +15,174 @@
...
   const [name, setName] = React.useState<string>(userStore?.name || "");
+  const [isSettingInEditMode, setIsSettingInEditMode] =
+    React.useState<boolean>(false);
+  const [editType, setEditType] = React.useState<string>("");
+  const [isEmailValid, setIsEmailValid] = React.useState(true);
+  const [errorMsg, setErrorMsg] = React.useState<
+    { message: string } | undefined
+  >(undefined);
+  const onClickClose = () => {
+    setIsSettingInEditMode(!isSettingInEditMode);
```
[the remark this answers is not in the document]
mxosman: Giving this a bump.

publisher/src/components/Settings/AccountSettings.tsx
mxosman: Giving this a bump as this function is still not entirely clear to me.

publisher/src/components/Settings/AccountSettings.tsx
```diff
@@ -15,72 +15,174 @@
...
+}
 
 export const AccountSettings = () => {
   const { userStore } = useStore();
   const [email, setEmail] = React.useState<string>(userStore?.email || "");
   const [name, setName] = React.useState<string>(userStore?.name || "");
+  const [isSettingInEditMode, setIsSettingInEditMode] =
+    React.useState<boolean>(false);
+  const [editType, setEditType] = React.useState<string>("");
+  const [isEmailValid, setIsEmailValid] = React.useState(true);
```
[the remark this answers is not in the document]
mxosman: Though not ideal, I won't block on this as the emails should hopefully all be validated before saving, and this will likely adjust when we move towards immediate validation as the user types.

mxosman (commented):
Hi @corypride! I just played around with this and it looks and feels great! I left a few comments and bumped some old ones, but after those are addressed, I think this will be good to go!

common/components/Modal/Modal.styled.tsx
```diff
@@ -79,21 +79,42 @@ export const Icon = styled.img`
...
+  jurisdictionsSettingsConfigs?: boolean;
+}>`
   display: flex;
   flex-direction: column;
-  align-items: center;
+  align-items: ${({ centerText, jurisdictionsSettingsConfigs }) => {
+    if (centerText) {
+      return `align-items: center;`;
+    }
+    if (jurisdictionsSettingsConfigs) {
```
mxosman, line 114: One suggestion with passing in variables like `jurisdictionsSettingsConfigs` - it's more clear if this were instead something like `unsetTextAlignment` or self-describing in a way. The benefit is that it's clearer when you set `unsetTextAlignment` that it's doing that one particular thing - whereas `jurisdictionsSettingsConfigs` gives me an impression that it's a configuration object of sorts that's doing a bunch of things, and one has to look at the functionality to determine that it's just unsetting the alignment. 

Also, anyone else who is using this `Description` that wants to do the same thing (unsetting alignment) in another component that's not related to jurisdiction settings can just use a descriptive prop (`unsetTextAlignment`) that's not specific to jurisdiction settings. LMK if that makes sense!

common/components/Modal/Modal.tsx
```diff
@@ -65,7 +74,7 @@ export function Modal({
   }, []);
 
   const Portal = (
-    <Styled.OuterWrapper modalBackground={modalBackground}>
+    <Styled.OuterWrapper {...{ modalBackground }}>
```
mxosman, line 77: nit: I'd revert this back to `modalBackground={modalBackground}`. It's only spreading one thing anyway.

common/components/RadioButton/RadioButton.styled.tsx
```diff
@@ -24,6 +24,7 @@ import { RadioButtonSize, WrapperSpacing } from "./types";
 
 export const RadioButtonsWrapper = styled.form<{
   spacing?: WrapperSpacing;
+  jurisdictionsSettingsConfigs?: boolean;
```
mxosman, line 27: It doesn't look like this is being used. You can remove this.

publisher/src/components/Settings/AgencySettingsDescription.tsx
```diff
@@ -16,27 +16,26 @@
 // =============================================================================
 
 import { Button } from "@justice-counts/common/components/Button";
-import { Input } from "@justice-counts/common/components/Input";
+import { NewInput } from "@justice-counts/common/components/Input";
+import { Modal } from "@justice-counts/common/components/Modal";
+// import { info } from "autoprefixer";
```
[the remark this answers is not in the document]
mxosman: > Can delete

This is still here btw.

publisher/src/components/Settings/AgencySettings.styles.tsx
```diff
@@ -24,7 +24,7 @@ import {
 } from "@justice-counts/common/components/GlobalStyles";
 import styled from "styled-components/macro";
 
-import searchIcon from "../assets/search-icon.png";
+// import searchIcon from "../assets/search-icon.png";
```
mxosman, line 27: You can delete this.

publisher/src/components/Settings/AgencySettings.styles.tsx
```diff
@@ -134,8 +134,8 @@ export const AgencySettingsBlock = styled.div<{
 export const AgencySettingsBlockTitle = styled.div<{
   isEditModeActive?: boolean;
   configured?: boolean;
+  agencySettingsConfigs?: boolean;
```
mxosman, line 137: It doesn't look like this is being used either - you can delete.

publisher/src/components/Settings/AgencySettings.styles.tsx
```diff
@@ -569,8 +557,13 @@ export const CheckboxWrapper = styled.div`
...
 
-export const Checkbox = styled.input`
-  appearance: none;
+export const CheckboxLabelWrapper = styled.div`
+  color: ${palette.solid.black};
+`;
+
+export const Checkbox = styled.input<{ squareCheckboxConfigs?: boolean }>`
+  appearance: ${({ squareCheckboxConfigs }) =>
+    squareCheckboxConfigs ? "auto" : "none"};
```
mxosman, line 566: I think we can use the `CheckboxOptions` component for consistency instead of adjusting the styled and using this? But, let me know if there's a reason that wouldn't work or I'm misunderstanding.

common/components/Modal/Modal.tsx
```diff
@@ -36,6 +37,10 @@ type ModalProps = Partial<{
   mediumTitle: boolean;
   customPadding?: string;
   children?: React.ReactNode;
+  onClickClose?: () => void | undefined;
```
mxosman, line 40: tiny nit: The `?` in `onClickClose` marks its type as also `undefined` - so no need for the extra `| undefined`.

common/components/Modal/Modal.tsx
```diff
@@ -75,24 +84,62 @@ export function Modal({
...
-                onClick={button.onClick}
-                borderColor={
-                  index === buttons.length - 1 ? undefined : "lightgrey"
-                }
-                buttonColor={
-                  index === buttons.length - 1
-                    ? primaryButtonColor()
-                    : undefined
-                }
+                label={<img src={xCloseLg} alt="X" />}
```
mxosman, line 96: Let's make the alt text "Close button" - that way a screen reader won't just say "X" but "Close button" which will be more useful

publisher/src/components/Settings/AgencySettingsURL.tsx
```diff
@@ -99,30 +113,37 @@ const AgencySettingsUrl: React.FC<{
...
-                label="Save"
-                onClick={handleSaveClick}
-                buttonColor="blue"
-              />
-            </EditModeButtonsContainer>
-          </>
+          <Modal
+            title="Agency URL"
+            description={
+              <AgencySettingsModalInputWrapperSmall error={!isURLValid}>
```
mxosman, line 119: Can you try something for me - what if instead of having an extra state for `isURLValid` and defaulting it to true, we instead only evaluate `urlText`?

So that means here, instead of `!isURLValid` we do `!validateAgencyURL(urlText)`? And I think we can remove all the `isURLValid` and `setIsURLValid` states. LMK if this doesn't work!

publisher/src/components/Settings/AgencySettings.styles.tsx
```diff
@@ -193,9 +184,22 @@ export const AgencySettingsInfoRow = styled.div<{ hasHover?: boolean }>`
...
+}>`
+  ${({ agencyDescriptionConfigs }) => {
+    if (agencyDescriptionConfigs) {
+      return (
+        `align-self: flex-start; margin-top: 12px; margin-bottom: 24px;` +
+        `${typography.body};` +
+        ` a {
+            text-decoration: none;
+            color: ${palette.solid.blue};
+          }`
```
mxosman, line 198: nit: you could write these in just one template string:

```
     return (`
        align-self: flex-start; 
        margin-top: 12px; 
        margin-bottom: 24px;
        ${typography.body};
         a {
            text-decoration: none;
            color: ${palette.solid.blue};
          }
     `)
```

mxosman (commented):
Hi @corypride! Thank you for all the work so far - let me know if you have any questions on the feedback! It might be easier to go through them in the `Files changed` tab so you can see the previous conversations as well. I'll need to do another pass as there were a lot of recent changes, but didn't want to delay the initial feedback. Please let me know once everything is ready to review again!

A couple of visual notes:

* The spacing here looks off compared to the Figma - can you take a look? 
  * Screenshot 2024-05-17 at 9 06 55 AM

* Can we remove the hover effect on these? 
  * Screenshot 2024-05-17 at 9 49 10 AM

publisher/src/components/Settings/AgencySettings.styles.tsx
corypride: I used checkbox because it was already being used on the AgencySettingsJurisdictions and Supervisions modals and the styling was already pretty much set consistent with the new styling of the figma (minus the size of the checkboxes which are supposed to be 16X16 and I need to change). I just added the CheckboxLabelWrapper to change the text color, and set the appearance to auto instead of none in order to remove the round checkboxes. Your call... change it or keep it?

mxosman: Thank you for explaining your thought process - it's not a huge blocker for me, but I would prefer to reuse the shared Checkbox component. It would keep things consistent and make it easier for future developers, if we decide to adjust all checkboxes, to update one spot instead of also finding custom checkbox components throughout the app to update as well.

mxosman (commented):
A few more notes - we're getting so close! 

Also - can you double check the hover state on each item in the "Exclude" section of the Jurisdictions modal? There's still a hover state there.

publisher/src/components/Settings/AgencySettings.styles.tsx
```diff
@@ -184,18 +174,33 @@ export const AgencySettingsInfoRow = styled.div<{ hasHover?: boolean }>`
   align-items: center;
 
   ${({ hasHover }) =>
-    hasHover &&
-    `&:hover {cursor: pointer; background-color: ${palette.highlight.grey2}}`}
+    hasHover
+      ? `&:hover {cursor: pointer; background-color: ${palette.highlight.grey2}}`
+      : `&:hover {cursor: pointer; background-color: none}}`};
```
mxosman, line 179: Instead of this ternary - we can keep it like it was before and just remove just remove the `background-color` entirely.

mxosman (commented):
Hi @corypride! Thank you for addressing all the blocking feedback - it's looking great! One small bug and a one-line deletion request - should be super quick and then this will be approved!

mxosman:
Once this is approved & merged in, let's sync up and do all the rebasing together and the final merge into main!

mxosman (approved):
Fantastic - thank you, sir! Great work on this, Cory!

PR #1341, opened 2024-05-22
[Publisher][Bug] Remove additional context logic from FormStore and DataEntryFormComponents

mxosman:
## Description of the change

Remove additional context logic from FormStore and DataEntryFormComponents as contexts are no longer handled by the record page.

There was a bug reported in staging for agency 11 where opening up a monthly record resulted in a crash. The crash was a result of the record page validation of previously saved inputs attempting to delete an error on the `metricsValues` object (in the FormStore - this object keeps track of the metric values and error state for each metric value) for a metric that did not exist in this object.

How is this happening? 
I'm not entirely sure when this happened or if it's always been this way, but it appears that when you add a value to the additional context field in the Metric Settings definition for a top-level metric, save it, and then delete the value and save it (so an empty string is saved) -- the empty string slips through this check here, and gets unnecessarily validated through `this.updateContextValue`:

https://github.com/Recidiviz/justice-counts/blob/c7bea896e57415e894b3a31a6fa8a5bc75ded3db/publisher/src/stores/FormStore.ts#L87-L88

It's still unclear whether this bug was here the entire time, or whether there was a relatively recent change in how empty string contexts are saved and later retrieved from the BE. But, I think the right move here would be to avoid running unnecessary logic for contexts in the FormStore and record page all together because they are not used in any way on the record page as they now live within the Metric Settings definitions and are handled by the `MetricConfigStore`.

Steps to repro:
1. Switch to the commit before the temporary fix via `git switch e695ea5cda6c015e5982f86853cbd579b74e061c --detach`
2. Choose any agency in staging
3. Go to Metric Settings and choose an available metric (or make one available)
4. Update the top-level metric definition by adding text in the additional context box, then save
5. Go back to the same metric definition and delete everything in the additional context box, then save
6. Find a record with matching frequency as the metric you just edited
7. Observe the unglamorous crash, with the following similar message in the console: `TypeError: Cannot read properties of undefined (reading 'SUPERVISION_FUNDING') at updateFieldErrorMessage (FormStore.ts:341:1)`

I tested this using the agencies in our playbook, going through the repro steps and ensuring there are no issues in the record page, going through and adding data to the records, reloading, and publishing, and really any areas that involve the FormStore.

## Type of change

## Related issues

Closes Recidiviz/recidiviz-data#28666

### Development

### Code review

### DIFF
--- a/publisher/src/components/Reports/DataEntryForm.test.tsx
+++ b/publisher/src/components/Reports/DataEntryForm.test.tsx
@@ -223,25 +223,3 @@ test("expect negative number value to add field error (formErrors should contain
     rootStore.formStore.metricsValues[0].PROSECUTION_STAFF.error?.message
   ).toBe("Please enter a valid number.");
 });
-
-test("expect empty value in metric field to add field error when there is a value in other fields such as contexts", async () => {
-  render(
-    <StoreProvider>
-      <MemoryRouter initialEntries={[`/agency/0/${REPORTS_LOWERCASE}/0`]}>
-        <Routes>
-          <Route
-            path={`/agency/:agencyId/${REPORTS_LOWERCASE}/:id`}
-            element={<ReportDataEntry />}
-          />
-        </Routes>
-      </MemoryRouter>
-    </StoreProvider>
-  );
-
-  const labels = await screen.findAllByLabelText("Total Staff");
-  fireEvent.change(labels[0], { target: { value: "" } });
-
-  expect(
-    rootStore.formStore.metricsValues[0].PROSECUTION_STAFF.error?.message
-  ).toBe("You are also required to enter a value for this field.");
-});
--- a/publisher/src/components/Reports/DataEntryFormComponents.tsx
+++ b/publisher/src/components/Reports/DataEntryFormComponents.tsx
@@ -18,7 +18,6 @@
 import { Input } from "@justice-counts/common/components/Input";
 import {
   Metric,
-  MetricContext,
   MetricDisaggregationDimensions,
   MetricDisaggregations,
 } from "@justice-counts/common/types";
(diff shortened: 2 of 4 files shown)


publisher/src/components/Reports/DataEntryFormComponents.tsx
```diff
@@ -158,68 +157,3 @@ export const DisaggregationDimensionTextInput = observer(
     );
   }
 );
-
```
mxosman, line 161: This entire component is not being used anywhere.

publisher/src/components/Reports/DataEntryForm.test.tsx
```diff
@@ -223,25 +223,3 @@ test("expect negative number value to add field error (formErrors should contain
...
-    </StoreProvider>
-  );
-
-  const labels = await screen.findAllByLabelText("Total Staff");
-  fireEvent.change(labels[0], { target: { value: "" } });
-
-  expect(
-    rootStore.formStore.metricsValues[0].PROSECUTION_STAFF.error?.message
-  ).toBe("You are also required to enter a value for this field.");
-});
```
mxosman, line 247: Test no longer necessary as there are no context fields in the record pages.

publisher/src/stores/FormStore.test.tsx
```diff
@@ -137,31 +137,13 @@ test("disaggregation dimension value handler updates the disaggregation dimensio
...
-    "NUMBER",
-    true
-  );
-
-  expect(
-    formStore.contexts[0].PROSECUTION_STAFF.PROGRAMMATIC_OR_MEDICAL_STAFF.value
-  ).toEqual("100");
-
-  expect.hasAssertions();
-});
```
mxosman, line 156: FormStore no longer needs to worry about the contexts, and as a result, `updateContextValue` will be deprecated.

publisher/src/stores/FormStore.ts
```diff
@@ -82,20 +82,6 @@ class FormStore {
...
-            reportID,
-            metric.key,
-            context.key,
-            normalizeToString(context.value),
-            context.required,
-            context.type,
-            metric.enabled
-          );
-        }
-      });
```
mxosman, line 98: No contexts in the record page means we don't have to loop through them and update/validate their values.

publisher/src/stores/FormStore.ts
```diff
@@ -173,38 +158,7 @@ class FormStore {
...
-            return {
-              ...context,
-              value: sanitizeInputValue(
-                contexts?.[context.key]?.value,
-                context.value,
-                context.type
-              ),
-              error: contextError,
-            };
-          }),
```
mxosman, line 207: All of this logic is unnecessary since we are deprecating `updateContextValue` and have no interest in the values of the context in the record page.

publisher/src/stores/FormStore.ts
```diff
@@ -278,19 +232,6 @@ class FormStore {
...
-          this.contexts[reportID]?.[metric.key]?.[context.key]?.value,
-          context.value,
-          context.type
-        );
-
-        combinedMetricValues.contexts.push({
-          key: context.key,
-          value: contextValue,
-        });
-      });
```
mxosman, line 292: Cleaning up anywhere else that has context logic.

publisher/src/stores/FormStore.ts
```diff
@@ -490,45 +431,6 @@ class FormStore {
...
-      this.validate(
-        contextType,
-        updatedValue,
-        required,
-        reportID,
-        metricKey,
-        contextKey
-      );
-    }
-  };
```
mxosman, line 530: And finally - since we have no interest in anything context-related in the record page, we can safely deprecate this method.

helperbot-recidiviz:

lilidworkin (approved):
Thank you so much for digging into this so deeply, figuring out the best fix, and testing so thoroughly!

mxosman:
Thank you for the quick review, Lili!

PR #1351, opened 2024-05-30
[Publisher][Bug] Fix metric definitions display for agencies with supervision + other non-supervision sectors

### LINKED ISSUE #1350: [Publisher][Bug] Fix metric definitions for agencies with supervision + other non-supervision sectors

mxosman:
## Description of the change

Fix metric definitions display for agencies with supervision + other non-supervision sectors with similar metrics (e.g. Funding, Staff, etc.) that are disaggregated by subpopulation in the supervision sector.

Huge shout out to @lilidworkin's playbook, as I stumbled into this while testing another PR going through the various agencies. 

This one is super sneaky... If you are in an agency that has, for example, Law Enforcement & Supervision sectors - both of which have a Funding metric - and within the Supervision sector, you disaggregate the Funding metric by subpopulations; when you look at the Funding metric's definitions within the Law Enforcement sector, you'll unfortunately see the list of definitions for the first disaggregated supervision subpopulation, and NOT the list of Law Enforcement Funding metric definitions. This only affects multi-sector agencies w/ a supervision sector w/ subpopulations and with similar metrics that are disaggregated in the supervision sector.

This change adds a check for the active system before determining which system-metric key to use. I think originally when I worked on this, I must've mistakenly assumed that if you were in a different sector that checks like `agencySupervisionSubsystems` would come up empty, but this is not true because it's scanning all the sectors belonging to an agency and will always come up if an agency has a supervision agency w/ subsectors.

Before: 
Screenshot 2024-05-30 at 10 52 28 AM

After:
Screenshot 2024-05-30 at 11 05 28 AM

## Type of change

## Related issues

Closes #1350 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricsConfiguration/MetricDefinitions.tsx
+++ b/publisher/src/components/MetricsConfiguration/MetricDefinitions.tsx
@@ -19,6 +19,7 @@ import { Dropdown } from "@justice-counts/common/components/Dropdown";
 import { Tooltip } from "@justice-counts/common/components/Tooltip";
 import {
   AgencySystem,
+  AgencySystems,
   MetricConfigurationSettings,
   SupervisionSubsystems,
 } from "@justice-counts/common/types";
@@ -80,10 +81,13 @@ function MetricDefinitions() {
         : settingsSearchParams.system
     );
 
-  const activeSystemMetricKey = replaceSystemMetricKeyWithNewSystem(
-    systemMetricKey,
-    selectedSupervisionSubsystem as AgencySystem
-  );
+  const activeSystemMetricKey =
+    settingsSearchParams.system === AgencySystems.SUPERVISION
+      ? replaceSystemMetricKeyWithNewSystem(
+          systemMetricKey,
+          selectedSupervisionSubsystem as AgencySystem
+        )
+      : systemMetricKey;
   const activeDisaggregationKeys =
     disaggregations[activeSystemMetricKey] &&
     Object.keys(disaggregations[activeSystemMetricKey]);
--- a/publisher/src/components/MetricsConfiguration/MetricDefinitions.tsx
+++ b/publisher/src/components/MetricsConfiguration/MetricDefinitions.tsx
@@ -102,12 +102,9 @@ function MetricDefinitions() {
     if (!metricDefinitionSettings[activeSystemMetricKey]) return true;
     const metricSettings = Object.values(
       metricDefinitionSettings[activeSystemMetricKey]
-    ).reduce(
-      (acc, metricSetting) => {
-        return { ...acc, ...metricSetting.settings };
-      },
-      {} as { [settingKey: string]: Partial<MetricConfigurationSettings> }
-    );
+    ).reduce((acc, metricSetting) => {
+      return { ...acc, ...metricSetting.settings };
+    }, {} as { [settingKey: string]: Partial<MetricConfigurationSettings> });
     /** Top-level metric context key will always be "INCLUDES_EXCLUDES_DESCRIPTION" */
     const hasContextValue = Boolean(
       contexts[activeSystemMetricKey].INCLUDES_EXCLUDES_DESCRIPTION.value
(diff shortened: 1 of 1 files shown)


lilidworkin (commented):
Thanks so much for catching this, @mxosman ! Could you explain why this fixes it? I'm having trouble intuiting why the updated logic is the right fix. It's not clear to me why the `replace SystemMetricKey` logic should only be run if the system is Supervision?

mxosman:
> Thanks so much for catching this, @mxosman ! Could you explain why this fixes it? I'm having trouble intuiting why the updated logic is the right fix. It's not clear to me why the `replace SystemMetricKey` logic should only be run if the system is Supervision?

Absolutely! It might be helpful to bring in the few lines above it to add more context (this is before this change):

https://github.com/Recidiviz/justice-counts/blob/c7bea896e57415e894b3a31a6fa8a5bc75ded3db/publisher/src/components/MetricsConfiguration/MetricDefinitions.tsx#L76-L86

When a supervision system is disaggregated, we need to replace the `systemMetricKey` because it'll be something like `SUPERVISION-SUPERVISION_FUNDING` and in disaggregated-mode, we need it to be something like `PROBATION-PROBATION_FUNDING`.

The above chunk of code naively assumed it was OK to do a `replaceSystemMetricKeyWithNewSystem` on all `systemMetricKeys` (even if it was a no-op) because it trusted the `useState` check in L79-81 would cover all cases:
* If a supervision system was combined and not disaggregated (subsystems are disabled), the initial value of `selectedSupervisionSubsystem` would be 'SUPERVISION', and the `activeSystemMetricKey` would be the same as the `systemMetricKey`: `SUPERVISION-SUPERVISION_FUNDING`.
* If a supervision system was disaggregated, the initial value of `selectedSupervisionSubsystem` would be the first subsystem in the agency's list (e.g. 'PAROLE'), and the `activeSystemMetricKey` would spit out something like `PAROLE_PAROLE_FUNDING`
* If it was a non-supervision system, the initial value of `selectedSupervisionSubsystem` would be the current non-supervision system (for most cases - more below), and the `activeSystemMetricKey` would be the same as the `systemMetricKey`: `JAILS-JAILS_FUNDING` (for example).

What the L79-81 check does not cover is an agency with multiple systems that include a supervision agency w/ enabled subsystems for a metric. In those cases, the initial value of `selectedSupervisionSubsystem` will always be the first enabled subsystem in the agency system's list, because `agencySupervisionSubsystems` will have enabled subsystems for the metric - and will replace the `systemMetricKey` (e.g. LAW_ENFORCEMENT-LAW_ENFORCEMENT_FUNDING) with `PAROLE_PAROLE_FUNDING` (assuming Parole was the first supervision subsystem on the list of the agency's systems).

With this change, the replacement of the system metric key is now isolated to supervision systems and not run for everything. I'm not sure if this long explanation made any sense - please let me know if there are any areas I can clarify further or any thoughts on adjustments to make! 

In the meantime, I'll adjust the `selectedSupervisionSubsystem` state name to be more clear, and add a clarifying comment to help clarify this behavior.

lilidworkin (approved):
Oh wow that definitely makes sense, thank you for the detailed explanation! I like the idea of adding a short comment about this. Tysm for catching and fixing 🙏

mxosman:
> Oh wow that definitely makes sense, thank you for the detailed explanation! I like the idea of adding a short comment about this. Tysm for catching and fixing 🙏

Thank you so so much for your eyes as always, Lili!! 🙏

publisher/src/components/MetricsConfiguration/MetricDefinitions.tsx
```diff
@@ -93,9 +102,12 @@ function MetricDefinitions() {
...
       metricDefinitionSettings[activeSystemMetricKey]
-    ).reduce((acc, metricSetting) => {
-      return { ...acc, ...metricSetting.settings };
-    }, {} as { [settingKey: string]: Partial<MetricConfigurationSettings> });
+    ).reduce(
+      (acc, metricSetting) => {
+        return { ...acc, ...metricSetting.settings };
+      },
+      {} as { [settingKey: string]: Partial<MetricConfigurationSettings> }
+    );
```
mxosman, line 110: Nothing changed here - I guess prettier decided to adjust the formatting of this block.

mxosman: Ah - this needs to be `yarn lint --fixed` - I think I have a different version of prettier on my VS code that formats this particular file differently on save.

PR #1378, opened 2024-06-10
[Publisher] Persist current metric after switching to a different child agency via the dropdown

### LINKED ISSUE #1332: Persist current metric after switching to a different child agency via the dropdown
With the new drop down for superagencies in Explore Data, would it be possible for the metric that is currently viewed to be the same metric viewed when you switch between child agencies?

nasaownsky:
## Description of the change

Added initial logic for metric persistence

## Type of change

## Related issues

closes #1332 

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataViz/MetricsDataChart.tsx
+++ b/publisher/src/components/DataViz/MetricsDataChart.tsx
@@ -188,6 +188,9 @@ export const MetricsDataChart: React.FC = observer(() => {
 
   useEffect(() => {
     if (currentMetric && currentSystem) {
+      const isMetricEnabled = enabledMetrics.some(
+        (metric) => metric.key === currentMetric.key
+      );
       window.history.replaceState(
         "",
         "",
@@ -196,7 +199,7 @@ export const MetricsDataChart: React.FC = observer(() => {
       dataVizStore.setInitialStateFromSearchParams();
       setSettingsSearchParams({
         system: currentSystem,
-        metric: currentMetric.key,
+        metric: isMetricEnabled ? currentMetric.key : undefined,
       });
     }
     /**
--- a/publisher/src/components/DataViz/MetricsDataChart.tsx
+++ b/publisher/src/components/DataViz/MetricsDataChart.tsx
@@ -188,9 +188,6 @@ export const MetricsDataChart: React.FC = observer(() => {
 
   useEffect(() => {
     if (currentMetric && currentSystem) {
-      const isMetricEnabled = enabledMetrics.some(
-        (metric) => metric.key === currentMetric.key
-      );
       window.history.replaceState(
         "",
         "",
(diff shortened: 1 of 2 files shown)


nasaownsky:
Hi again, @mxosman! After tweaking agencies dropdown I came up with this solution. It persists metric pretty well, but the thing is - if metric is enabled for current child agency we viewing and not enabled for the one we want to select next - it will also persists. I tried various approaches, but so far had no success. Mostly I tried to find out if metric is enabled utilizing `agencyMetrics` and `metricsBySystem`, but the problem is that this metrics info will only update in the future, when we actually select child agency we want to visit. Maybe you have some ideas regarding this issue? I will appreciate any help at this point. Thank you so much!

mxosman:
Hi @nasaownsky! Thank you so much for working on this and for the comment. I'm working on a focused task and will follow up with you as soon as I break away!

mxosman:
I totally see what you're talking about - such a good point. We query the metrics based on the current agency... so there is no way to know in advance what the child agency's list of metrics will be until we switch to it and a request is made. Hmm... still chewing on this.

mxosman:
@nasaownsky - here's an idea. It looks like there's some logic where if the metric search param doesn't exist, it falls back on the first open metric. Maybe that will be useful in our case, and we don't need to know whether or not the metric is enabled in advance.

So there's a `useEffect` here that we might be able to use:

https://github.com/Recidiviz/justice-counts/blob/eb6bb4f75ab9585f044bb64d0802e1806e1b334e/publisher/src/components/DataViz/MetricsDataChart.tsx#L189-L208

I can't remember all of the details of why this is setup this way, but this runs every time there's a change in the metrics/system params to keep the data viz store and the URL in sync. I _think_ we can just do a check in L199 - where it's setting the search params for the metric - and see if the `currentMetric` metric is enabled. If it is enabled, we can keep using that key -- but if it's not enabled, we can pass `undefined` and hopefully it resolves to the first available metric on the next re-render.

LMK if this makes sense?

nasaownsky:
@mxosman Yes, totally worked, thank you! Please, test my changes and see if there are any issues they can cause!

publisher/src/components/DataViz/MetricsDataChart.tsx
```diff
@@ -196,7 +199,7 @@ export const MetricsDataChart: React.FC = observer(() => {
       dataVizStore.setInitialStateFromSearchParams();
       setSettingsSearchParams({
         system: currentSystem,
-        metric: currentMetric.key,
+        metric: isMetricEnabled ? currentMetric.key : undefined,
```
mxosman, line 202: I think you can check `currentMetric.enabled` here without the `some` check.

mxosman (approved):
> @mxosman Yes, totally worked, thank you! Please, test my changes and see if there are any issues they can cause!


Amazing! Thank you so much, @nasaownsky! Just left one small adjustment suggestion - otherwise it looks and works great!

PR #1409, opened 2024-06-26
[Publisher] Update Explore Data page to allow toggle between metric reporting frequencies

### LINKED ISSUE #1368: [Publisher] Toggle between metric reporting frequencies in "Explore Data" tab
Agencies want to be able to view both CY and FY data on the same chart in the Explore Data page. Rather than allowing mixed frequency data to show on the same chart, we want to add some sort of toggle on the Explore Data page so that agencies can switch back and forth between CY, FY, and monthly charts.

[Implementation Notes from @mxosman]

Reference: [Figma Design](https://www.figma.com/design/dvX3yg7FMGJSWFaqHWrWgQ/Justice-Counts-Publisher?node-id=11652-6610&t=IBLfdBRyZcHOFDoG-4)

* Quick note: all of the notes on both panels are relevant, but since Explore Data page is still in the old design, please reference the right panel for how it should look.
* Additional things included are:
    * updating the copy for one of the time period dropdown options ("All" to "All time")
    * updating the copy for text below the chart
    * removing the blue banner

Relevant files: [MetricsDataChart.tsx](https://github.com/Recidiviz/justice-counts/blob/main/publisher/src/components/DataViz/MetricsDataChart.tsx), [DatapointsStore.ts](https://github.com/Recidiviz/justice-counts/blob/main/publisher/src/stores/DatapointsStore.ts)

The dropdown will contain all of the unique frequencies extracted from all of the datapoints an agency produced.

Example: if an agency has recorded data for a monthly metric, calendar year metric (starting month in January), fiscal year metric (starting month in July), and an annual metric with a custom starting month (let's say March) -- the dropdown should include the following options: "Monthly", "Calendar Year", "Fiscal Year", "Annual: March".

Since we get a list of all datapoints, we can loop through them and group them in their respective unique frequency bucket and give users the option to filter the chart data by frequency.

Please connect with me (@mxosman) if you have any questions at all (or if I can help you ramp up)!

nasaownsky:
## Description of the change

Added initial copy and UI updates

## Type of change

## Related issues

closes #1368 

### Development

### Code review

### DIFF
--- a/common/components/DataViz/DatapointsView.tsx
+++ b/common/components/DataViz/DatapointsView.tsx
@@ -35,15 +35,19 @@ import {
 import Legend from "@justice-counts/common/components/DataViz/Legend";
 import { MetricInsights } from "@justice-counts/common/components/DataViz/MetricInsights";
 import {
+  datapointFilterByFrequency,
   sortDatapointDimensions,
   transformDataForBarChart,
   transformDataForMetricInsights,
 } from "@justice-counts/common/components/DataViz/utils";
 import {
+  Datapoint,
   DatapointsGroupedByAggregateAndDisaggregations,
   DataVizAggregateName,
   DataVizCountOrPercentageView,
   dataVizCountOrPercentageView,
+  DataVizFrequencyViewDisplayName,
+  dataVizFrequencyViewDisplayName,
   DataVizTimeRangeDisplayName,
   DataVizTimeRangesMap,
   DimensionNamesByDisaggregation,
@@ -65,9 +69,11 @@ type DatapointsViewProps = {
   timeRange: DataVizTimeRangeDisplayName;
   disaggregationName: string;
   countOrPercentageView: DataVizCountOrPercentageView;
+  frequencyView: DataVizFrequencyViewDisplayName;
   setTimeRange: (timeRange: DataVizTimeRangeDisplayName) => void;
   setDisaggregationName: (disaggregation: string) => void;
   setCountOrPercentageView: (viewSetting: DataVizCountOrPercentageView) => void;
+  setFrequencyView: (frequencyView: DataVizFrequencyViewDisplayName) => void;
   metricName: string;
   metricFrequency?: ReportFrequency;
   metricStartingMonth?: number;
@@ -121,9 +127,11 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
       timeRange,
       disaggregationName,
       countOrPercentageView,
+      frequencyView,
       setTimeRange,
       setDisaggregationName,
       setCountOrPercentageView,
+      setFrequencyView,
       metricName,
       metricFrequency,
       metricStartingMonth,
@@ -142,7 +150,7 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
     const [mobileFiltersVisible, setMobileFiltersVisible] =
       React.useState<boolean>(false);
 
-    const selectedData =
+    const aggregatedData =
       (disaggregationName !== NoDisaggregationOption &&
         Object.values(
           datapointsGroupedByAggregateAndDisaggregations?.disaggregations[
@@ -153,7 +161,18 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
       [];
 
     // all datapoints have annual frequency
-    const isAnnualOnly = !selectedData.find((dp) => dp.frequency === "MONTHLY");
+    const isAnnualOnly = !aggregatedData.find(
+      (dp) => dp.frequency === "MONTHLY"
+    );
+    // all datapoints have monthly frequency
+    const isMonthlyOnly = !aggregatedData.find(
+      (dp) => dp.frequency === "ANNUAL"
+    );
+
+    const selectedData = aggregatedData.filter((dp: Datapoint) =>
+      datapointFilterByFrequency(dp, frequencyView)
+    );
+
     const disaggregations = Object.keys(dimensionNamesByDisaggregation || {});
     const disaggregationOptions = [...disaggregations];
     disaggregationOptions.unshift(noDisaggregationOption);
@@ -258,6 +277,14 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
           onClick: () => setCountOrPercentageView(option),
           highlight: countOrPercentageView === option,
         }));
+
+      const frequencyDropdownOptions: DropdownOption[] =
+        dataVizFrequencyViewDisplayName.map((option) => ({
+          key: option,
+          label: `${option} Reporting Frequency`,
+          onClick: () => setFrequencyView(option),
+          highlight: frequencyView === option,
+        }));
       return (
         <DatapointsViewControlsContainer>
           <Dropdown
@@ -278,6 +305,14 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
             <Dropdown
               label={countOrPercentageView}
               options={countOrPercentageDropdownOptions}
+              size="small"
+              caretPosition="right"
+            />
+          )}
+          {!isMonthlyOnly && (
+            <Dropdown
+              label={`${frequencyView} Reporting Frequency`}
+              options={frequencyDropdownOptions}
               size="small"
               caretPosition="right"
             />
--- a/common/components/DataViz/DatapointsView.tsx
+++ b/common/components/DataViz/DatapointsView.tsx
@@ -176,28 +176,35 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
 
     const datapointsByFrequencyView = aggregatedData.reduce((acc, dp) => {
       const startDate = new Date(dp.start_date);
-      const startMonthName = startDate
-        .toLocaleString("en-US", { month: "long" })
-        .toUpperCase() as DataVizAnnualFrequencyView;
-      const view: DataVizFrequencyView =
+      const startMonthName = monthsByName[
+        startDate.getUTCMonth()
+      ].toUpperCase() as DataVizAnnualFrequencyView;
+      const dpFrequencyView: DataVizFrequencyView =
         dp.frequency === "MONTHLY" ? "MONTHLY" : startMonthName;
+
+      const hasMatchingMetricFrequencyView =
+        metricFrequencyView === dpFrequencyView;
 
       if (metricFrequencyView) {
         if (!acc[metricFrequencyView]) {
           acc[metricFrequencyView] = [];
         }
-        acc[metricFrequencyView].push(dp);
+        if (hasMatchingMetricFrequencyView) {
+          acc[metricFrequencyView].push(dp);
+        }
       }
 
-      if (!acc[view]) {
-        acc[view] = [];
+      if (!hasMatchingMetricFrequencyView) {
+        if (!acc[dpFrequencyView]) {
+          acc[dpFrequencyView] = [];
+        }
+        acc[dpFrequencyView].push(dp);
       }
-      acc[view].push(dp);
 
       return acc;
     }, {} as FrequencyDataMap);
 
-    const datapointsFrequencyView = Object.keys(
+    const datapointsFrequencyViews = Object.keys(
       datapointsByFrequencyView
     ) as DataVizFrequencyView[];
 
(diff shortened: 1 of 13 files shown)


nasaownsky:
Hi @mxosman! 

Based on research and your comments I ended up with [this](https://github.com/Recidiviz/justice-counts/pull/1409/commits/46371a5b3b82c28a0cac45b4813d82aeebc7801c) update. Is my understanding and solution right or am I still missing something?

I know it should be updated to hide some frequency view dropdown options based on available datapoints, but I just wanna share it as a proof of concept for you to check it. So please leave your thoughts on it and tell me if there is something I don't get right! Thank you!

mxosman:
> Hi @mxosman!
> 
> Based on research and your comments I ended up with [this](https://github.com/Recidiviz/justice-counts/pull/1409/commits/46371a5b3b82c28a0cac45b4813d82aeebc7801c) update. Is my understanding and solution right or am I still missing something?
> 
> I know it should be updated to hide some frequency view dropdown options based on available datapoints, but I just wanna share it as a proof of concept for you to check it. So please leave your thoughts on it and tell me if there is something I don't get right! Thank you!

Will take a look this morning - thank you for patience and for the proof of concept!

mxosman:
> Hi @mxosman!
> 
> Based on research and your comments I ended up with [this](https://github.com/Recidiviz/justice-counts/pull/1409/commits/46371a5b3b82c28a0cac45b4813d82aeebc7801c) update. Is my understanding and solution right or am I still missing something?
> 
> I know it should be updated to hide some frequency view dropdown options based on available datapoints, but I just wanna share it as a proof of concept for you to check it. So please leave your thoughts on it and tell me if there is something I don't get right! Thank you!

Your understanding and solution is spot on! This is exactly the behavior that we're looking for - thank you for thinking through this. It was tough to explain for some reason. Please let me know if you have any other questions as you continue through this!

nasaownsky:
Hello @mxosman! So I've filtered out frequency dropdown display names based on data availability, but there is one problem we need to sort out. 

I see in the mocks that custom starting month is displayed as "Annual: March". I assume that this means "Annual: `customStartingMonthName`", right? But in practice it can be that there are many different custom starting months in datapoints, and displaying them as separate dropdown options and then filtering out data could be too cumbersome (if that makes sense). 

What if instead we just display "Annual: Other" as dropdown option?User could see starting month on the chart anyways and it will be much easier for us to filter all this datapoints.

mxosman:
> Hello @mxosman! So I've filtered out frequency dropdown display names based on data availability, but there is one problem we need to sort out.
> 

> 


Hi @nasaownsky! 

> I see in the mocks that custom starting month is displayed as "Annual: March". I assume that this means "Annual: `customStartingMonthName`", right?

Yes, this is correct!

>  But in practice it can be that there are many different custom starting months in datapoints, and displaying them as separate dropdown options and then filtering out data could be too cumbersome (if that makes sense).
> What if instead we just display "Annual: Other" as dropdown option?User could see starting month on the chart anyways and it will be much easier for us to filter all this datapoints.

Unfortunately having an "Annual: Other" where all of the non-standard frequencies go isn't an option as the outlined behavior is a requirement for this feature. If for example an agency records data for a metric that is Annual (March) frequency and another Annual (May) -- they want to be able to see _only_ the Annual (March) datapoints in the chart -- or -- only the Annual (May) datapoints in the chart. Having an "Annual: Other" option will show mixed frequencies, which is not what they want.

As far as it being cumbersome, are there any hurdles you're running into that would be helpful to talk through/brainstorm? I was originally thinking that we could go through all of the datapoints and categorize them by their frequency (maybe via a reducer):

Rough sketch of structure
```
{
  MONTHLY: [... monthly datapoints],
  ANNUAL: {
    JANUARY: [... calendar year datapoints],
    MARCH: [... March starting month datapoints],
    JULY: [... fiscal year datapoints],
  }
}
```

From this, we have the info we need for the dropdowns, and depending on the state/selection, we refer to this object and render the datapoints. But, LMK if there's something I'm not considering. Happy to keep talking through this tricky part!

common/components/DataViz/DatapointsView.tsx
```diff
@@ -165,23 +203,32 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
...
 
+    useEffect(() => {
+      if (!isAnnualOnly) {
+        setFrequencyView("MONTHLY");
+      } else {
+        setFrequencyView(datapointsFrequencyView[0]);
+      }
+
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [metricName, isAnnualOnly]);
```
mxosman, line 214: One important thing is, this should by default be set to the current metric's frequency. So, if a metric like Expenses is set to Annual (March), for example, then that should be the default view when a user lands on the page.

common/components/DataViz/DatapointsView.tsx
```diff
@@ -152,7 +160,37 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
...
 
     // all datapoints have annual frequency
-    const isAnnualOnly = !selectedData.find((dp) => dp.frequency === "MONTHLY");
+    const isAnnualOnly = !aggregatedData.find(
+      (dp) => dp.frequency === "MONTHLY"
+    );
+    // all datapoints have monthly frequency
+    const isMonthlyOnly = !aggregatedData.find(
+      (dp) => dp.frequency === "ANNUAL"
+    );
```
mxosman, line 169: This might not be a big deal since we're handling the case where there are no datapoints in the BarChart component, but wouldn't these both be `true` if there are no datapoints at all? 

This is not a fully formed thought, so please take it with a grain of salt, but I'm wondering if it might be better to determine this after calculating the `datapointsByFrequencyView`?

mxosman (commented):
This is looking fantastic, @nasaownsky! Just a few more things left in the comments, and one quick thing re: the dates - it looks like the calculation to display "Calendar Year" and "Fiscal Year" are off - would you mind taking another look? The calendar and fiscal year are showing up in the custom annual format.

Screenshot 2024-07-16 at 9 09 38 AM

common/components/DataViz/DatapointsView.tsx
nasaownsky: Something like [this](https://github.com/Recidiviz/justice-counts/pull/1409/commits/f43efc6b81e2421afb068a83aa56e4d6dd1067e7)?

nasaownsky:
>  it looks like the calculation to display "Calendar Year" and "Fiscal Year" are off

After our discussion in issue thread I thought that we distribute frequencies by the starting month, as you say there:

> The ANNUAL frequency datapoints need the starting_month as well to group them appropriately (because you can have an ANNUAL frequency that begins in January (Calendar Year), or July (Fiscal Year) or some custom starting month like April.

So the dropdown displays options accordingly: if starting month is JANUARY - Calendar year, if JULY - Fiscal year, and if other month - Annual: Other month.

In your screenshot starting months are DECEMBER and JUNE, so the dropdown displays them as Annual: December and Annual: June respectively. 

If starting months are JANUARY and JULY it displays them appropriately:  

Снимок экрана 2024-07-17 в 13 22 30

common/components/DataViz/DatapointsView.tsx
nasaownsky: Hmm, I believe this will not have any impact on app behavior since we only use this values for filtering data, which is not the case where there is no data at all. In fact, with current logic I believe we want `isMonthlyOnly ` to be true -- in that case it will not display frequencies dropdown at all if there is no data.

mxosman:
Hi @nasaownsky! Yes - your understanding is correct! As I'm writing this and looking at your screenshot, it just hit me what the discrepancy might be... I think we need to use something standardized like `UTC` when calculating dates because our time-zones are different.

Here's a video and screenshots of my experience -- I'm also noticing that by default it's showing all of the frequencies:

https://github.com/user-attachments/assets/8d5d7cd7-df84-46d8-9914-aaa170928ae3

Screenshot 2024-07-17 at 10 30 24 AM

Screenshot 2024-07-17 at 10 30 33 AM

common/components/DataViz/DatapointsView.tsx
mxosman: Sounds great - thanks for thinking through this with me.

common/components/DataViz/DatapointsView.tsx
mxosman: Yeah - that seems like it's going in the right direction, but the behavior I'm seeing now shows all of the frequencies by default for some reason.

nasaownsky: @mxosman Hmm, now when I thinking about it I'm not sure -- what would be the expected behavior? It seems like the metric frequency is almost always different from datapoints frequencies (especially if it's annual frequency). So what data would be appropriate to display?

mxosman: It should match the current behavior in `main` - if you adjust the frequency of a metric in the Metric Settings, and go to the Explore Data page, it will only show datapoints that match the currently set frequency. 

This is a good agency to try it with b/c there's a Funding metric that has data for multiple frequencies (monthly, annual CY, annual FY, and annual: April): https://publisher-staging.justice-counts.org/agency/452 -- feel free to adjust the Funding metric frequency to any of the 4 listed above and go to the Explore Data page. LMK if this makes sense!

nasaownsky:
@mxosman thank you for your detailed explanations, it helps a LOT and I don't know what would I do without your help here! I updated the logic and switched to UTC, so please check my changes and if they work correctly.

common/components/DataViz/DatapointsView.tsx
```diff
@@ -165,23 +222,37 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
...
+        ].toUpperCase() as DataVizAnnualFrequencyView;
+
+        setFrequencyView(metricAnnualFrequencyView);
+        setMetricFrequencyView(metricAnnualFrequencyView);
+      } else {
+        setFrequencyView("MONTHLY");
+        setMetricFrequencyView("MONTHLY");
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [metricName, metricFrequency]);
```
mxosman, line 238: Would you mind adding a comment here to explain the default behavior - something like "Displays metric's current frequency datapoints by default"?

mxosman (commented):
> @mxosman thank you for your detailed explanations, it helps a LOT and I don't know what would I do without your help here! I updated the logic and switched to UTC, so please check my changes and if see they work correctly.

Oh, I'm so glad it's been helpful (and makes sense - haha, this one was tricky to explain) -- I don't know what I would do without YOUR help! You're doing the hardest parts of this very tricky feature.

The changes fixed the date behavior - yay - thank you for that! The only other thing I'm noticing that's off is when you switch the metric to `Monthly` frequency, and go back to Explore Data and switch to view a different frequency, the chart adds in monthly no-data bars. Probably easier explained with a video (I switch to monthly frequency in the second half):

https://github.com/user-attachments/assets/0d14c7d2-5f49-40bb-aa60-83930711658e

Aside from that, everything is looking great to me! Almost there, and then all you have to worry about is your relaxing vacation coming up soon. 😄

common/components/DataViz/DatapointsView.tsx
```diff
@@ -276,7 +280,7 @@ export const DatapointsView = forwardRef<never, DatapointsViewProps>(
             selectedData,
             selectedTimeRangeValue,
             countOrPercentageView,
-            metricFrequency,
+            frequencyView === "MONTHLY" ? "MONTHLY" : "ANNUAL",
             startingMonth
           )}
```
nasaownsky, line 285: The problem was that frequency was not changing in `transformDataForBarChart`, fixed that!

nasaownsky:
@mxosman Woohoo! I hope we finally made it! Right on time before I'm leaving 😄 Thank you again for your amazing cooperation with me on this task. Next time we'll meet in two weeks, and I look forward to work with you again on a new even more exciting and ambitious tasks!

lilidworkin:
@nasaownsky just chiming in to say thank you for your amazing work on all of these tasks!!! Have a wonderful and relaxing vacation and we'll see you in two weeks 😊

mxosman (approved):
AMAZING!!! Yay! I'm glad it was a tiny thing! Looks and works wonderfully! 

+1000 to everything Lili said, thank you so so much for all of your patience and incredibly hard work on all of these tasks. I don't know what we would do without your help. Hope you have an incredible and relaxing time off and don't think one bit about coding - haha. You will be missed in the meantime, but we all look forward to working with you again when you return!

nasaownsky:
Thank you so much Mahmoud and Lili! ❤️

PR #1421, opened 2024-07-03
[Publisher] refetch agencies every time that they are edited in the admin panel

nichelle-hall:
## Description of the change

This update updates the front-end functionality to refetch agency data after any updates to an agency via the admin panel. This ensures that any changes made to a super agency are accurately reflected in its child agencies ([BE PR](https://github.com/Recidiviz/recidiviz-data/pull/31147/files)).

https://github.com/Recidiviz/justice-counts/assets/19961693/2c566fbf-fd13-4f5b-bb18-17ba491c552b

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
@@ -310,7 +310,10 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
       }
 
       /** Refetch when changes are made to superagency or child agencies so all related agencies are updated */
-      const shouldRefetch = hasSuperagencyUpdate || hasChildAgencyUpdates;
+      const shouldRefetch =
+        hasSuperagencyUpdate ||
+        hasChildAgencyUpdates ||
+        hasTeamMemberOrRoleUpdates;
       const response = await saveAgencyProvisioningUpdates(shouldRefetch);
 
       setShowSaveConfirmation({
--- a/publisher/src/components/AdminPanel/UserProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/UserProvisioning.tsx
@@ -231,7 +231,11 @@ export const UserProvisioning: React.FC<ProvisioningProps> = observer(
           (id) => !deletedAgenciesIDs.has(id)
         ),
       ]);
-      const response = await adminPanelStore.saveUserProvisioningUpdates();
+
+      const shouldRefetch = hasAgencyUpdates;
+      const response = await adminPanelStore.saveUserProvisioningUpdates(
+        shouldRefetch
+      );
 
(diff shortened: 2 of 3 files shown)


mxosman (approved):
LGTM!

lilidworkin (commented):
Hold on @mxosman @nichelle-hall -- would like to understand this a bit better, since it seems like we were deliberately _not_ doing this previously and instead controlling it via a `refetch` parameter.

So why the change? And is it better just to isolate the place where we do need this, and pass a value of `refetch = true` in that case?

mxosman:
> Hold on @mxosman @nichelle-hall -- would like to understand this a bit better, since it seems like we were deliberately _not_ doing this previously and instead controlling it via a `refetch` parameter.
> 
> So why the change? And is it better just to isolate the place where we do need this, and pass a value of `refetch = true` in that case?

Hi Lili! Thanks for calling this out! Previously, we were setting the `refetch` to true only when there was an update to the superagency/child agency list so those changes propagate throughout all super/child agencies (the other option at the time was to add more logic to update the FE states of all of the child agencies/superagencies that were modified - but we opted just to refetch because it was simpler and gave us a reliable up-to-date copy of everything). 

Now that the need is for any updates to propagate to superagencies and child agencies, I felt comfortable with us widening the refetch to everything (despite it not being the most optimal solution) since that's probably the simplest way to ensure all the dependencies of your changes are reflected after your hit save. But, now that I'm thinking about this some more after your comment, if this change is motivated just by updating user information and having that propagate, we can totally keep the `refetch` logic as you suggested, and actually just update this flag below to include `hasTeamMemberOrRoleUpdates`:

https://github.com/Recidiviz/justice-counts/blob/4492eda02b1f1741d43398addf77467544375dc8/publisher/src/components/AdminPanel/AgencyProvisioning.tsx#L313

I agree this is a better way to handle this - instead of refetching everything, continue to isolate it until we reach the point where we need every update to propagate. Sorry, I didn't think this through all the way when I was reviewing.

[And, if we do eventually remove the `refetch`, we should remove it from everything (the param in the save function and the AgencyProvisioning component)]

mxosman (changes requested):
Gently rescinding my approval in favor of Lili's consideration & suggestion. Sorry for the hasty approval!

lilidworkin:
Awesome, thank you for taking another look at this @mxosman and for the super detailed explanation! That path forward makes perfect sense to me, appreciate it.

nichelle-hall:
Updated video!

https://github.com/Recidiviz/justice-counts/assets/19961693/9e595296-6b74-400a-955a-a8cbda192dc9

mxosman (approved):
This looks good to me, Nichelle! Thank you for the adjustments and the video - and thank you for the recommendation, Lili!

PR #1488, opened 2024-09-05
[Publisher] Filter out disabled metrics and dimensions

### LINKED ISSUE #1420: Filter out data for disabled metrics and breakdowns in the Review Page and Explore Page
A user can enter data for a metric or breakdown, and then disable that metric or breakdown. An example is [here](https://publisher-staging.justice-counts.org/agency/148/records/3685). Scroll down to the Open Cases metric. Note that the "Other Caseload" and "Unknown Caseload" breakdowns are disabled, but someone previously entered a value of 24 for each:

Screenshot 2024-07-02 at 11 21 54 AM

When you click "Review" in the top right corner, you see these values on the Review Page:

Screenshot 2024-07-02 at 11 23 03 AM

Similarly, if you to go the [Explore Page ](https://publisher-staging.justice-counts.org/agency/148/data?system=defense&metric=defense_caseloads_people&disaggregation=Case+Severity+Type), you see these values when you hover over the bar for January:

Screenshot 2024-07-02 at 11 23 43 AM

Same thing happens if you disable an entire metric. Note that for this agency, the Cases Disposed metric is disabled (you see this if you scroll to the bottom of the January 2023 report) but its data still appears on the Review Page. It doesn't appear on the Explore page though -- so that part is working properly.

We should filter out data for disabled metrics or breakdowns on both the Review Page and the Explore Page!

nasaownsky:
## Description of the change

1. Filtered disabled metrics from review page

## Type of change

## Related issues

closes #1420 

### Development

### Code review


nasaownsky:
@mxosman Phew, this one is tricky! I was digging through the codebase and the managed to get the solution for disabled metrics on the review page. But with dimensions it's more complicated: I see that dimensions filtered in `ReportStore` on line 303 (at least for the review page), and if we follow the data path for this value we end up in `getMultipleReportsWithDatapoints` method, where we get response from `reports?agency_id...` endpoint. It contains `datapoints`, but without `enabled` property, so my question is: how hard it would be to include this property to this response? Because throughout the app I see that there are dimensions objects with and without `enabled` property in different places, so it is a little bit confusing for me. 

And if it is impossible, maybe you have some ideas how else could I filter out this disabled dimensions? I see that you written the majority of code on this feature, so maybe you have better understanding of what's going on there 😅

mxosman:
> @mxosman Phew, this one is tricky! I was digging through the codebase and managed to get the solution for disabled metrics on the review page. But with dimensions it's more complicated: I see that dimensions filtered in `ReportStore` on line 303 (at least for the review page), and if we follow the data path for this value we end up in `getMultipleReportsWithDatapoints` method, where we get response from `reports?agency_id...` endpoint. It contains `datapoints`, but without `enabled` property, so my question is: how hard it would be to include this property to this response? Because throughout the app I see that there are dimensions objects with and without `enabled` property in different places, so it is a little bit confusing for me.
> 
> And if it is impossible, maybe you have some ideas how else could I filter out this disabled dimensions? I see that you written the majority of code on this feature, so maybe you have better understanding of what's going on there 😅

Hi @nasaownsky! Yeah - I figured this was going to be tricky... Let me follow up with you on all of this when I'm able to properly context switch!

mxosman:
> @mxosman Phew, this one is tricky! I was digging through the codebase and managed to get the solution for disabled metrics on the review page. But with dimensions it's more complicated: I see that dimensions filtered in `ReportStore` on line 303 (at least for the review page), and if we follow the data path for this value we end up in `getMultipleReportsWithDatapoints` method, where we get response from `reports?agency_id...` endpoint. It contains `datapoints`, but without `enabled` property, so my question is: how hard it would be to include this property to this response? Because throughout the app I see that there are dimensions objects with and without `enabled` property in different places, so it is a little bit confusing for me.
> 
> And if it is impossible, maybe you have some ideas how else could I filter out this disabled dimensions? I see that you written the majority of code on this feature, so maybe you have better understanding of what's going on there 😅

Thanks for doing all that sleuthing, @nasaownsky! I'm digging back in to get more context as I don't fully remember how it all connects. But, I do know that the datapoints not having an `enabled` indicator has made things like this super challenging. I don't imagine it would be difficult to do this. Let me dig into the backend and see whether or not it would be straight forward to surface this.

mxosman:
OK - it's not feasible to do due to BE complexity in our current structure. There must be a way where we can take the info from the datapoints (like the metric or dimension key) and query either the metric setting store or report store for the metric. Can you see which of those stores would be best for the component to query from? And when we do, is there some helper method or something we can come up with to make these queries easier (datapoint -> enabled status)?

nasaownsky:
@mxosman Hmm.. I was able to use `FormStore` to filter disabled dimensions out of Review page. But with Explore Data page it is kind of different flow. I assume that I should filter dimensions on Explore Data page somewhere near `BaseDatapointsStore`, right? But I also assume that I can't access `FormStore` there, at least I tried that and failed. Maybe there are some dimensions stuff near that `BaseDatapointsStore` that you know of?

mxosman:
Hi @nasaownsky - I very much appreciate your patience with my slow response. Getting things ready for a launch next week but will respond as soon as I can!

mxosman:
Planning on ramping back up into this today - thank you for your patience!

mxosman:
OK - still haven't gone all the way through, but dug into the `DatapointsStore` -- here's my quick incomplete thought:

We might have an opportunity here to access the report store metrics where we can find the `enabled` property:
https://github.com/Recidiviz/justice-counts/blob/8307a2d20f16a06c81ae20f75cf760cb1da530fe/publisher/src/stores/DatapointsStore.ts#L81-L83

It might be helpful to create a helper function similar to `getMetricKeyToFrequencyMap` and either store a mapping of the metrics/disaggregations/dimensions enabled status in the store so it's accessible anytime we get datapoints from this path OR we can inject it to each rawDatapoint so we always have it available within each datapoint object.

Now - for the publish review page, I might need to spend more time thinking, but my naive impression is that we can make a base helper function both stores can use that can take in datapoints, and check against the `ReportStore` metrics' enabled status and use that on `combinedFilteredDatapointsFromAllReports`:

https://github.com/Recidiviz/justice-counts/blob/8307a2d20f16a06c81ae20f75cf760cb1da530fe/publisher/src/stores/ReportStore.ts#L255-L257

LMK if this makes sense/where I can clarify further! It's so painful sometimes to look through my old code - thank you for working through the abstractions my past self setup.

nasaownsky:
@mxosman Thank you for your research on this, I'm sure we will figure something out together!!

1. As for publish review page - I actually already figured that out and implemented solutions -- please check them out! Or do you think this solutions are much inconvenient for us to use? LMK your thoughts!

2. I was thinking about how we should implement filtering on Explore data page and it seems to me that I have come up with a very interesting observation. Let's take for example Defense' (148) agency January 2023 record: we will see that Open Cases metric has total value (120) which consists of three ENABLED dimensions -- Felony Caseload (24), Misdemeanor Caseload (24), Mixed Caseload (24) and two other DISABLED dimensions. Now if we go to Explore Data page we will see this picture: 
Снимок экрана 2024-09-23 в 14 47 32

Initially I was thinking that we could simply remove disabled dimensions form breakdown view, but if we do that the breakdown will look something like:

• Total____120

• Felony Caseload____24 (20%)
• Misdemeanor Caseload____24 (20%)
• Mixed Caseload____24 (20%)

But this is not correct because the sum of this three dimensions would be equal 72. And as I understand we actually should display disabled dimensions, but maybe with an UNKNOWN label? What do you think about it?

mxosman:
> @mxosman Thank you for your research on this, I'm sure we will figure something out together!!
> 
> 1. As for publish review page - I actually already figured that out and implemented solutions -- please check them out! Or do you think this solutions are much inconvenient for us to use? LMK your thoughts!

Oh, thank you! Will review tomorrow morning - we should also consider the other review page flows `Bulk

> 2. I was thinking about how we should implement filtering on Explore data page and it seems to me that I have come up with a very interesting observation. Let's take for example Defense' (148) agency January 2023 record: we will see that Open Cases metric has total value (120) which consists of three ENABLED dimensions -- Felony Caseload (24), Misdemeanor Caseload (24), Mixed Caseload (24) and two other DISABLED dimensions. Now if we go to Explore Data page we will see this picture:
> 
> Снимок экрана 2024-09-23 в 14 47 32
> Initially I was thinking that we could simply remove disabled dimensions form breakdown view, but if we do that the breakdown will look something like:
> 
> • Total____120
> 
> • Felony Caseload____24 (20%) • Misdemeanor Caseload____24 (20%) • Mixed Caseload____24 (20%)
> 
> But this is not correct because the sum of this three dimensions would be equal 72. And as I understand we actually should display disabled dimensions, but maybe with an UNKNOWN label? What do you think about it?

Ugh - that's such a great observation - thank you _**so much**_ for flagging this! Hmm... since we don't do a calculation of the total in the Publisher UI (like the user now can enter any total they want and it doesn't necessarily have to match up with the dimensions) let's do this: if the dimension has no value, let's filter it out. If it does have a value - I like your idea of adding a label to the dimension name. Something like "Mixed Caseload (disabled)".

[Ideally, if we did compute the total based on dimension values, we'd probably get the total updating correctly for free from filtering it out.]

> 3. As for suggestion on using some helper function in `DatapointsStore` -- the thing is that as I understand we only have dimensions enabled statuses in `FormStore` and `MetricConfigStore`, but I can't access nor initialize these stores in `DatapointStore` or even access them in `ConnectedDatapointsView` via `useStore` ☹️

Oh, is the dimension enabled status not available in the ReportStore? There's already a link between the DatapointsStore and ReportStore -- I'll take another look in the morning, but I could've sworn the metric and dimension info were available in the ReportStore.

mxosman:
Hi @nasaownsky - just following up on #3. So, you can access the `ReportStore`'s `agencyMetrics` for the relevant metrics, disaggregations and dimensions and check their enabled status. Let me know if you're having trouble or if I'm misunderstanding -- here's a screenshot:

Screenshot 2024-09-24 at 9 52 42 AM

publisher/src/components/Reports/DataEntryReview.tsx
```diff
@@ -119,8 +119,20 @@ const DataEntryReview = () => {
...
+            formStore.disaggregations?.[reportID]?.[metric.key] || {}
+          ).flatMap((dimensions) => Object.keys(dimensions));
+
+          const enabledDatapoints = datapointsByMetric[metric.key].filter(
+            (datapoint) =>
+              datapoint.dimension_display_name === null ||
+              enabledDatapointsDimensions.includes(
+                datapoint.dimension_display_name
+              )
+          );
```
mxosman, line 132: Hmm - we should consider the `SpreadsheetReview` flow as well as the `BulkActionReview` flow. Right now, these changes work on the `DataEntryReview`, but if I go to the Records page and bulk publish that same record I was working on, it gives me a "There is no data to review. Please go back and enter data in order to Review & Publish." message.

The bulk publish and data entry metrics come from here in the `ReportStore`: https://github.com/Recidiviz/justice-counts/blob/main/publisher/src/stores/ReportStore.ts#L238

To repro this -- edit a record with disabled dimensions, then go back to the Records page > click Bulk Actions > click Publish > select the record you edited > click Review and Publish.

mxosman:
Hi Ilya! Just checking in to see if you needed any help with this/if you are running into issues I can help unblock.

nasaownsky:
Hi Mahmoud! I stepped back from this PR for past two weeks to focus on other tasks as I felt a bit struggling with it, but I definitely will jump back on it this week. Sorry for this, I hope this is not urgent task! Let me know if it is, so then I'll jump back on it right in!

mxosman:
Oh no apologies necessary at all - it's not urgent and I'll definitely let you know if/when it becomes urgent. I'm glad you were able to focus on other tasks - mostly checking in to see if I can unblock in any way. Sorry, this one is gnarly - but, please don't hesitate to let me know if I can help. Thank you again for all your help as always.

nasaownsky:
Ugh, that take so much time and effort to figure out! I'm so sorry for my tardiness with this one, @mxosman! Now it finally should work as expected.

publisher/src/stores/ReportStore.ts
```diff
@@ -300,14 +301,24 @@ class ReportStore {
           return acc;
         }, {} as PublishReviewMetricErrors);
 
+      const currentAgencyMetrics: Metric[] = Object.values(
+        await this.initializeReportSettings(currentAgencyId)
```
mxosman, line 305: This feels very odd - why do we need to do this?

publisher/src/stores/ReportStore.ts
```diff
@@ -300,14 +301,24 @@ class ReportStore {
...
+            datapointMatchingEnabledDimension(dp, currentAgencyMetrics) // Filter out disabled dimensions
         );
 
       const datapointsByMetric =
         DatapointsStore.keyRawDatapointsByMetric(filteredDatapoints);
-      const datapointsEntries = Object.entries(datapointsByMetric);
+
+      const datapointsEntries = Object.entries(datapointsByMetric).filter(
+        ([metricKey]) =>
+          currentAgencyMetrics.find((d) => d.enabled && d.key === metricKey)
```
mxosman, line 320: nit: instead of `d` can we use `metric` for clarity?

common/components/DataViz/utils.ts
```diff
@@ -730,6 +730,28 @@ export const getMetricKeyToFrequencyMap = (
...
+        disaggregation.dimensions
+          .filter((dimension) => dimension.enabled)
+          .map((dimension) => dimension.key)
+          .includes(dp.dimension_display_name as string)
+      )
+    );
+  }
+
+  return true;
+};
```
mxosman, line 753: Nice! Gnarly computation but I can't think of a way around having to dig in that deep.

mxosman (commented):
Ilya - thank you for working on this gnarly piece! I appreciate your diligence with this one. I left a question for my understand, but manually tested and it feels great. If you wouldn't mind doing the same for a few other agencies, that would be great too.

Just two small requests:
* Can we write a unit test for the new util function?
* This was probably existing before this change, but I noticed in the Explore Data page, when a user has entered 0 in a report it appears as Not Reported instead of 0. Would you mind making an adjustment to show the 0 entry? I feel like zeros are being filtered out in the datapoints somewhere upstream of the bar chart tooltips. If this wasn't a result of this change and/or you'd rather focus on it in a separate PR, let me know and I'll create a separate ticket.

publisher/src/stores/ReportStore.ts
nasaownsky: I slightly updated this -- we don't need the calculation, but we need to call this initialize function, because in other case `this.agencyMetrics` is an empty array.

nasaownsky:
Hey @mxosman! Applied the fixes and added unit test.

>If you wouldn't mind doing the same for a few other agencies, that would be great too.

Could you clarify this? I think this should work for all agencies.

common/stores/BaseDatapointsStore.ts
```diff
@@ -90,7 +90,7 @@ abstract class DatapointsStore {
         });
         // the res object could be overwritten based on the datapoint order since the logic is grouping by start_date
         // and this check essentially skips all null datapoints and writes correct data
-      } else if (dp.value) {
+      } else if (dp.value || dp.value === 0) {
```
mxosman, line 93: Thank you for adding!

publisher/src/stores/ReportStore.ts
```diff
@@ -300,14 +301,24 @@ class ReportStore {
           return acc;
         }, {} as PublishReviewMetricErrors);
 
+      await this.initializeReportSettings(currentAgencyId);
```
mxosman, line 304: Thanks for explaining this! A bit painful to be making two calls here, but I suppose it's unavoidable to get the info we need. 

Can we do a quick check here before calling this if we've already called it/if we have values for `this.agencyMetrics` or `this.metricsBySystem`? When a user goes to the Explore Data page, we call this endpoint - and I'm thinking it would be nice to not have to call it again if we already have the info we need. We should test and make sure that if we add a conditional and only call it once, that we make sure that it doesn't result in a stale state (like if a user goes to the Explore Data, `initializeReportSettings` is called and not again here, and a user makes an update to disable/enable a dimension, that when we run this function again and bypass calling `initializeReportSettings`, that we'll be operating with the updated `enabled` values). Does that make sense?

mxosman (commented):
> Hey @mxosman! Applied the fixes and added unit test.
> 

Thank you so much for the adjustments! Looks much better! Left one comment about calling the `initializeReportSettings`, but happy to accept after some clarification.

> > If you wouldn't mind doing the same for a few other agencies, that would be great too.
> 
> Could you clarify this? I think this should work for all agencies.
Yeah, mostly just saying that we shouldn't assume it works for all agencies without testing 2-3+ different types of agencies. Because we don't have robust unit and integration tests, and there are a lot of random edge cases between agencies - it's better to test the change across a few different types of agencies.

publisher/src/stores/ReportStore.ts
nasaownsky: Totally! Tested it across the board and in multiple agencies -- no issues at all, everything worked as expected.

mxosman (approved):
Brilliant work on this, @nasaownsky! I appreciate your patience and diligence on this so so much!

publisher/src/stores/ReportStore.ts
```diff
@@ -301,7 +301,9 @@ class ReportStore {
           return acc;
         }, {} as PublishReviewMetricErrors);
 
-      await this.initializeReportSettings(currentAgencyId);
+      if (!this.agencyMetrics.length) {
+        await this.initializeReportSettings(currentAgencyId);
+      }
```
mxosman, line 306: Thank you so much!

PR #1498, opened 2024-09-10
[Publisher][Bug] Explore Data: Breakdowns show missing data

nasaownsky:
## Description of the change

> Description here

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review


mxosman:
Thank you for all the sleuthing @nasaownsky! I have no idea why that is happening. I think ultimately the UI shouldn't care about the ordering of things and display the accurate views. Can you dig into what we need to do in order for the UI to stay agnostic to the ordering?

nasaownsky:
I'll try! The problem not in the UI itself, but in datapoints preparing method, which gives them wrong frequency depending on the order of the data.

mxosman:
Sorry by UI - I was meaning the frontend in general including the datapoints prep the stores are doing. If there's a way the FE can handle these cases, that would be preferred b/c the ordering shouldn't matter on the FE - it's so strange that it's creating this kind of issue.

mxosman:
Wait... my mind is absolutely blown! Taking a closer look shortly.

mxosman:
Incredible sleuthing, @nasaownsky!

>by checking if datapoint actually has value we are getting rid of overlapping. Besides, only datapoints with values ​​will reach this part of that method in any case.

Can you elaborate on this part? I'm having trouble understanding why checking for the presence of a value and frequency prevents the overwriting.

nasaownsky:
@mxosman Yes, you're right! Frequency check is unnecessary there.
Just checked it twice, comparing this change's charts with staging charts, and it seems all of them are the same. But you can also check this, so that I don't sound unfounded!

mxosman (approved):
> @mxosman Yes, you're right! Frequency check is unnecessary there. Just checked it twice, comparing this change's charts with staging charts, and it seems all of them are the same. But you can also check this, so that I don't sound unfounded!

Amazing - looks good to me too! One small request - can you add a comment on that `else` conditional with a brief explanation on why we're adding this for our future selves?

Absolutely incredible sleuthing, @nasaownsky! I appreciate it so much!

nasaownsky:
Thank you @mxosman! How should we merge this one?

mxosman:
> Thank you @mxosman! How should we merge this one?

Ah, you can rebase this on top of main and undo the injected datapoints array in the `DatapointsStore` if the rebase doesn't clear it back to this: https://github.com/Recidiviz/justice-counts/blob/8307a2d20f16a06c81ae20f75cf760cb1da530fe/publisher/src/stores/DatapointsStore.ts#L84

PR #1508, opened 2024-09-17
[Publisher] Don't filter out 0 values in `Review Data` page

### LINKED ISSUE #1499: [Publisher] When uploading 0s, Bulk Upload review page shows "-" rather than true 0.
Bulk upload review page: when uploading data that includes 0s, review page shows "-" rather than true 0. The 0 is recorded and saved in the record, so this is just a display issue on the review page.

Screenshot 2024-09-12 at 10 16 50 AM

Screenshot 2024-09-12 at 10 16 56 AM

nichelle-hall:
## Description of the change

**Context**: Matt reported that when uploading 0 values in a sheet via Bulk Upload, these values were not displayed on the Review Data page. Instead, the system acted as if no data had been supplied. 

BEFORE
image

**Investigation**: After investigating, I confirmed that while the backend correctly saved the 0 values, the frontend was filtering them out. The frontend logic treated 0 values as if they were missing or undefined, leading to them being excluded from display.

**Fix**: I updated the frontend filtering logic to ensure that only undefined values are filtered out. This allows 0 values to be displayed properly on the Review Data page, as they are valid data points. I tested locally by creating the super/ child agencies in my local environment and tested with the sheet Matt supplied. 

AFTER
Screenshot 2024-09-16 at 8 01 52 PM

@mxosman  I didn't realize that this was already assigned to Ilya! It was a quick fix and I wanted to confirm that it wasn't a result of my refactor changes so I went and knocked it out. 

## Related issues

Closes #1499

### Development

### Code review

### DIFF
--- a/publisher/src/stores/ReportStore.ts
+++ b/publisher/src/stores/ReportStore.ts
@@ -302,7 +302,7 @@ class ReportStore {
 
       const filteredDatapoints =
         combinedFilteredDatapointsFromAllReports.filter(
-          (dp) => dp.value !== null
+          (dp) => dp.value || dp.value === 0 // Filter out null values
         );
 
       const datapointsByMetric =
--- a/publisher/src/components/DataUpload/SpreadsheetReview.tsx
+++ b/publisher/src/components/DataUpload/SpreadsheetReview.tsx
@@ -127,7 +127,7 @@ export function SpreadsheetReview({
   ) => {
     const allDatapoints = metrics
       .flatMap((metric) => metric.datapoints)
-      .filter((dp) => dp.value);
+      .filter((dp) => dp.value || dp.value === 0); // Filter out null values
     const datapointsByMetricNameByAgencyName = allDatapoints.reduce(
       (acc, dp) => {
         if (!dp.agency_name || !dp.metric_display_name) return acc;


mxosman (commented):
Thanks for working on this, Nichelle! There are a few other areas to consider updating and testing as well that point to the review page -- the `DataEntryReview` and `BulkActionReview`. Or alternatively, we could do this filtering in the `ReviewMetrics` component b/c that's where all of the review pages feed the metrics to.

If you could include a comment explaining the filter, that would be super helpful since there's a ton of filtering happening on the FE and it would be good to tag what each one's purpose is.

publisher/src/stores/ReportStore.ts
```diff
@@ -302,7 +302,7 @@ class ReportStore {
 
       const filteredDatapoints =
         combinedFilteredDatapointsFromAllReports.filter(
-          (dp) => dp.value !== null
+          (dp) => dp.value || dp.value === 0
         );
```
nichelle-hall, line 307: This is the datapoints source for both `Bulk Action Review` and `Data Entry Review`.

mxosman: Ah great call updating the source!!

mxosman (approved):
LGTM to me Nichelle! Thank you for handling this! Just another request to add comments in those two areas for posterity.

nichelle-hall:
> LGTM to me Nichelle! Thank you for handling this! Just another request to add comments in those two areas for posterity.

For sure! I'll add the comments before merging.

morden35:
Thank you for this!!

PR #1511, opened 2024-09-19
[Publisher] Only query an agency if the user visits that agency's page.

brandon-hills:
## Description of the change

Only query an agency if the user visits that agency's page.

Closes [#28486](https://app.zenhub.com/workspaces/justice-counts-622109ea3bb6e0001d2d06ad/issues/gh/recidiviz/recidiviz-data/28486)

Important:  This change depends on [brandon/get-agency-endpoint](https://github.com/Recidiviz/recidiviz-data/pull/33457), so do not merge this until [brandon/get-agency-endpoint](https://github.com/Recidiviz/recidiviz-data/pull/33457) is merged.

## Testing
Tested this extensively locally and on playtesting.

Try testing this change out yourself at https://brandontest---publisher-web-b47yvyxs3q-uc.a.run.app/, which uses [brandon/one-at-a-time](https://github.com/Recidiviz/justice-counts/pull/1511) for the frontend, and [brandon/get-agency-endpoint](https://github.com/Recidiviz/recidiviz-data/pull/33457) for the backend.

Currently, we are querying ALL of a user's agencies during load-time (or page refresh). Instead, let's only query the agency that we are visiting so that we can drastically reduce the time it takes to load Publisher.

This change does the following:

* removes the call to 'api/users/' (which creates/updates the user and loads ALL of a user's agencies into the userStore). *Replaces 'api/users/' with a single call to 'api/user_dropdown' (which loads _just_ the list of agency names and ids that a user has access to) so that we can render the dropdown menu
* Calls 'api/agency/<agency_id>/agency_group' whenever we visit an agency page (if the agency is not already in the userStore). This fetches _both_ the requested agency as well as that agency's superagency (only if it's a child agency and the user has access to that superagency). The superagency is necessary so that we can render the child agency dropdown in the metrics page.

## Reason for change
This is necessary because current the run-time for populating the UserStore grows proportional to the number of agencies that a User has access to. CSG and Recidiviz users currently have access to over two thousand agencies, and the number of agencies we have onboarded to Publisher is growing every day.

Today, it takes upwards of 20s to load publisher on Prod.

Screenshot 2024-09-22 at 3 07 31 PM

This change will reduce that time to ~2 seconds, and make page-rendering time roughly constant, no longer increasing proportionally to the number of agencies a user has access to. This will be very important for the future as we continue onboarding new agencies.

Screenshot 2024-09-22 at 3 16 28 PM


publisher/src/App.tsx
```diff
@@ -46,6 +46,29 @@ const App: React.FC = (): ReactElement => {
     trackNavigation(location.pathname + location.search);
   }, [location]);
 
+  useEffect(() => {
```
brandon-hills, line 49: Thanks for helping me find this spot @mxosman !! 

Placing the fetchAgencyData call here means that we are fetching that agency's information every time the user navigates to a page that has /agency/<agency_id>/ in the path, which is all pages that rely on having that agency's data in the UserStore!

publisher/src/components/Menu/Menu.tsx
```diff
@@ -85,33 +84,18 @@ const Menu: React.FC = () => {
...
-                : ""
-            }`,
-            onClick: () => {
-              navigate(`/agency/${agency.id}/${pathWithoutAgency}`);
-              handleCloseMobileMenu();
-            },
-            highlight: agency.id === currentAgency?.id,
-          };
-        })
+  const agencyDropdownOptions: DropdownOption[] = userStore.dropdownAgencies
```
brandon-hills, line 87: Agency dropdown names are now being formatted (appending state codes if necessary) in the backend endpoint instead of here!

publisher/src/stores/UserStore.ts
```diff
@@ -31,8 +35,12 @@ class UserStore {
 
   api: API;
 
+  dropdownAgencies: DropdownAgency[] | undefined;
+
   userAgencies: UserAgency[] | undefined;
 
+  queriedAgencies: string[] = [];
```
brandon-hills, line 42: Queried agencies is used to make sure that we only make a backend query for the agency once

publisher/src/stores/UserStore.ts
```diff
@@ -163,6 +170,83 @@ class UserStore {
     return undefined;
   }
 
+  /**
```
brandon-hills, line 173: Oops, this docstring is out-of-date. Let me update it!

publisher/src/stores/UserStore.ts
```diff
@@ -163,6 +170,83 @@ class UserStore {
...
+   * agencies relevant to the agency.
+   * For example, if the agency is a child agency, this fetches its superagency and all
+   * of the child agencies of that superagency. If it is a superagency, we fetch that
+   * agency and all of its child agencies. If the agency is neither a child or
+   * superagency, we just fetch that one agency.
+   *
+   * @param {string} agencyId - The ID of the agency to fetch data for.
+   * @returns {Promise<UserAgency | undefined>} - The agency object or undefined if not found.
+   */
+  async fetchAgencyGroupData(
```
brandon-hills, line 185: We are only fetching the agency and it's superagency (if one exists). We dont' need to fetch a superagency's child agencies, since it turns out that we are calling another endpoint /agencies/<super_agency_id>/children to render the super/child agency dropdowns.

This is an exciting find because it means that this homepage speedup will be similarly fast even for agency's with a ton of child agencies!

publisher/src/stores/UserStore.ts
mxosman: I'm actually wondering if we need this separate property and functionality to keep track of this and instead infer from the `userAgencies` whether or not we've made the query. If the property exists in the `userAgencies`, I think we can safely assume it's already been queries, but there might be something I'm not thinking about.

publisher/src/App.tsx
mxosman: Nice!! A couple of things with this `useEffect`:

1. We'll need to add a dependency array so the `useEffect` knows exactly when to re-run. In this case, we want it to re-run anytime the agency id in the URL params change. We can define a dependency array as the second argument for the `useEffect` and include the agency id in that check. That way, when just the agency id changes we can re-run the effect.

2. A quick refactor proposal to clean up a lot of the logic here:

```
  // We can define these outside since the loader conditional also uses this
  const agencyMatch = location.pathname.match(/\/agency\/(\d+)/);
  const agencyId = agencyMatch ? agencyMatch[1] : "";

  useEffect(() => {
    // Since all of these conditions `return` we can consolidate them in one conditional
    if (
      !userStore.userInfoLoaded ||
      !agencyId ||
      userStore.queriedAgencies.includes(agencyId)
    ) {
      return;
    }

    const fetchAgencyData = async () => {
      await userStore.fetchAgencyGroupData(agencyId);
      userStore.queriedAgencies.push(agencyId); // Moved this to after if we're still using this to keep track of what we've queried
    };

    fetchAgencyData();
  }, [agencyId, userStore]);
```

A quick way to test when this is going to fire off is by adding a console.log within the useEffect before the first line - each time you see it in the console is when the effect is triggered.

I haven't checked to see whether this works - but let me know if you run into any issues!

mxosman: There are some tiny side effects that we can think about smoothing out that might be tricky, and that's the full page loader in between each agency switch which gives a visible blink on a good connection. But, we can cross this bridge later!

mxosman (commented):
Full-stack Brandon Hills! Amazing work diving into the frontend and all of the mobx complexity - super impressed with what you've figured out so far!! Thank you for the detailed description and comments throughout - and blown away by the performance improvements to this flow! 

Let me know if you have any questions or where I can clarify further w/ my feedback!

brandon-hills:
> We'll need to add a dependency array so the useEffect knows exactly when to re-run. In this case, we want it to re-run anytime the agency id in the URL params change. We can define a dependency array as the second argument for the useEffect and include the agency id in that check. That way, when just the agency id changes we can re-run the effect.

@mxosman Thanks for the pointer on the dependency array!! Just added it in

brandon-hills:
>I'm actually wondering if we need this separate property and functionality to keep track of this and instead infer from the userAgencies whether or not we've made the query. If the property exists in the userAgencies, I think we can safely assume it's already been queries, but there might be something I'm not thinking about.

You're so right! I think we can actually get right of queriedAgencies entirely! I was actually using it before to make sure that we weren't querying for the agency data more than once, but adding the dependency array like you suggested solves that problem!

publisher/src/App.tsx
```diff
@@ -46,6 +46,29 @@ const App: React.FC = (): ReactElement => {
...
+    // Only load agency data after the user store has been constructed.
```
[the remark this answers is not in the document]
brandon-hills: Ah, I didn't realize that! This means that it's probably a useless thing to check - I'll remove it!

publisher/src/App.tsx
brandon-hills: Thanks for the refactor code!! Thats a lot better 💛

brandon-hills: @mxosman Thanks for the pointer on the dependency array!! Just added it in

publisher/src/stores/UserStore.ts
brandon-hills: You're so right @mxosman ! I think we can actually get right of queriedAgencies entirely! I was actually using it before to make sure that we weren't querying for the agency data more than once, but adding the dependency array like you suggested solves that problem!

common/types.ts
```diff
@@ -96,6 +96,11 @@ export interface PublicUserAgency {
   systems: AgencySystem[];
 }
 
+export interface DropdownAgency {
+  agency_id: number;
+  dropdown_name: string;
+}
```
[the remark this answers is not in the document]
brandon-hills: @mxosman The other export interface structs in this file are using snake case for the variable names!

I'll definitely just follow your lead on this one, but just wanted to double check!

publisher/src/stores/UserStore.ts
brandon-hills: Nope! `/agencies/<super_agency_id>/children` is called elsewhere in the codebase, so we don't need to worry about it here! 

I'll update the function doc-string, it's out of date from when I _thought_ that this function needed to fetch child agencies too.

publisher/src/stores/UserStore.ts
```diff
@@ -163,6 +170,83 @@ class UserStore {
...
+  ): Promise<UserAgency | undefined> {
+    if (!agencyId) {
+      return undefined;
+    }
+    try {
+      const response = (await this.api.request({
+        path: `/api/agency/${agencyId}/agency_group`,
+        method: "GET",
+      })) as Response;
+      if (response && response instanceof Response) {
```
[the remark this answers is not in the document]
brandon-hills: @mxosman you're kind to say that it's bc i was trying to add extra security. It's really just that I only vaguely understand how Response parsing works 😆 

Thanks for the fix!!

brandon-hills:
> Full-stack Brandon Hills! Amazing work diving into the frontend and all of the mobx complexity - super impressed with what you've figured out so far!! Thank you for the detailed description and comments throughout - and blown away by the performance improvements to this flow!

> Let me know if you have any questions or where I can clarify further w/ my feedback!

Thank you so much @mxosman !! 😁 😁 This was a fun challenge! 

Also thanks for the super thorough review, I learned a lot even just responding to your review comments!

publisher/src/stores/UserStore.ts
```diff
@@ -163,6 +170,83 @@ class UserStore {
...
+              const index = this.userAgencies.findIndex(
+                (a) => a.id === userAgency.id
+              );
+              if (index !== -1) {
+                this.userAgencies[index] = userAgency;
+              } else {
+                // Add to the list if the new agency doesn't exist already.
+                this.userAgencies = this.userAgencies.concat(userAgency);
+              }
+            });
```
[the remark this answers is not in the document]
brandon-hills: You explained that perfectly @mxosman ! (tho lmk if i actually understood what you had in mind haha)

I used userAgenciesById as the variable name!

Agreed, this is a great optimization

publisher/src/stores/UserStore.ts
```diff
@@ -163,6 +170,83 @@ class UserStore {
...
+        }
+      }
+      showToast({
+        message: "Failed to fetch agency data.",
+        color: "red",
+      });
+      runInAction(() => {
+        this.loadingError = true;
+      });
+      return undefined;
```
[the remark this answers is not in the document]
brandon-hills: Yes thank you! Thats way cleaner now

publisher/src/stores/UserStore.ts
brandon-hills: Updated!

brandon-hills:
Also @mxosman I just re-deployed this to https://brandontest---publisher-web-b47yvyxs3q-uc.a.run.app/ so that it's easy for you to inspect the code changes!

I used `./recidiviz/tools/deploy/justice_counts/deploy_for_playtesting.sh -b brandon/get-agency-endpoint -f brandon/one-at-a-time -a publisher -t brandontest`

publisher/src/stores/UserStore.ts
mxosman: Beautiful! TYSM for updating!

publisher/src/App.tsx
mxosman: Yeah, if it's OK, let me think about the blinking some more today before giving a green light incase it's something small we can adjust! If we do a follow up - we'll want to do a fast follow as it's a bit undesirable and I'm sure will be flagged by CSG. 

Otherwise - this is looking **_fantastic_**!

publisher/src/stores/UserStore.ts
mxosman: Oh amazing!! It looks so much cleaner!

publisher/src/stores/UserStore.ts
mxosman: Gotcha!! Thank you for updating the doc-string!

common/types.ts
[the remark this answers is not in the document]
mxosman: Oh wow - we've been using it for all the types in that common file. It's not blocking for me at all - you can leave it as is! Normally we'd convert the backend properties to snakeCase - but, since we've already set a precedent keeping it is totally fine!

publisher/src/stores/UserStore.ts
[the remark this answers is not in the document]
mxosman: Thank you, Brandon! Agreed - looks much cleaner now!

publisher/src/stores/UserStore.ts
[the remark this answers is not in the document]
mxosman: Hahaha! 🤣  Omg - that cracked me up. 

FWIW - normally when we typecast and say `as ThisType`, we're explicitly overriding TypeScript's helpful type inference and saying "No, TypeScript - you are wrong, it's 100% this type that I'm setting and will always be" - which is not entirely safe since it could be the other types that TypeScript is inferring. So, we do everything we can to minimize type casting unless we're super sure we know what we're doing. For this one -- the `request` function we created can technically return a `Body` or `string` type - which is rare and only when we're hitting certain endpoints (though I'm not sure if it's still the case that we need these other types). For your future TS development, I'd avoid/minimize as much type casting as you can, otherwise it negates a lot of the benefits of TS.

publisher/src/stores/UserStore.ts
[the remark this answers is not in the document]
mxosman: You understood and implemented it PERFECTLY! I'm so glad it made sense. Love the variable name and your implementation of it. Beautiful!

publisher/src/App.tsx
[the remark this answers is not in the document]
mxosman: It's super unintuitive when all of this happens and at which order!

mxosman (commented):
Wow Brandon! Thank you for all of these changes and discussion! I want to give a little more thought this afternoon about the blinking and just want to be sure `App.tsx` is truly the best area to make this happen. Otherwise, FANTASTIC WORK figuring all of this out and implementing it! 👏 🐐 👏 🐐

mxosman:
OK @brandon-hills - I have a proposal to help the full page blinking if you're up for it. This will move things away from the `App` component to the `Router` + an update in the `Menu` components, which I actually think would work better.

**Move logic from App component to Router**
* We can move the `useEffect` logic from the `App` component to the `Router` component's existing `useEffect`. This is nice because it already has everything we need and the `agencyId` extracted from the params using a `useParams()` hook. 
  * My only other suggestion is to add a brief comment within the `useEffect` to describe the purpose of the logic.
https://github.com/Recidiviz/justice-counts/blob/400158a9368be034c538854edb66c680fd44ddee/publisher/src/router/Router.tsx#L45-L49
* After the `useEffect`, we can add the loading conditional + the `Header` component like so:
```
  if (!userStore.getAgency(agencyId))
    return (
      <PageWrapper>
        <Header />
        <Loading />
      </PageWrapper>
    );
```
  * This alone should help the full page blinking because it persists the header - but you'll notice it now creates a different blinking within the dropdown menu part of the header that lists the agency name b/c we're waiting for the query for the agency to update its name. The good thing is, we already have all of the names with `dropdownAgencies`.
  * The last thing we need to do here is wrap the `Router` component in the `observer` HOC (higher-order function) so that it can listen and respond to changes in the mobX stores used within the component. 

```
// From this:
export const Router = () => {
   // ...component body
   return (<> ... stuff ... </>);
}

// To this:
export const Router = observer(() => {
   // ...component body
   return (<> ... stuff ... </>);
})
```

**Update Menu component**
* Here, all we need to do is update the `currentAgency?.name` to read from the `dropdownAgencies` name as a fallback, while we query for the full agency object, if that makes sense. At this point, the UI will hide the name when switching between agencies because it's waiting for the full agency object to read the name.
* We can make things a bit easier to get the thin agency object name if we do the same thing we did with `userAgencies` and create a map similar to `userAgenciesById`. 
* Then, we can update the following to `const agencyName = currentAgency?.name ?? userStore.dropdownAgenciesById[agencyId].name`:

https://github.com/Recidiviz/justice-counts/blob/d44f0fc386cdceb023fb3d5a7e84efe8f21015a2/publisher/src/components/Menu/Menu.tsx#L46

  * BTW, I noticed that we don't use this `agencyName` throughout when we should in place of `currentAgency?.name` - we can update the following areas in L194 and L200 to use the newly updated `agencyName`:

https://github.com/Recidiviz/justice-counts/blob/d44f0fc386cdceb023fb3d5a7e84efe8f21015a2/publisher/src/components/Menu/Menu.tsx#L192-L211
 
  * There's still a final blink b/c the `dropdownAgencies` adds the state code to the name where as the full agency object does not. We can strip the state code in the FE or we can add the state code to both - up to you!

OK - let me know if this makes sense or if it feels overwhelming! In short, you're moving the logic you ironed out in `App` to a different component, making that component an observer, updating 3 lines in the Menu, and a mapping of the dropdown agencies for ease in the UserStore (+ the state code thing).

common/types.ts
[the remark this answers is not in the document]
brandon-hills: i'll leave as-is then, thanks @mxosman !

publisher/src/stores/UserStore.ts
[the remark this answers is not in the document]
brandon-hills: Woohoo!! Well thank you again for explaining it so thoroughly!

brandon-hills:
> FWIW - normally when we typecast and say as ThisType, we're explicitly overriding TypeScript's helpful type inference and saying "No, TypeScript - you are wrong, it's 100% this type that I'm setting and will always be" - which is not entirely safe since it could be the other types that TypeScript is inferring. So, we do everything we can to minimize type casting unless we're super sure we know what we're doing. For this one -- the request function we created can technically return a Body or string type - which is rare and only when we're hitting certain endpoints (though I'm not sure if it's still the case that we need these other types). For your future TS development, I'd avoid/minimize as much type casting as you can, otherwise it negates a lot of the benefits of TS.

@mxosman Thanks for the super helpful teaching moment here! My future TS development sends you many thanks 😁 

Honestly I've learned so much from this PR hahaha

brandon-hills:
@mxosman You're anti-blinker instructions are LEGENDARY !

The solution is brilliant - I love how the header stays constant even when the rest of the page is loading!

And thank you for the super thorough instructions once again - I learned a lot 😁 

I ended up just using the dropdown agency name for the header always! This means that the state code will also be there. I'm thinking that should be fine, but LMK if you think otherwise!

brandon-hills:
Also @mxosman  I redeployed this to https://brandontest---publisher-web-b47yvyxs3q-uc.a.run.app/ to make it easy to test!

mxosman (approved):
Brandon... I wish I could truly and accurately express how incredible all of this work is! The optimization is pure butter - it really feels super delightful. You've taught me a lot just working on figuring out a way to optimize the homepage loading. Thank you for implementing this + entertaining all of the changes. I'm so happy to hear that there were some nuggets of info you found useful throughout this - nothing that you wouldn't have figured out on your own anyway!

Anyway, many thanks again for tackling the FE portion with such ease! 🙏 👏  Huge kudos on such an impactful full-stack, end-to-end implementation!!

brandon-hills:
OMG hahaha 
@mxosman That gif perfectly captures my emotions submitting this 😆 I'm thrilled to get this in today, especially excited to deploy it later and see if it truly is smooth-like-butter 

Thank you again for your awesome guidance and code reviews, especially on such short notice! I _definitely_ couldn't have done this without you!

PR #1528, opened 2024-10-03
[Publisher][Bug] Fix bug where adding team members to the admin panel clears out all other team members

mxosman:
## Description of the change

Fix bug where adding team members to the admin panel clears out all other team members due to recent optimizations of fetching team members separately.

This issue was raised by CSG members who noticed that adding a team member to an agency resulted in clearing out all other team members from the agency aside from the one recently added. Because we now fetch team members separately and ONLY when an agency card is clicked vs. before where they were populated for each agency immediately after the admin panel loads, we need to await the fetch of team members before updating them in the admin panel so all references to the team members will be the freshly fetched list.

Follow up TODO: Add unit tests to cover this area

## Type of change

## Related issues

Closes #XXXX

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
@@ -127,8 +127,15 @@ export const AgencyProvisioningOverview = observer(() => {
   const searchAndFilter = (e: React.ChangeEvent<HTMLInputElement>) => {
     setSearchInput(e.target.value);
   };
-  const editAgency = (agencyID: string | number) => {
-    const selectedAgency = agenciesByID[agencyID][0];
+  const editAgency = async (agencyID: string | number) => {
+    // Fetch the team associations for the agency.
+    await adminPanelStore.fetchAgencyTeam(String(agencyID));
+    /**
+     * Since we're performing a fetch and updating the store and using the updated value
+     * within this function, we need to use the `adminPanelStore._` instead of the destructured
+     * reference so mobX stays in sync with the update we just made to `agenciesByID`.
+     */
+    const selectedAgency = adminPanelStore.agenciesByID[agencyID][0];
     setSelectedAgencyID(agencyID);
     updateAgencyID(+agencyID);
     updateAgencyName(selectedAgency.name);
@@ -301,11 +308,7 @@ export const AgencyProvisioningOverview = observer(() => {
           : filteredAgencies.map((agency) => (
               <Styled.Card
                 key={agency.id}
-                onClick={() => {
-                  // Fetch the team associations for the agency.
-                  adminPanelStore.fetchAgencyTeam(String(agency.id));
-                  editAgency(agency.id);
-                }}
+                onClick={() => editAgency(agency.id)}
               >
                 {/* Name, State, ID */}
                 <Styled.TopCardRowWrapper>
--- a/publisher/src/components/AdminPanel/AdminPanel.test.tsx
+++ b/publisher/src/components/AdminPanel/AdminPanel.test.tsx
@@ -627,7 +627,7 @@ test("Clicking the `Create Agency` button opens the create agency modal", async 
   const createNewAgencyModalTitle = screen.getByText("Create New Agency");
   const editUserModalTitle = screen.queryByText("Edit Agency Information");
   const agencyInformationTab = screen.getByText("Agency Information");
-  const teamMemberRolesTab = await screen.getByText("Team Members & Roles");
+  const teamMemberRolesTab = screen.getByText("Team Members & Roles");
   const nameInput = screen.getByText("Name");
   const stateInput = screen.getByText("State");
   const countyInput = screen.getByText("County");
@@ -676,9 +676,12 @@ test("Clicking on an existing agency card opens the edit agency modal", async ()
   fireEvent.click(agencyProvisioningTab);
 
   const agency1Card = screen.getByText("Super Agency");
-  await act(() => fireEvent.click(agency1Card));
+  fireEvent.click(agency1Card);
 
-  const editAgencyModalTitle = screen.getByText("Edit Agency Information");
+  const editAgencyModalTitle = await screen.findByText(
+    "Edit Agency Information"
+  );
+
   const createNewAgencyModalTitle = screen.queryByText("Create New Agency");
   const agencyInformationTab = screen.getByText("Agency Information");
   const teamMemberRolesTab = screen.getByText("Team Members & Roles");
@@ -747,9 +750,9 @@ test("Team members tab renders with add/remove buttons and users who are connect
   fireEvent.click(agencyProvisioningTab);
 
   const agency1Card = screen.getByText("Super Agency");
-  await act(() => fireEvent.click(agency1Card));
+  fireEvent.click(agency1Card);
 
-  const teamMemberRolesTab = screen.getByText("Team Members & Roles");
+  const teamMemberRolesTab = await screen.findByText("Team Members & Roles");
   fireEvent.click(teamMemberRolesTab);
 
   const addUsersButton = screen.getByText("Add Users");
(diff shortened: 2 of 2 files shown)


publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
```diff
@@ -127,8 +127,15 @@ export const AgencyProvisioningOverview = observer(() => {
   const searchAndFilter = (e: React.ChangeEvent<HTMLInputElement>) => {
     setSearchInput(e.target.value);
   };
-  const editAgency = (agencyID: string | number) => {
-    const selectedAgency = agenciesByID[agencyID][0];
+  const editAgency = async (agencyID: string | number) => {
+    // Fetch the team associations for the agency.
+    await adminPanelStore.fetchAgencyTeam(String(agencyID));
```
mxosman, line 132: So, before when I said it was a one line change in the AgencyProvisioning, though it worked, I ended up going backt back further upstream to find the root of why things were out of sync since the AdminPanelStore eventually updates the agency's team member field that everything references. It turns out that we need to `await` the fetch populating the list of team members in the agency object, and use the `adminPanelStore.__` to stay in sync with the updated list we just fetched -- that way, by the time we include the team members in the store's `agencyProvisioningUpdates` in L149, it will be the fetched list and not an empty list.

LMK if this is confusing and I'll explain in a different way.

mxosman: This feels like a more proper solution though I hate that it waits for the fetch to open the modal. I'll think of a different solution or update a ticket I have for adding a loader to the Team Member & Roles tab to instead add a loader to the page while the modal opens.

lilidworkin (commented):
This solution makes a lot of sense to me!

Did you test locally? Should we deploy to playtesting to test there to be safe, and also to see how bad the lag is?

Do we also need to make an equivalent change [here](https://github.com/Recidiviz/justice-counts/blob/23e9ccc9b1af3cd74847781d852dc554fa938313/publisher/src/components/AdminPanel/UserProvisioningOverview.tsx#L226-L227) for editing a user?

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
lilidworkin: Yea this seems like a great solution! I think it's fine for now, as long as the lag isn't CRAZY bad. Maybe Ilya can add a loader a a fast-follow?

mxosman:
> This solution makes a lot of sense to me!
> 
> Did you test locally? Should we deploy to playtesting to test there to be safe, and also to see how bad the lag is?

Yes, I tested locally + staging -- will deploy to playtesting for an extra sanity check!

> Do we also need to make an equivalent change [here](https://github.com/Recidiviz/justice-counts/blob/23e9ccc9b1af3cd74847781d852dc554fa938313/publisher/src/components/AdminPanel/UserProvisioningOverview.tsx#L226-L227) for editing a user?

Ah, good call! Let me take a look and see. I was scrambling yesterday afternoon and tunnel-visioned on this particular fix that I didn't think about the user provisioning.

helperbot-recidiviz:

mxosman:
> Do we also need to make an equivalent change [here](https://github.com/Recidiviz/justice-counts/blob/23e9ccc9b1af3cd74847781d852dc554fa938313/publisher/src/components/AdminPanel/UserProvisioningOverview.tsx#L226-L227) for editing a user?

Interestingly enough, this flow seems to work just fine. The core issue with the agency provisioning is here in L281:

https://github.com/Recidiviz/justice-counts/blob/23e9ccc9b1af3cd74847781d852dc554fa938313/publisher/src/components/AdminPanel/AgencyProvisioning.tsx#L280-L291

...where it's using the `agencyProvisioningUpdates.team` as a reference to the existing team members. The one line change that worked prior to this change was using `currentTeamMembers` instead of `agencyProvisioningUpdates.team`:

https://github.com/Recidiviz/justice-counts/blob/23e9ccc9b1af3cd74847781d852dc554fa938313/publisher/src/components/AdminPanel/AgencyProvisioning.tsx#L204-L211

I'll need to dig in more to better understand why the `agencyProvisioningUpdates` and the `selectedAgency` team are crossing the finish line at different times (and it seems that the `agencyProvisioningUpdates.team` doesn't ever quite get the update for some reason)...

The `UserProvisioning` only ever references the `selectedAgency` and not the `userProvisioningUpdates` for the current list of teams.

To keep everything in-line between the two flows, we could just go back to the original solution of using `currentTeamMembers` and everything will play nicely because it's referencing the `selectedAgency` and once the team is fetched and updated in the store, it will re-render with the proper team on `selectedAgency.team`. Whereas using the `agencyProvisioningUpdates.team` seems to not ever re-render with the proper team likely due to the store not updating the `agencyProvisioningUpdates.team` by the time it's fetched? I'm not 100% sure - would have to test that assumption. 

Either way, using the `currentTeamMembers` didn't quite feel like it was addressing the root cause at the time and I figured going a bit upstream would make more sense, but now that I'm writing this out and thinking more about it - it might make sense to stick with `currentTeamMembers` b/c the `selectedAgency` (or in the case of UserProvisioning - `selectedUser`) will always have the latest values.

Does any of my ramble make sense? Please let me know if it doesn't and I'll explain in a different way!

TLDR - I'm debating going back to the original solution and scrapping the changes in the overview so both UserProvisioning and AgencyProvisioning are congruent -- or updating the UserProvisioning to do the fetching in the same way here to keep that congruent. Both options will have the same results - it's just the latter blocks the modal opening until everything is fetched, whereas the former opens the modal and re-renders it when everything is fetched and updated in the store.

lilidworkin:
@mxosman thank you for digging into this and for the super thoughtful explanation! I think I'll ultimately defer to you here. I'm tempted to say I like the cleanliness/safety of just blocking the modal opening until everything is fetched -- that feels like the most understandable solution and maybe the least prone to bugs getting introduced in the future? But I definitely don't feel strongly. +1 to being consistent across both flows, though.

mxosman:
> @mxosman thank you for digging into this and for the super thoughtful explanation! I think I'll ultimately defer to you here. I'm tempted to say I like the cleanliness/safety of just blocking the modal opening until everything is fetched -- that feels like the most understandable solution and maybe the least prone to bugs getting introduced in the future? But I definitely don't feel strongly. +1 to being consistent across both flows, though.

OK - the more I think about it - it actually makes more sense to go with the one-line solution. That actually feels the most congruent across the board - both areas should reference the `selectedAgency`/`selectedUser` as the source of truth for the things that are being fetched separately so the UI re-renders when the store is updated. It doesn't make sense to make this same update in the UserProvisioning when that flow references the `selectedUser` and doesn't reference `userProvisioningUpdates.agencies` since `selectedUser` will always have and display the latest update. So the change in this PR accommodates the usage of `agencyProvisioningUpdates.team` in addition to the `selectedAgencies` (which is already naturally in sync). I don't think we need to also accommodate using a separate reference when we can (and imo should) use the central source of truth `selectedAgencies` for the additional fetch of team members.

Both modals technically wait for the fetch to complete - this additional `await` waits for the store to update before the modal is ready, which is not necessary b/c we can add a loader within the agencies box (for user provisioning) and in the team members tab (for agency provisioning) - there's already a task assigned to add the loaders in both areas.

Sorry for the back and forth here... I have the one-line solution PR here: https://github.com/Recidiviz/justice-counts/pull/1529 with a deploy to playtesting as well to sanity check. It works fine for me in staging.

mxosman:
I'll record a loom that should hopefully explain this better than my words!

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
morden35: this explanation/solution makes sense to me!

mxosman:
https://www.loom.com/share/b657651b698b4430ac2ce3711b755b35 - OK I got cut off at 5 minutes but left a comment finishing the thought.

mxosman:
Closing in favor of #1529

PR #1557, opened 2024-10-17
[Publisher] Add "Other" textbox to race breakdown

### LINKED ISSUE #1491: [Publisher] Add "Other" free text box field for Race/Ethnicity breakdowns
For all Race/Ethnicity breakdowns, when a user selects "Other", we should display a multi-line text input field to capture their definition/clarification of the "Other" selection. The positioning of this text field can be below the "Other" checkbox for clarity.

Screenshot 2024-09-06 at 10 56 35 AM

The text input should be the same style as the one in metric definitions but with slightly less height as to not overpower the rest of the options.:

Screenshot 2024-09-06 at 10 58 26 AM

The backend work for this has still not begun, but you can assume that the endpoint and payload will be the same, except the `Other` dimension will include an additional field: `other_description` (or something similar).

nasaownsky:
## Description of the change

Added "Other" textbox to race breakdown

## Type of change

## Related issues

closes #1491 

### Development

### Code review


mxosman:
> Hi @mxosman! So after playing with this around, I think maybe injecting component into `CheckboxOptions` not the best idea? 

Ah, do you mind clarifying your thoughts on this? Is it moreso challenging to refactor the component to inject a text input? I could see us doing this more often - but just wanted to understand the challenges you ran into!

> In this case we can place textbox either before checkbox or after it, and I personally think the before solution is better, because it will always be visible to the user. What do you think?

Hmm... the biggest thing I don't like about the textbox before is the big distance between 'Other' and the textbox - it makes it feel less intuitive what that input is connected to, despite the description in the box. Ideally, it makes the most sense from a UX perspective that it happens right after the 'Other' checkbox, but if that's too technically clunky, the next best option IMO is below the checkboxes because it keeps it within the closest proximity to the 'Other' checkbox. To help with it always being visible and to enhance its connection to the 'Other' checkbox, we could add a `scrollTo` reaction that scrolls to the bottom of the container when 'Other' is checked. E.g.

```
// 
       <Styled.ScrollableInnerWrapper ref={containerRef}>
```

// move the `isOtherChecked && (...input component)` below the checkboxes

// within the `if (key === "Other") {` conditional we can add the scroll behavior

```

> Also, how do you like my implementation of `other_description` into `updatedDimensions`? Let me know if it suits our needs for this!
> 
> And last one: I was thinking about displaying existing `other_description` in textbox, and noticed that if we selecting "Yes" in `Are you able to record an individual’s ethnicity...` section -- all of the Other Ethnicities are selected, and if selecting "No" -- only Other Unknown Ethnicity is selected. Am I understanding correctly that their `other_description` should be the same after all for all ethnicities?

File: publisher/src/components/MetricsConfiguration/RaceEthnicitiesModalForm.tsx [hunk 1]
```diff
@@ -207,9 +212,25 @@ function RaceEthnicitiesModalForm({
             capture for race?
           </Styled.ToggleSwitchesListHeader>
           <Styled.CheckboxWrapper>
+            {isOtherChecked && (
+              <Styled.InputWrapper>
+                <NewInput
+                  type="text"
+                  label=""
+                  placeholder={`If the listed categories do not adequately describe this breakdown, please describe additional definition/clarification of the "Other" selection.`}
```
mxosman, line 220: Let's adjust this to just be `Please describe additional definition/clarification of the "Other" selection.`

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -1192,6 +1193,15 @@ class MetricConfigStore {
...
       (Object.values(raceEthnicitiesDimensions) as UpdatedDimension[]);
 
+    if (otherDescription) {
+      const otherDimension = dimensions.find(
+        (dimension) => dimension.race === "Other" && dimension.enabled
+      );
+      if (otherDimension) {
+        otherDimension.other_description = otherDescription;
+      }
+    }
```
mxosman, line 1203: Let me think about this section a bit more - I think it's right, but just want to make sure it's mapping to the correct Other/Ethnicity.

mxosman:
Oh, one thing I realized I forgot to include in the ticket - can we make filling out the text field a requirement if a user is going to select "Other"? Meaning, if they enter nothing, we should keep "Other" unchecked. Sorry for the last minute complexity!

publisher/src/stores/MetricConfigStore.tsx
mxosman: OK - I took a closer look. This is _almost_ there -- so basically there are 3 "Other" fields in the background of this feature: 
1. "Other" race + "Unknown Ethnicity"
2. "Other" race + "Not Hispanic or Latino" ethnicity
3. "Other" race + "Hispanic or Latino" ethnicity

What gets updated should match the grid that's displayed in the UI. So, when a user is in the...

* "Unknown Ethnicity" state (they answered "No" for being able to record an individual's ethnicity, and "Hispanic Or Latino" is NOT selected) - then it should update the "Other / Unknown Ethnicity"
* "Hispanic or Latino" state (they answered "No" for being able to record an individual's ethnicity, and "Hispanic Or Latino" IS SELECTED) - then it should update the "Other / Not Hispanic or Latino"
* "Not Hispanic or Latino" state (they answered "Yes" for being able to record an individual's ethnicity) - then it should update all 3 -  "Other / Unknown Ethnicity" AND  "Other / Not Hispanic or Latino" AND  "Other / Hispanic or Latino"

LMK if this is confusing (because it is) and I can give some more context as to why these fields work this way! But the TLDR is - the `other_description` should be set matching the blue dots in the UI depending on the state we're in if it's easier to think about it this way?

nasaownsky:
Hi @mxosman! I've updated textbox logic according to your comments, so please check out my changes and if it's working correctly!

In the meantime I only have to find a way for this to work properly:
>can we make filling out the text field a requirement if a user is going to select "Other"? Meaning, if they enter nothing, we should keep "Other" unchecked

Thank you!

mxosman:
Thanks, @nasaownsky - will dive into this tomorrow morning!

mxosman:
Nice @nasaownsky! It looks like it's working exactly as intended in all 3 cases I tested. Thank you for updating the logic + positioning of it and working through the last piece. Almost there! I'll review once everything is in place.

nasaownsky:
Hi @mxosman! I think I actually need help with this.

So, after playing with BE update @nichelle-hall provided I assume I still need to work with dimensions itself, because `RaceEthnicitiesModalForm` operate on dimension level and because of the structure of `ethnicitiesByRace`. But after updating my logic I still loose my changes after page restart even on [this](https://github.com/Recidiviz/justice-counts/issues/1491#issuecomment-2546345451) test branch, where it should be working. I have feeling that I am missing something but I can't determine what exactly. Maybe you have some ideas? Any thoughts would be much appreciated, thank you in advance!

mxosman:
> #1491

Hi @nasaownsky! I think I see the confusion here - you'll be writing to the context field in the layer above dimensions - the disaggregations. So, whatever the user enters in the "Other" Race & Ethnicities field, will be saved to and read from the `disaggregations` object w/ the key `"global/race_and_ethnicity"`.  Within this `disaggregation` object, there will be a `contexts` field which is an array that includes the `OTHER_RACE_DESCRIPTION` context that all input for "Other" will be saved to/read from. [Not to be confused with the `contexts` field within the dimensions] Does that make sense? Sorry for the confusion here! 

Instead of this: `this.dimensions[systemMetricKey][disaggregationKey][dimensionKey].contexts`
You might be able to access it via (this is rough and just to illustrate - don't use this verbatim): `this.disaggregations[systemMetricKey][disaggregationKey].contexts.find(...)` (where `disaggregationKey === "global/race_and_ethnicity"` and you're finding the `OTHER_RACE_DESCRIPTION` object)

Screenshot 2024-12-23 at 4 49 31 PM

nasaownsky:
> Within this disaggregation object, there will be a contexts field which is an array that includes the OTHER_RACE_DESCRIPTION context that all input for "Other" will be saved to/read from.

Right, but then this disaggregation context has only spot for one "Other" description, and how should we write three different versions of it there? For `Other / Unknown Ethnicity`, `Other / Not Hispanic or Latino` and `Other / Hispanic or Latino`

mxosman:
> > Within this disaggregation object, there will be a contexts field which is an array that includes the OTHER_RACE_DESCRIPTION context that all input for "Other" will be saved to/read from.
> 
> Right, but then this disaggregation context has only spot for one "Other" description, and how should we write three different versions of it there? For `Other / Unknown Ethnicity`, `Other / Not Hispanic or Latino` and `Other / Hispanic or Latino`

We've simplified it from 3 separate fields to just that one field. So any input in the "Other" just goes in that one field - you no longer have to worry about 3 separate fields as this one will cover all 3.

nasaownsky:
Hey @nichelle-hall! I've updated logic to save other description in disaggregations contexts, and it sends payload to `/api/agencies/${agencyId}/metrics`:

Снимок экрана 2024-12-30 в 16 20 51

but it seems like it is not saved in BE, because if I reload the page there is no value in the context:

Снимок экрана 2024-12-30 в 16 26 58

***
Could you please check this for me? I'm in your [test](https://github.com/Recidiviz/justice-counts/issues/1491#issuecomment-2546345451) env. Thank you so much in advance!

mxosman:
> Hey @nichelle-hall! I've updated logic to save other description in disaggregations contexts, and it sends payload to `/api/agencies/${agencyId}/metrics`:
> 
> Снимок экрана 2024-12-30 в 16 20 51
> but it seems like it is not saved in BE, because if I reload the page there is no value in the context:
> 
> Снимок экрана 2024-12-30 в 16 26 58
> Could you please check this for me? I'm in your [test](https://github.com/Recidiviz/justice-counts/issues/1491#issuecomment-2546345451) env. Thank you so much in advance!

Hi @nasaownsky - Nichelle is out until next Monday - but, I pulled in your latest changes and tried myself, and am not seeing the `contexts` in the payload - just the `key` and `dimensions`. Did you push up all of your changes? The BE saves the JSON blob and returns it too - so it should just return whatever you sent, not sure what's going on - but first, I'm not able to repro your payload.


https://github.com/user-attachments/assets/d228dfde-2b32-461c-94b2-eab0073fb6e8

nasaownsky:
Hello @mxosman! Happy new year, hope you had great holidays!

I'm also out til Wednesday, but decide to respond as I saw your message and I think I know what is going wrong for you specifically -- I think you still using default env url, try change that to `REACT_APP_PROXY_HOST=https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/` -- this is Nichelle's test env with changes to BE that we need for this PR. I double checked right now -- it sends payload to BE as expected.

mxosman:
> Hello @mxosman! Happy new year, hope you had great holidays!
> 
> I'm also out til Wednesday, but decide to respond as I saw your message and I think I know what is going wrong for you specifically -- I think you still using default env url, try change that to `REACT_APP_PROXY_HOST=https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/` -- this is Nichelle's test env with changes to BE that we need for this PR. I double checked right now -- it sends payload to BE as expected.

Happy New Year to you, Ilya! Hope you had a lovely holiday as well - and are enjoying your time off (don't feel obligated to respond on your days off). My environment was pointed to Nichelle's test URL - but the payload is coming from the frontend, and I'm not sure why I don't see it included in the PUT request. When you do `git log` - do you show the same thing as my video? Any active changes you haven't committed?

mxosman:
Oh strange, I tried it again this morning and now I see the payload. Maybe I was in the wrong env after all, but could've sworn I adjusted it and re-ran before playing with it and recording. Sorry about that! Totally see what you're seeing - let's see what Nichelle things when she returns and go from there. In the meantime, I'll do a review at some point today since this should hopefully be mostly there.

publisher/src/components/MetricsConfiguration/RaceEthnicitiesModalForm.tsx
```diff
@@ -118,12 +123,53 @@ function RaceEthnicitiesModalForm({
...
+  const currentOtherDescription =
+    disaggregations[systemMetricKey]?.[
+      RACE_ETHNICITY_DISAGGREGATION_KEY
+    ].contexts?.find((context) => context.key === "OTHER_RACE_DESCRIPTION")
+      ?.value || "";
+
+  const [otherDescription, setOtherDescription] = useState(
+    currentOtherDescription
+  );
+  const [isOtherChecked, setOtherChecked] = useState(Boolean(otherDescription));
```
mxosman, line 137: Maybe we should check for the presence of `currentOtherDescription` instead of the `otherDescription` state value for the initializer.

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -1192,12 +1203,23 @@ class MetricConfigStore {
       raceEthnicitiesDimensions &&
       (Object.values(raceEthnicitiesDimensions) as UpdatedDimension[]);
 
+    const otherDescriptionContext = this.disaggregations[systemMetricKey]?.[
+      RACE_ETHNICITY_DISAGGREGATION_KEY
+    ].contexts?.find((context) => context.key === "OTHER_RACE_DESCRIPTION");
+
+    if (otherDescription !== undefined) {
+      if (otherDescriptionContext) {
```
mxosman, line 1211: nit: could we simplify and just do something like `if (otherDescription && otherDescriptionContext)`

publisher/src/stores/MetricConfigStore.tsx
```diff
@@ -463,6 +471,8 @@ class MetricConfigStore {
     this.dimensions[systemMetricKey][disaggregationKey][
       dimensionKey
     ].description = dimensionData.description;
+    this.dimensions[systemMetricKey][disaggregationKey][dimensionKey].contexts =
+      dimensionData.contexts;
```
mxosman, line 475: I don't think we need to reference the dimension context anymore, right?

publisher/src/components/MetricsConfiguration/RaceEthnicitiesModalForm.tsx
```diff
@@ -118,12 +123,53 @@ function RaceEthnicitiesModalForm({
...
+
+  useEffect(() => {
+    setOtherDescription(currentOtherDescription);
+  }, [currentOtherDescription]);
```
mxosman, line 141: I'm wondering if we need this `useEffect`? If `currentOtherDescription` is defined inline, and used to initialize `currentOtherDescription`, when do we expect the value of `currentOtherDescription` to change for this to be useful? I worry we're going to run into some unexpected race conditions with these useEffects.

publisher/src/components/MetricsConfiguration/RaceEthnicitiesModalForm.tsx
```diff
@@ -210,10 +257,24 @@ function RaceEthnicitiesModalForm({
...
               onChange={({ key, checked }) => {
-                setRacesStatusObject({
-                  ...racesStatusObject,
-                  [key]: !checked,
-                });
+                if (key === "Other") {
+                  setOtherChecked(!checked);
+
+                  if (checked) {
+                    // Clear otherDescription if "Other" is unchecked
```
mxosman, line 264: Thanks for the comment!

nichelle-hall:
Hey Ilya,

I was testing our code end-to-end and found an issue on my side, but also noticed an error on your end. It looks like the contexts aren’t being sent over the first time—only on the second attempt when I redo the request to add another description.

I’ve uploaded a screen recording to illustrate the issue. Let me know your thoughts! 


https://github.com/user-attachments/assets/389a7fe2-1a34-47bd-98bb-d512ccd6657e


I've pushed up my bug fix to same link I sent earlier.

mxosman:
> > @nichelle-hall - do you know whether or not the backend sends an empty OTHER_RACE_DESCRIPTION context field if a user has never filled it out?
> 
> It didn't! But I just pushed up a fix so that now it does. That said, I'm still seeing the error.

Gotcha! I just tried it again w/ a few agencies and it's working for me on the first go. Would you mind trying again and triple checking your `.env` is pointed to `https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/`? Hopefully it'll work - maybe it was in a funky state before.

nichelle-hall:
I'm now seeing the data saved on the first edit! I must have been on the wrong BE branch.

nasaownsky:
@mxosman @nichelle-hall Hi all! This is utterly strange but now this is not working for me at all. When I check `  this.disaggregations[systemMetricKey]?.[RACE_ETHNICITY_DISAGGREGATION_KEY]` in `MetricConfigStore.tsx` it shows me that `contexts` is undefined. I triple checked .env (it's `https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/`), reinstalled dependencies, cleared all the caches -- but nothing had effect. It seems like I don't receive contexts from BE. It's especially strange since it's working for you, that I have this issue. Maybe you have any ideas?

nichelle-hall:
> @mxosman @nichelle-hall Hi all! This is utterly strange but now this is not working for me at all. When I check ` this.disaggregations[systemMetricKey]?.[RACE_ETHNICITY_DISAGGREGATION_KEY]` in `MetricConfigStore.tsx` it shows me that `contexts` is undefined. I triple checked .env (it's `https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/`), reinstalled dependencies, cleared all the caches -- but nothing had effect. It seems like I don't receive contexts from BE. It's especially strange since it's working for you, that I have this issue. Maybe you have any ideas?

You know what! I think I made a mistake and wrote over some changes on my playtesting branch. I'l redeploy now!

nasaownsky:
@nichelle-hall It's working now, thank you!

nasaownsky:
@mxosman thank you for great comments, now it should be ready!

mxosman:
Whew - thank you for the changes! I was sooo close to approving, but ran into a strange issue specifically with Jails > Total Population Race & Ethnicity metric - for some reason this isn't sending a context object - can you take another look? [Super confirmed that I'm using the right - which is Nichelle's playtest link]

https://github.com/user-attachments/assets/0fa647d3-1fdf-496b-974d-f6659e242d1a

@nichelle-hall - wondering if you can take a look too to see if you can repro and if there might be something in the BE related to this particular metric?

mxosman:
Hi @nasaownsky/@nichelle-hall - oh strange - there is something weird going on with that agency... I'm getting a server error too.

You should still be able to test this issue with a different agency or by creating a new agency... Can you try that, please? You can point the environment to regular `publisher-staging` URL and not the `nichelletest` since the backend changes were merged in.

mxosman:
OK, hello again @nasaownsky & @nichelle-hall. I did some digging and it looks like for the metrics that are affected (metrics in Jails and Prisons, and probably some others I haven't tested yet) that the backend is not sending over the OTHER_RACE_DESCRIPTION context field. 

@nichelle-hall - I saw that you added these in to the TIGs in your #36052 PR - any hunches as to why it's not sending them over? You can try this agency: http://localhost:3000/agency/1693/metric-config?system=jails&metric=jails_total_population - if you look at the network response when you load the Total Daily Populations metric, it doesn't show anything with a `OTHER_RACE_DESCRIPTION` key.

@nasaownsky - is it necessary for the FE to have this field before it's able to send it in the PUT request? If not, is there a way that we can just always send this context field regardless of whether or not the backend sends it initially?

nasaownsky:
@mxosman I played around with implementation and I think we could eliminate this issue on FE side by introducing `defaultOtherDescriptionContext` object. Tested with newly created Jails agencies and it seems like the issue you describing is gone!

nasaownsky:
@mxosman Thank you for your detailed investigation! I played around with implementation and I think we could eliminate this issue on FE side by introducing defaultOtherDescriptionContext object. Tested with newly created Jails agencies and it seems like the issue you describing is gone!

mxosman:
> @mxosman Thank you for your detailed investigation! I played around with implementation and I think we could eliminate this issue on FE side by introducing defaultOtherDescriptionContext object. Tested with newly created Jails agencies and it seems like the issue you describing is gone!

Great call! I was hoping we could do exactly this - just send it anyway and not depend on the BE giving it to us. Thanks for the adjustment. One small issue I noticed - the "Other" doesn't check after you save:

https://github.com/user-attachments/assets/add482e1-7b9b-4477-9b00-d9866a9348ec

Could you take another look and spend a bit more time manually testing different agencies/metrics to make sure things are working as intended?

nichelle-hall:
@mxosman @nasaownsky 

I know you have a plan of attack, but I'm also looking into a fix. The BE should be sending the contexts over!

nasaownsky:
>One small issue I noticed - the "Other" doesn't check after you save

@mxosman Hmm, this one actually could be tricky without the context. Let me think about it some more.

@nichelle-hall Oh, great to hear! I think it should fix that problem for good, with the context sending from BE.

mxosman:
> @nasaownsky @mxosman
> 
> I've found the bug and pushed up the changes to `https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/`.

Ah, amazing - that fixes everything! @nasaownsky & @nichelle-hall - thank you both SO MUCH for all your work on this - what a journey! I did another round of tests and I couldn't break it. I'm glad we're finally able to ship this! 👏 👏 👏 

We'll wait for the backend fix to go through review and get merged in before I give this one the green light to make sure everything is in sync - but consider this approved. Thank you both again so much!

nasaownsky:
Yaaaay! 🚀

nichelle-hall:
The BE fix has been merged!

mxosman (approved):
Thank you @nichelle-hall! Good to go!

PR #1558, opened 2024-10-17
[Publisher][Admin Panel] Fix create new users via secondary flow bugs

### LINKED ISSUE #1442: [Publisher][Admin Panel][Bug] Creating Multiple New Users via Secondary Modal Flow
When creating/updating an agency in the admin panel, the secondary modal (Create New User) flow has a small, but noticeable issue when creating more than one user. After creating the first user, that first user gets added to the Team Members & Roles w/ a default role. But, after creating a second user, the first user is no longer on the Team Members & Roles tab and is instead overwritten by the second user. I believe that both users save - but the UI does not reflect that.

To reproduce: in admin panel create new agency and add agency info. Click team members & roles and create new user. Add user info click save. New user appears in list of team members. Click create new user again then add an additional new user. Add user info a save. 2nd new user shows up in list of team members, but first added user does not. When actually viewing the agency after saving, both new users appear to be have been set up correctly - so the issue is just the first user no longer appearing on the team members pane in the create agency pop up.

nasaownsky:
## Description of the change

> Description here

## Type of change

## Related issues

closes #1442 

### Development

### Code review


mxosman (commented):
Oh, nice! I'm glad you were able to quickly figure out what was going on! This one stumped me when I was looking at it a while back - let me spend some more time ramping back up on this.

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
```diff
@@ -608,7 +608,7 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
         {} as UserRoleUpdates
       );
 
-      if (!selectedAgency) {
+      if (!selectedAgency && !secondaryCreatedId) {
```
mxosman, line 611: Can you help me understand this small change? Does this only affect the Create Agency flow?  I see that we're skipping the auto adding of CSG team members feature if we're in the secondary Create New User flow, but am worried that we're having a clashing of the useEffects.

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
```diff
@@ -117,6 +117,7 @@ export const AgencyProvisioningOverview = observer(() => {
     if (!activeSecondaryModal) {
       setSelectedAgencyID(undefined);
       resetAgencyProvisioningUpdates();
+      setUserId(undefined);
```
mxosman, line 120: Ah nice catch!

mxosman (commented):
Wow - thank you so much for figuring this out! I left a couple questions. Can I also ask you to also add some unit tests to make sure that this flow is working as intended? One via Create Agency and the other via an existing agency - checking that the newly added users + existing users persist after saving.

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
```diff
@@ -638,9 +638,11 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
     useEffect(() => {
       const newMember = users.find((user) => user.id === secondaryCreatedId);
       if (secondaryCreatedId && newMember) {
-        setSelectedTeamMembersToAdd((prev) =>
-          toggleAddRemoveSetItem(prev, +secondaryCreatedId)
-        );
+        setSelectedTeamMembersToAdd((prev) => {
+          const newSet = new Set(prev);
+          newSet.add(+secondaryCreatedId);
+          return newSet;
```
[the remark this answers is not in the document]
nasaownsky: Yeah, I also thought it should work, but it isn't, I believe this was the problem with `delete` part of `toggleAddRemoveSetItem` -- we don't need it in this case.

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
nasaownsky: This part actually rewrites `selectedTeamMembersToAdd` every time we add new user, and because of that we loose previously added one. I believe this will not affect other useEffects because we only need this to be called once anyway.

mxosman: Gotcha! This must've been re-running any time a new user was added via the secondary flow as well.

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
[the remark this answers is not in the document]
mxosman: Ah, hmm... that's so strange that the `delete` would interfere. Let me think about this some more - because I imagine that the `secondaryCreatedId`s are different for each newly created user, so it shouldn't have to delete.

mxosman (commented):
Thank you for the clarifications, Ilya! I just want to wrap my head around the delete portion of the `toggleAddRemoveSetItem`. If you could please add some integration tests to make sure this flow is working as intended and doesn't drop existing or newly created users - and works for both Create and Updating an agency, I'd greatly appreciate it! This is a critical flow, so want to make sure we're covered.

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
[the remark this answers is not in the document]
nasaownsky: Yes, it is true, but due to the complicated structure of `AgencyProvisioning` and usage of different useEffects in it I believe previous `secondaryCreatedId` might be sended to `toggleAddRemoveSetItem`. Because of this it is getting deleted, and then the new id (updated one) got written to the Set. I'm pretty sure that is the reason. 

Using it this updated way we make sure the Set contains all the newly created users.

nasaownsky:
Glad to help, @mxosman! I already work on that coverage!

mxosman:
> Glad to help, @mxosman! I already work on that coverage!

I don't see the tests in this PR - maybe I misunderstood your comment?

nasaownsky:
> > Glad to help, @mxosman! I already work on that coverage!
> 
> I don't see the tests in this PR - maybe I misunderstood your comment?

I mean I already working on tests, sorry for misunderstanding!

mxosman:
> > > Glad to help, @mxosman! I already work on that coverage!
> > 
> > 
> > I don't see the tests in this PR - maybe I misunderstood your comment?
> 
> I mean I already working on tests, sorry for misunderstanding!

Oh oh!! Sorry! Gotcha now! 😄 Thank you so much!

publisher/src/components/AdminPanel/AdminPanel.test.tsx
```diff
@@ -884,3 +884,64 @@ test("Deleting a user deletes a card to the list of team members", async () => {
...
+  const emailInput1 = await screen.findByLabelText("Email");
+  fireEvent.change(emailInput1, { target: { value: "user13@email.com" } });
+
+  const saveButton1 = screen
+    .getAllByText("Save")
+    .find((button) => button.hasAttribute("disabled"));
+  await waitFor(() => {
+    expect(saveButton1).not.toHaveAttribute("disabled");
+    if (saveButton1) fireEvent.click(saveButton1);
+  });
```
nasaownsky, line 927: Hi @mxosman! I stumble across the problem -- when trying to save changes via secondary flow we have "Save" button in disabled state if fields not populated, but it seems like testing library never updates this disabled state and thus I can't click on "Save" button. Do you have any ideas on how can I overcome this problem?

I'm trying to use `waitFor` here, but so far no results with it, although logically it should have helped with this. Thank you in advance for any help!

mxosman: Ah, hmm... so there are probably 2 save buttons in the DOM. One for the first modal and the other for the Create New User modal. The logic in L921-923 searches for the one with a `disabled` property, and your `waitFor` is checking to see if that `disabled` element does NOT have a disabled property - when it always will be disabled since we're `find`ing the one that is disabled (first modal).

So, I think if you target the correct Save button, it should hopefully not have the disabled property after you entered in the proper inputs. LMK if this makes sense/helps!

nasaownsky: The thing is the first modal's button never gets disabled, so when I do `getAllByText` it is returning me array with two buttons, one of which is disabled. So I assume that this disabled button is from the second modal. But it is disabled. So I assume `fireEvent` won't have any effect on it? So I actually have no idea how to handle this.

mxosman: Hmm, maybe I'm missing something - please let me know if I'm not making sense.

If you go through the flow yourself outside of this test environment - via Agency Provisioning tab, and opening an agency - the agency modal will have the Save button disabled because there are no changes (similar to this test environment - no changes were made on the Agency modal) - can you confirm you're able to see this?

More importantly, the logic doesn't quite make sense because we're searching for a button that has the disabled attribute, and on the next lines expecting it to not have the disabled attribute.

See added comments:
```
   // Here, we are searching for the button that HAS a disabled attribute
  const saveButton1 = screen
    .getAllByText("Save")
    .find((button) => button.hasAttribute("disabled")); 

  await waitFor(() => {
   // Then, we're somehow expecting the button that we found that HAS a disabled attribute to no longer have it?
    expect(saveButton1).not.toHaveAttribute("disabled"); 
    if (saveButton1) fireEvent.click(saveButton1);
  });
```

Let me know if this makes sense or if I'm missing something. Ultimately, I think you're looking to target the second Save button because that one should not be disabled because of your `fireEvent.change` that filled out the required fields.

mxosman: Hmm... it's probably because of the setTimeout in the modal after saving... Oi. I think it's unavoidable to wait for the saving setTimeout - so try doing something like this:
`  const nameInput2 = await screen.findByLabelText("Name", {}, { timeout: 2000 } )`

publisher/src/components/AdminPanel/AdminPanel.test.tsx
```diff
@@ -912,36 +912,45 @@ test("Adding users via secondary flow", async () => {
...
 
-  fireEvent.click(screen.getByText("Create User"));
-  const nameInput2 = await screen.findByLabelText("Name");
-  fireEvent.change(nameInput2, { target: { value: "user14" } });
-  const emailInput2 = await screen.findByLabelText("Email");
-  fireEvent.change(emailInput2, { target: { value: "user14@email.com" } });
+  const asideElement = screen.getAllByRole("complementary")[2];
+  // Sorry, there was an issue creating a user. Please try again.
+  // eslint-disable-next-line testing-library/no-debugging-utils
+  screen.debug(asideElement);
```
nasaownsky, line 932: @mxosman Strangely this flow doesn't work here in test environment. Maybe I forget about some store properties?

mxosman:
Hi @nasaownsky - thank you again so much for trying to wrangle writing the tests for this and for the time you spent on it. It's not worth spending more time on it, I'm good with us reverting the test changes here and merging this in. Thank you for testing it manually E2E. I did the same with both the existing agency and new agency flows and can confirm it's working as expected. I agree with the current structure that this is challenging to test. I assumed that we could reference the `AdminPanelStore.agencyProvisioningUpdates.team` to confirm the changes - and maybe we can try some other time - but I feel like that might not work as expected since the newly created user objects are only sending the user id and the role.

Anyway - it was worth a shot, but let me know once things are reverted and I'll accept this!

nasaownsky:
Should be a-ok now @mxosman!

mxosman (approved):

PR #1615, opened 2024-12-17
[Publisher][Metric Settings] Update metric definitions to support radio input

### LINKED ISSUE #1602: [Publisher][Metric Settings] Supervision sectors: Update Revocations metric definitions checkboxes to behave as radio inputs (only allowing one selection) for a new definition prompt)
For Supervision sectors, the Revocations metric definition (see below screenshot) will have a new set of questions:

Prompt: If reporting breakdowns for the Revocations metric, please select what your data represents:

Choices:
* Data reflects the most recent violation type
* Data reflects the most serious violation type in a person’s supervision term

Reference definition that will be updated: 
Image

The backend will add this new question with the following `description` property as the "key": `"Revocation metric data type"`

We need to update the checkbox behavior to allow the user to _only_ select one or the other and not both - this behavior _only_ applies to this question for this metric definition, so we'll need to update the logic to listen for the above `description` key in the `includes_excludes` property and adjust its behavior to allow the user to only choose one of the options.

This will be hacky and gnarly, but the only way I can think of currently to satisfy this new requirement for these new sets of questions. Let me know if anything feels confusing or where I can clarify further! @nichelle-hall will update the backend to send these new definition questions to the FE, so most of it should work automatically out of the box - it's just the behavior of the checkboxes that need to adjust for just this new question.

nasaownsky:
## Description of the change

Updated metric definitions and checkbox group logic to support radio input

## Type of change

## Related issues

closes #1602 

### Development

### Code review

### DIFF
--- a/common/components/CheckboxOptions/CheckboxOptions.tsx
+++ b/common/components/CheckboxOptions/CheckboxOptions.tsx
@@ -49,20 +49,9 @@ export const CheckboxOptions: React.FC<CheckboxOptionsProps> = ({
   options,
   onChange,
 }) => {
-  const [radioKey, setRadioKey] = useState<string>("");
   const [otherDescriptionValue, setOtherDescriptionValue] = useState<
     string | null
   >(null);
-
-  const handleChange = (key: string, checked: boolean) => {
-    if (multiselect) {
-      onChange({ key, checked });
-    } else {
-      const newKey = radioKey !== key ? key : "";
-      setRadioKey(newKey);
-      onChange({ key: newKey, checked: true });
-    }
-  };
 
   return (
     <Styled.CheckboxContainer>
@@ -81,11 +70,11 @@ export const CheckboxOptions: React.FC<CheckboxOptionsProps> = ({
               <Styled.Checkbox
                 id={key}
                 type={multiselect ? "checkbox" : "radio"}
-                checked={multiselect || !radioKey ? checked : radioKey === key}
+                checked={checked}
                 onChange={() =>
                   onChangeOverride
                     ? onChangeOverride()
-                    : handleChange(key, checked)
+                    : onChange({ key, checked })
                 }
                 disabled={disabled}
               />
--- a/common/components/CheckboxOptions/CheckboxOptions.tsx
+++ b/common/components/CheckboxOptions/CheckboxOptions.tsx
@@ -39,17 +39,30 @@ export type CheckboxOption = {
 export type CheckboxOnChangeParams = { key: string; checked: boolean };
 
 export type CheckboxOptionsProps = {
+  multiselect?: boolean;
   options: CheckboxOption[];
   onChange: (onChangeParams: CheckboxOnChangeParams) => void;
 };
 
 export const CheckboxOptions: React.FC<CheckboxOptionsProps> = ({
+  multiselect = true,
   options,
   onChange,
 }) => {
+  const [radioKey, setRadioKey] = useState<string>("");
   const [otherDescriptionValue, setOtherDescriptionValue] = useState<
     string | null
   >(null);
+
+  const handleChange = (key: string, checked: boolean) => {
+    if (multiselect) {
+      onChange({ key, checked });
+    } else {
+      const newKey = radioKey !== key ? key : "";
+      setRadioKey(newKey);
+      onChange({ key: newKey, checked: true });
+    }
+  };
 
   return (
     <Styled.CheckboxContainer>
@@ -67,12 +80,12 @@ export const CheckboxOptions: React.FC<CheckboxOptionsProps> = ({
             <Styled.CheckboxOptionsWrapper>
               <Styled.Checkbox
                 id={key}
-                type="checkbox"
-                checked={checked}
+                type={multiselect ? "checkbox" : "radio"}
+                checked={multiselect || !radioKey ? checked : radioKey === key}
                 onChange={() =>
                   onChangeOverride
                     ? onChangeOverride()
-                    : onChange({ key, checked })
+                    : handleChange(key, checked)
                 }
                 disabled={disabled}
               />
--- a/publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
+++ b/publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
@@ -164,17 +164,18 @@ function DefinitionModalForm({
   // handlers
   const handleChooseDefaults = () => {
     const defaultSettings = Object.entries(currentSettings).reduce(
-      (acc, [includesExcludesKey, { settings }]) => {
+      (acc, [includesExcludesKey, { multiselect, settings }]) => {
         return {
           ...acc,
           [includesExcludesKey]: {
+            multiselect,
             settings: Object.entries(settings).reduce(
               (innerAcc, [settingKey, setting]) => {
                 return {
                   ...innerAcc,
                   [settingKey]: {
                     ...setting,
-                    included: setting.default,
+                    included: multiselect ? setting.default : "No",
                   },
                 };
               },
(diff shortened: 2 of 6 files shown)


nasaownsky:
Hi @nichelle-hall @mxosman! 

One small nit: when `"multiselect": false`, I assume we should have `"default": "Yes"` only for one setting, because radio only allowing one selection. Currently BE returns `"default": "Yes"` for both settings.

nichelle-hall:
> Hi @nichelle-hall @mxosman!
> 
> One small nit: when `"multiselect": false`, I assume we should have `"default": "Yes"` only for one setting, because radio only allowing one selection. Currently BE returns `"default": "Yes"` for both settings.

Oh, good point! I'll look into that, that must be a bug.

mxosman:
Ah, great catch @nasaownsky - and thank you @nichelle-hall for working on a fix!

common/components/CheckboxOptions/CheckboxOptions.tsx
```diff
@@ -39,18 +39,31 @@ export type CheckboxOption = {
...
     string | null
   >(null);
 
+  const handleChange = (key: string, checked: boolean) => {
+    if (multiselect) {
+      onChange({ key, checked });
+    } else {
+      const newKey = radioKey !== key ? key : "";
+      setRadioKey(newKey);
+      onChange({ key: newKey, checked: true });
```
mxosman, line 63: So I tested this and the switch to radio buttons looks and works beautifully - I think my brain is a bit fried so I'll give this a better review in the morning, but can you help me understand what we're doing with the whole `radioKey` strategy in general and specifically here in this `else` block? I'm not 100% following.

publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
```diff
@@ -193,25 +194,46 @@ function DefinitionModalForm({
...
+        ...currentSettings,
+        [includesExcludesKey]: {
+          ...currentSettings[includesExcludesKey],
+          settings: Object.fromEntries(
+            Object.entries(currentSettings[includesExcludesKey].settings).map(
+              ([key, setting]) => [
+                key,
+                {
+                  ...setting,
+                  included: key === settingKey ? "Yes" : "No",
```
mxosman, line 211: Oi - what a doozy! So if I'm understanding this correctly - if we're selecting an option, this method gets run and passes in a settingKey (the setting we just selected) - and here we map through all of them and only sets the one that matches the setting we just clicked on to "Yes", and all others to "No", right?

common/components/CheckboxOptions/CheckboxOptions.tsx
nasaownsky: Basically, if we in `multiselect` we preserve old `onChange` behavior, and in the else block we just store our current key (since radio should only handle one selection), and the return this key and `true` for checked, since it'll always be checked (meaning we can't uncheck it by clicking on itself).

publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
nasaownsky: That is correct!

nichelle-hall:
> > Hi @nichelle-hall @mxosman!
> > One small nit: when `"multiselect": false`, I assume we should have `"default": "Yes"` only for one setting, because radio only allowing one selection. Currently BE returns `"default": "Yes"` for both settings.
> 
> Oh, good point! I'll look into that, that must be a bug.
> 
> Update: CSG didn't specify that either of them should be checked to start. So they will both be "No". I'll let you know once that change is merged.

@nasaownsky, the change is merged!

common/components/CheckboxOptions/CheckboxOptions.tsx
mxosman: Goootcha! For clarity, can we update the `radioKey` var name to be `selectedRadioOptionKey`?

nasaownsky:
@nichelle-hall great, thank you! 

Do I need to do some additional actions to see the change? I still get "Yes" as defaults: 

Снимок экрана 2024-12-18 в 19 04 19

common/components/CheckboxOptions/CheckboxOptions.tsx
```diff
@@ -67,12 +80,12 @@ export const CheckboxOptions: React.FC<CheckboxOptionsProps> = ({
             <Styled.CheckboxOptionsWrapper>
               <Styled.Checkbox
                 id={key}
-                type="checkbox"
-                checked={checked}
+                type={multiselect ? "checkbox" : "radio"}
+                checked={multiselect || !radioKey ? checked : radioKey === key}
```
mxosman, line 84: I think this will evaluate as follows based on the operator precedence: `(multiselect || !radioKey) ? checked : radioKey === key` - just double checking that's what we want, and NOT `multiselect || (!radioKey ? checked : radioKey === key)`, right?

nasaownsky: Yeah, first option -- for second option we need to provide parentheses as you described.

mxosman (approved):
This looks great to me, @nasaownsky! Giving early approval - if you could please do a test once @nichelle-hall deploys to her playtesting before merging to sanity check I'd greatly appreciate it. Thank you for helping us cross the finish line with this one!

nichelle-hall:
> > @nasaownsky, the change is merged!
> 
> > @nichelle-hall great, thank you!
> > Do I need to do some additional actions to see the change? I still get "Yes" as defaults:
> 
> Thank you @nichelle-hall! Would you mind deploying the backend `main` and this branch to your playtesting?
> 
> @nasaownsky - the change is merged in but not yet deployed to regular staging. Once @nichelle-hall deploys to her playtesting link, you can update your `env` to point to that link and you'll be able to test the changes.

@mxosman oops! Forgot to deploy to playtesting. @nasaownsky The changes are now available on `nicelletest---publisher-web-b47yvyxs3q-uc.a.run.app`

nasaownsky:
@mxosman @nichelle-hall I'd like to test, but with updated `env` the app just throws me `No authentication configuration detected. Please refresh the page or try again later.` screen. Any ideas maybe?

nasaownsky:
@mxosman I was testing changes with @nichelle-hall updates to BE and stumble across annoying bug with radio selection, but then I realized that we actually don't need any logic inside `CheckboxOptions` since we controlling checked state  from the outside in `DefinitionModalForm`. New changes working flawlessly -- but please test it yourself once more and tell me your thoughts about it! Thank you!

mxosman:
> @mxosman I was testing changes with @nichelle-hall updates to BE and stumble across annoying bug with radio selection, but then I realized that we actually don't need any logic inside `CheckboxOptions` since we controlling checked state from the outside in `DefinitionModalForm`. New changes working flawlessly -- but please test it yourself once more and tell me your thoughts about it! Thank you!

Ah - new logic makes much more sense. Thanks for adjusting that. Out of curiosity, what was the bug you ran into? I didn't notice anything with the old logic.

Changes work fine for me! This is good to go - thank you, @nasaownsky!

nasaownsky:
> Out of curiosity, what was the bug you ran into? I didn't notice anything with the old logic.

Basically, when selecting default values via `Justice Counts definition` we manipulating with `checked` value of `CheckboxOption` from the outside, but in the new logic because of this code -- `checked={multiselect || !radioKey ? checked : radioKey === key}` -- when not in `multiselect` there were troubles with controlling the state of `checked` both for selecting defaults and manual selecting, because it only took into account keys, but not `checking` state from the outside. And combining both of this states logically was creating some cases when radio input inconsistently required two clicks to select.

mxosman:
Goootcha! Thanks for explaining that!

PR #1628, opened 2025-01-02
[Publisher][Bug] Fix bug where Explore Data page shows first Combined Supervision datapoints when Supervision sector is disaggregated

### LINKED ISSUE #1629: [Publisher][Bug] Explore Data: Incorrectly displaying a disabled metric's charts
Bug Report from CSG: 

Agency 1474 in Publisher has a bug under the Explore Data Page for the Funding metric.

Problem: When I click on the Explore Data page, the Funding metric appears with data from CY2022 for Supervision. However, the agency turned off Supervision and is only reporting for Other Supervision (all of their metrics should be under Other Supervision).
When we click on the Explore Data page, we shouldn't see data for Supervision Funding. However, Publisher is showing us data for a metric that is turned off.

---

The issue is for Supervision sectors who have a Supervision (Combined) metric value in a report and later disaggregate it, the Explore Data page undesirably displays the first disabled Supervision (Combined) metric on load instead of the first enabled metric (which would be one of the disaggregated Supervision metrics, e.g. - Parole/Other Supervision/etc.)

mxosman:
## Description of the change

Fix bug where Explore Data page shows first Combined Supervision datapoints when Supervision sector is disaggregated. Bug was reported by CSG where an agency that has the Supervision sector disaggregated shows a Supervision Combined Funding metric's datapoints by default when first visiting the Explore Data page. This is caused by the current implementation relying on the `system` param which is `undefined` on load and subsequently gets set as the first agency's system instead of the first system with an enabled metric. 

The change proposed here sets the initial search params after initializing the report settings (our current way to infer the enabled/disabled status of a metric for now), which more accurately displays the charts for the first enabled metric and avoids the issue caused by the assumption that the first system in a user's agency includes an enabled metric. 

This component would benefit greatly from a deeper refactor to avoid future risk of race conditions with how we're retrieving metrics to determine the enabled/disabled status and how we're subsequently setting the search params. For now, this should unblock CSG and consistently produce the expected behavior until we can dedicate more time to exploring a more holistic refactor.

## Type of change

## Related issues

Closes #1629

### Development

### Code review

### DIFF
--- a/publisher/src/components/DataViz/MetricsDataChart.tsx
+++ b/publisher/src/components/DataViz/MetricsDataChart.tsx
@@ -32,7 +32,6 @@ import {
 } from "@justice-counts/common/components/GlobalStyles";
 import { useWindowWidth } from "@justice-counts/common/hooks";
 import {
-  AgencySystem,
   AgencySystems,
   ReportFrequency,
   SupervisionSubsystems,
@@ -103,7 +102,7 @@ export const MetricsDataChart: React.FC = observer(() => {
   const currentMetric = currentSystem
     ? metricsBySystem[currentSystem]?.find(
         (m) => m.key === (metricSearchParam || enabledMetrics[0]?.key)
-      ) || metricsBySystem[currentSystem]?.[0]
+      ) || enabledMetrics[0]
     : undefined;
   const metricName = currentMetric?.display_name || "";
   const metricFrequency =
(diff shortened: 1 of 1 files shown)


publisher/src/components/DataViz/MetricsDataChart.tsx
```diff
@@ -276,11 +276,34 @@ export const MetricsDataChart: React.FC = observer(() => {
...
+                              setSettingsSearchParams({
+                                system: system as AgencySystem,
+                                metric: metric.key,
+                              })
+                            }
+                          >
+                            {metric.display_name}
+                          </Styled.MetricItem>
+                        ))}
+                      </Styled.MetricsItemsContainer>
```
mxosman, line 305: Moves `MetricsItemsContainer` to the `currEnabledMetrics.length > 0` case so it doesn't show when there are no enabled metrics.

publisher/src/components/DataViz/MetricsDataChart.tsx
```diff
@@ -167,6 +167,18 @@ export const MetricsDataChart: React.FC = observer(() => {
...
+      const refreshedEnabledMetrics = Object.values(result)
+        .flatMap((metric) => metric)
+        .filter((metric) => metric.enabled);
+      const firstEnabledSystemKey = refreshedEnabledMetrics[0]?.system.key;
+      const firstEnabledMetricKey = refreshedEnabledMetrics[0]?.key;
+
+      setSettingsSearchParams({
+        system: firstEnabledSystemKey,
+        metric: firstEnabledMetricKey,
+      });
```
mxosman, line 181: This is the biggest change - once we kick off initialization, we can update the params with the first enabled system and metric key, greatly simplifying the `currentSystem` and `currentMetric` logic in L102, 104 respectively.

Overall, this component still suffers a bit from the round-about way we're inferring the enabled metrics (querying the `ReportStore` instead of from the datapoints themselves), so a more holistic solution and refactor should be explored.

helperbot-recidiviz:

danawillow (approved):
Stamping!

mxosman:
Thank you, Dana!

PR #1510, opened 2024-09-18
[Publisher] Require completion of free text description when enabling "Other" breakdown

nasaownsky:
## Description of the change

Require "Other" breakdown description in order to enable it

## Type of change

> All pull requests must have at least one of the following labels applied (otherwise the PR will fail):

| Label                       	| Description                                                                                               	|
|-----------------------------	|-----------------------------------------------------------------------------------------------------------	|
| Type: Bug                   	| non-breaking change that fixes an issue                                                                   	|
| Type: Feature               	| non-breaking change that adds functionality                                                               	|
| Type: Breaking Change       	| fix or feature that would cause existing functionality to not work as expected                            	|
| Type: Non-breaking refactor 	| change addresses some tech debt item or prepares for a later change, but does not change functionality    	|
| Type: Configuration Change  	| adjusts configuration to achieve some end related to functionality, development, performance, or security 	|
| Type: Dependency Upgrade      | upgrades a project dependency - these changes are not included in release notes                             |

## Related issues

closes #1426 

### Development

**This box MUST be checked by the submitter prior to merging**:
- [ ] **Double- and triple-checked that there is no Personally Identifiable Information (PII) being mistakenly added in this pull request**

These boxes should be checked by the submitter prior to merging:
- [ ] Tests have been written to cover the code changed/added as part of this pull request

### Code review

These boxes should be checked by reviewers prior to merging:

- [ ] This pull request has a descriptive title and information useful to a reviewer
- [ ] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [ ] Potential security implications or infrastructural changes have been considered, if relevant

PR #1639, opened 2025-01-16
[Publisher] Fix "Other" checking and unchecking behavior in Metric Settings

### LINKED ISSUE #1621: [Bug] Metric Settings: "Other" checking and unchecking behavior not working consistently
There appears to be a bug with how the "Other" checkbox is behaving. In some instances, the "Other" is automatically checked when there are no definitions available. And, in others, it appears to not save the checking and entry of the other definition. Please see videos below demonstrating buggy behavior:

https://github.com/user-attachments/assets/6c1737fe-20c4-42be-ae27-ee01baa6e45a

https://github.com/user-attachments/assets/df0e5bb8-ef8a-4ccb-a8e7-d0925f9ca9a3

nasaownsky:
## Description of the change

Fixed "Other" checking and unchecking behavior in Metric Settings page

## Type of change

## Related issues

closes #1621 

### Development

### Code review

### DIFF
--- a/publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
+++ b/publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
@@ -419,7 +419,7 @@ function DefinitionModalForm({
     ? metrics[systemMetricKey]?.description
     : currentDimension?.description;
 
-  const handleOtherDimensionUpdate = () => {
+  const handleOtherDimensionUpdate = async () => {
     const otherDimensionKey =
       currentDimensions && getOtherDimensonKey(currentDimensions);
 
@@ -435,12 +435,11 @@ function DefinitionModalForm({
         otherDimensionKey,
         !!currentContexts.ADDITIONAL_CONTEXT.value
       );
-      saveMetricSettings(updatedSetting, agencyId);
+      await saveMetricSettings(updatedSetting, agencyId);
     }
   };
 
   const handleCloseModal = () => {
-    handleOtherDimensionUpdate();
     closeModal();
   };
 
@@ -567,7 +566,8 @@ function DefinitionModalForm({
           {!isReadOnly && !hasNoSettingsAndNoContext && (
             <Button
               label="Save"
-              onClick={() => {
+              onClick={async () => {
+                await handleOtherDimensionUpdate();
                 handleSaveSettings();
                 handleCloseModal();
               }}
--- a/publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
+++ b/publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
@@ -297,6 +297,46 @@ function DefinitionModalForm({
             contextsToSave.push({ key, value });
           });
         }
+
+        // below we updating any "Other" dimension separately because it depends on ADDITIONAL_CONTEXT value
+        const otherDimensionKey =
+          currentDimensions && getOtherDimensonKey(currentDimensions);
+        if (
+          otherDimensionKey &&
+          activeDimensionKey === otherDimensionKey &&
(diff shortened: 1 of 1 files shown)


nasaownsky:
Hi @mxosman! Please test this changes and see if my fix solved the problem for you! I tried many different approaches, but it seems like only this particular solution is fixing it for me. 

Note: there could be already checked "other" dimensions (before [this](https://github.com/Recidiviz/justice-counts/pull/1510) update) -- in that case they'll be checked without description since there is no way for FE to uncheck them from BE passively. I think this explains why "other" dimension is checked in the start of the second video from the task description (Prisons II agency).

nasaownsky:
The thing I don't like about this solution is that the "Settings saved" toast popups twice when we hit "Save" button. After thinking some more about this and trying to find another solutions this is the only one that 100% solves the bug for me.

mxosman:
Hi @nasaownsky - digging into it this now.

> there could be already checked "other" dimensions 
That makes sense - I'm OK with the behavior there.

>The thing I don't like about this solution is that the "Settings saved" toast popups twice when we hit "Save" button
Hmm, yeah - I'm not a fan of that b/c that means it's sending two requests each time. Any idea why it's sending two requests?

Also separately, but potentially related - one thing I'm noticing too is that clicking the X button to close a modal sends a request, which might be contributing to some strange behavior. Do you know why that's happening? I thought the close modal code just updated local state.

mxosman:
>Also separately, but potentially related - one thing I'm noticing too is that clicking the X button to close a modal sends a request, which might be contributing to some strange behavior. Do you know why that's happening? I thought the close modal code just updated local state.

Ah - I see you already addressed this.

publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
```diff
@@ -567,7 +566,8 @@ function DefinitionModalForm({
           {!isReadOnly && !hasNoSettingsAndNoContext && (
             <Button
               label="Save"
-              onClick={() => {
+              onClick={async () => {
+                await handleOtherDimensionUpdate();
```
mxosman, line 570: Ah, we're calling `saveMetricSettings` twice with these two methods... What's the motivation behind awaiting this? Is there no way to consolidate these two save methods to avoid two redundant requests and still solve the issue?

mxosman: In the `handleSaveSettings` - I see we're referencing and looping through the `currentSettings` object - could we update that local state when folks update the Other Dimension so it's picked up by the `saveMetricSettings` in the `handleSaveSettings` function? 

Oi - the `DefinitionModalForm` file is quite convoluted - thanks for swimming in this...

nasaownsky:
Hi @mxosman! I dug deeper into saving logic and came up with the solution, that should address both inconsistent saving behavior and sending two separate requests issues. Please, check my changes and see if it works as expected now. I tested it under different conditions and it resolved the problem for me 100%.

publisher/src/components/MetricsConfiguration/DefinitionModalForm.tsx
```diff
@@ -297,6 +297,46 @@ function DefinitionModalForm({
...
+          activeDimensionKey === otherDimensionKey &&
+          currentContexts.ADDITIONAL_CONTEXT?.value !== undefined
+        ) {
+          updateDimensionEnabledStatus(
+            systemSearchParam,
+            metricSearchParam,
+            activeDisaggregationKey,
+            otherDimensionKey,
+            !!currentContexts.ADDITIONAL_CONTEXT.value
+          );
```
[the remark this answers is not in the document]
mxosman: Ah, gotcha! Thank you for clarifying!!

mxosman (approved):
Looks and works great, @nasaownsky! Thank you so much for this fix!

PR #1655, opened 2025-01-27
[Publisher] Gate create records route from non-admins

### LINKED ISSUE #1011: [Publisher] Gate `/reports/create` route from non-admins
There are cases where a user does not have a role assigned or has a READ ONLY role (or similar non-admin role) that does not allow a user to create a record. Currently, the Create Record button in the Records page is only visible to admin users, but it does not prevent access to the `/reports/create` route if a user goes to the URL directly or via a Create Record task card in the home page that links to the URL.

Two action items to accomplish this are:
* Gate the route from non-admin users (could add a simple redirect in the create users page)
* Add tooltip on the Home page Create Record task cards that reads: "You do not have permissions to create records, please reach out to your administrator for more assistance"

Inspired by this Sentry error: https://recidiviz-inc.sentry.io/issues/4559224133/?environment=production&project=4504532096516096&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

nasaownsky:
## Description of the change

1. Added tooltip to Task Cards for read only users

## Type of change

## Related issues

closes #1011 

### Development

### Code review

### DIFF
--- a/publisher/src/components/Home/TaskCard.tsx
+++ b/publisher/src/components/Home/TaskCard.tsx
@@ -28,7 +28,8 @@ import * as Styled from "./Home.styled";
 export const TaskCard: React.FC<{
   metadata: TaskCardMetadata;
   isSuperagency?: boolean;
-}> = observer(({ metadata, isSuperagency }) => {
+  isUserReadOnly?: boolean;
+}> = observer(({ metadata, isSuperagency, isUserReadOnly }) => {
   const { formStore } = useStore();
   const navigate = useNavigate();
   const {
@@ -54,8 +55,10 @@ export const TaskCard: React.FC<{
               action.label === taskCardLabelsActionLinks.uploadData.label
             )
               return;
+            const isReadOnlyCreateRecordTaskCard =
+              isUserReadOnly && action.path === "records/" && !recordID;
             const tooltipAnchorID =
-              action.path === "upload"
+              action.path === "upload" || isReadOnlyCreateRecordTaskCard
                 ? `${HomeStore.replaceSpacesAndParenthesesWithHyphen(
                     title
                   )}-tooltip-anchor`
@@ -113,6 +116,14 @@ export const TaskCard: React.FC<{
                     anchorId={tooltipAnchorID}
                     position="top"
                     content="You can also upload data for other metrics within the same file"
+                    centerText
+                  />
+                )}
+                {tooltipAnchorID && isReadOnlyCreateRecordTaskCard && (
+                  <Tooltip
+                    anchorId={tooltipAnchorID}
+                    position="top"
+                    content="You do not have permissions to create records, please reach out to your administrator for more assistance"
                     centerText
                   />
                 )}
--- a/publisher/src/components/Home/TaskCard.tsx
+++ b/publisher/src/components/Home/TaskCard.tsx
@@ -57,7 +57,7 @@ export const TaskCard: React.FC<{
               return;
             const isReadOnlyCreateRecordTaskCard =
               isUserReadOnly &&
-              action.path === taskCardLabelsActionLinks.createRecord.path &&
+              action.path === taskCardLabelsActionLinks.manualEntry.path &&
               !recordID;
             const tooltipAnchorID =
               action.path === taskCardLabelsActionLinks.uploadData.path ||
--- a/publisher/src/components/Home/TaskCard.tsx
+++ b/publisher/src/components/Home/TaskCard.tsx
@@ -56,11 +56,14 @@ export const TaskCard: React.FC<{
             )
               return;
             const isReadOnlyCreateRecordTaskCard =
-              isUserReadOnly && action.path === "records/" && !recordID;
+              isUserReadOnly &&
+              action.path === taskCardLabelsActionLinks.createRecord.path &&
+              !recordID;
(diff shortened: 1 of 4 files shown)


nasaownsky:
Hi @mxosman! Are we sure we don't gate it anywhere else? I'm trying to reach `/reports/create` and `/records/create` urls from read only user and can't do that.

nasaownsky:
Hi @mxosman! Are we sure we weren't gate it anywhere else before? I'm trying to reach `/reports/create` and `/records/create` urls under read only user and can't do that.

publisher/src/components/Home/TaskCard.tsx
```diff
@@ -54,8 +55,10 @@ export const TaskCard: React.FC<{
               action.label === taskCardLabelsActionLinks.uploadData.label
             )
               return;
+            const isReadOnlyCreateRecordTaskCard =
+              isUserReadOnly && action.path === "records/" && !recordID;
```
mxosman, line 59: nit: Can we create enums/consts for the path names incase we need to change them for any reason? I was thinking defining them in this file https://github.com/Recidiviz/justice-counts/blob/nasaownsky/1011-gate-create-records/publisher/src/components/Home/types.ts - updating the `taskCardLabelsActionLinks` links there and importing the enum here.

publisher/src/components/Home/TaskCard.tsx
```diff
@@ -54,8 +55,10 @@ export const TaskCard: React.FC<{
...
             const tooltipAnchorID =
-              action.path === "upload"
+              action.path === "upload" || isReadOnlyCreateRecordTaskCard
```
mxosman, line 61: Out of curiosity, why do we need the `|| isReadOnlyCreateRecordTaskCard` here for the anchor id?

mxosman (commented):
Thank you for working on this so quickly, @nasaownsky!

> Hi @mxosman! Are we sure we weren't gate it anywhere else before? I'm trying to reach `/reports/create` and `/records/create` urls under read only user and can't do that.

Oh you're absolutely right! We already implemented this - thanks for calling it out! I think what we didn't do was update the old ticket and the work you've done here (adding a tooltip to the buttons that take you there).

Sorry I didn't call this out before in the ticket - I just thought of it; but one thing that would be nice to implement is disabling those links on the cards and adjusting the color to be grey, just to avoid the redirecting all together.

One other thing worth addressing - it looks like there are two tooltips overlapping each other - the new one and the old one. It's hard to notice unless you look close or see the quick millisecond switch when you're hovering over them. Can you take another look at that?
Screenshot 2025-01-27 at 1 09 13 PM

publisher/src/components/Home/TaskCard.tsx
nasaownsky: So because of the condition `action.path === "upload"` we will create `tooltipAnchorID` only for upload path, but for the second tooltip we need it for `/records` path.

nasaownsky:
@mxosman I also noticed that for the read only users we have task cards with `Manual Entry` buttons that lead us to `agency/:agencyID/records/:recordID` pages, where they actually can't manual entry anything, but can see already existed existed data. Maybe rename `Manual Entry` in this cases?

publisher/src/components/Home/TaskCard.tsx
mxosman: Ah right - thank you for clarifying!

mxosman:
> @mxosman I also noticed that for the read only users we have task cards with `Manual Entry` buttons that lead us to `agency/:agencyID/records/:recordID` pages, where they actually can't manual entry anything, but can see already existed existed data. Maybe rename `Manual Entry` in this cases?

Good call - thank you for thinking of this! I thought about this too, and I think this is OK - as long as it leads them to the page, I'm fine with it for now continuing to operate as usual because they can still view the data. We'd also technically have this with the "Enter Data" link and the "Manual Entry" button when you go to the "Enter Data" page, but it's not worth adjusting all of those either. It's just the cases where they can't access the page that I think is worth addressing.

publisher/src/components/Home/TaskCard.tsx
```diff
@@ -54,10 +55,15 @@ export const TaskCard: React.FC<{
...
+              isUserReadOnly &&
+              action.path === taskCardLabelsActionLinks.manualEntry.path &&
+              !recordID;
             const tooltipAnchorID =
-              action.path === "upload"
+              action.path === taskCardLabelsActionLinks.uploadData.path ||
+              isReadOnlyCreateRecordTaskCard
                 ? `${HomeStore.replaceSpacesAndParenthesesWithHyphen(
-                    title
+                    `${title}-${action.label}`
```
mxosman, line 66: Nice!

mxosman (approved):
Awesome!! Thank you so much for this, Ilya! 🙏 Looks and feels great!

nasaownsky:
> Good call - thank you for thinking of this! I thought about this too, and I think this is OK - as long as it leads them to the page, I'm fine with it for now continuing to operate as usual because they can still view the data. We'd also technically have this with the "Enter Data" link and the "Manual Entry" button when you go to the "Enter Data" page, but it's not worth adjusting all of those either. It's just the cases where they can't access the page that I think is worth addressing.

@mxosman Got it, thank you for clarification and for other great advices!

PR #1657, opened 2025-01-30
[Publisher][Admin Panel] Add Vendor Management Modal

### LINKED ISSUE #1646: [Publisher][Admin Panel] Metrics Reporting Agency Feature: Vendor Management
In order to provide the list of vendors in the reporting agency dropdowns, CSG and admins should have a way of entering a list of vendors. This will be handled through a new flow in the Agency Provisioning page via a "Manage Vendors" button that will appear to the left of the "Create Agency" button.

When a user clicks on the "Manage Vendors" button, it should open up a new modal that allows them to do the following:
* Add a new vendor by entering a vendor name and URL
  * URLs should have basic URL validation
  * Add button is disabled until both fields are completed and URL is validated
* View a list of added vendors
* Hover over the list of vendors to see more controls (Edit and Remove)
* Remove a vendor by clicking the Remove button
* Edit a vendor by clicking the Edit button
  * Clicking this should populate the "Enter vendor name" and "Enter vendor URL" with the existing vendor name and URL
  * The "Add" button should now be "Update" and should ideally be disabled unless they make an edit
* Save changes once they've finished modifying the list of vendors
  * Save button should be disabled if no changes have been made or if there are any validation issues

Reference: [Designs](https://www.figma.com/design/ycrbtuWFuTlAuBraaVsLQx/Recidiviz-x-CSG-(Publisher--%3E-Dashboard)?node-id=1-84&t=xWrzkjDmezcitnSe-1)

nasaownsky:
## Description of the change

Adding Vendor Management Modal UI and logic

## Type of change

## Related issues

closes #1646 

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
@@ -39,6 +39,7 @@ import {
   UserProvisioning,
 } from ".";
 import * as Styled from "./AdminPanel.styles";
+import { VendorManagementModal } from "./VendorManagementModal";
 
 export const AgencyProvisioningOverview = observer(() => {
   const { adminPanelStore } = useStore();
@@ -62,6 +63,7 @@ export const AgencyProvisioningOverview = observer(() => {
   } = adminPanelStore;
 
   const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isVendorsModalOpen, setIsVendorsModalOpen] = useState(false);
   const [searchInput, setSearchInput] = useState<string>("");
   const [showAgenciesWithLiveDashboards, setShowAgenciesWithLiveDashboards] =
     useState(false);
@@ -175,6 +177,13 @@ export const AgencyProvisioningOverview = observer(() => {
   return (
     <>
       <ScrollToTop />
+
+      {isVendorsModalOpen && (
+        <VendorManagementModal
+          closeModal={() => setIsVendorsModalOpen(false)}
+        />
+      )}
+
       {isModalOpen && (
         <>
           <Modal>
@@ -285,8 +294,14 @@ export const AgencyProvisioningOverview = observer(() => {
           </Styled.InputLabelWrapper>
         </Styled.InputLabelContainer>
 
-        {/* Create Agency Button */}
         <Styled.ButtonWrapper>
+          {/* Manage Vendors Button */}
+          <Button
+            label="Manage Vendors"
+            onClick={() => setIsVendorsModalOpen(true)}
+            buttonColor="blue"
+          />
+          {/* Create Agency Button */}
           <Button
             label="Create Agency"
             onClick={openModal}
--- a/publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
@@ -34,6 +34,7 @@ import {
   AgencyKey,
   AgencyProvisioning,
   AgencyWithTeamByID,
+  Environment,
   Setting,
   SettingType,
   UserProvisioning,
@@ -42,7 +43,7 @@ import * as Styled from "./AdminPanel.styles";
 import { VendorManagementModal } from "./VendorManagementModal";
 
 export const AgencyProvisioningOverview = observer(() => {
-  const { adminPanelStore } = useStore();
+  const { adminPanelStore, api } = useStore();
   const {
     loading,
     agencies,
@@ -296,11 +297,13 @@ export const AgencyProvisioningOverview = observer(() => {
 
         <Styled.ButtonWrapper>
           {/* Manage Vendors Button */}
-          <Button
-            label="Manage Vendors"
-            onClick={() => setIsVendorsModalOpen(true)}
-            buttonColor="blue"
-          />
+          {api.environment !== Environment.PRODUCTION && (
+            <Button
+              label="Manage Vendors"
+              onClick={() => setIsVendorsModalOpen(true)}
+              buttonColor="blue"
+            />
+          )}
           {/* Create Agency Button */}
           <Button
             label="Create Agency"
--- a/publisher/src/components/AdminPanel/VendorManagementModal.tsx
+++ b/publisher/src/components/AdminPanel/VendorManagementModal.tsx
@@ -107,6 +107,7 @@ export const VendorManagementModal: React.FC<VendorManagementModalProps> =
       (nameValue !== editNameValue || urlValue !== editlUrlValue); // Changes detected in "Update" mode
 
     useEffect(() => {
+      // Handle return to "Add" mode when manually removing all values in "Update" mode
       if (!nameValue && !urlValue) {
         handleClearVendorInfo();
       }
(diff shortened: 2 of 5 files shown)


nasaownsky:
Hi @nichelle-hall @mxosman! A few question from the development:

1. The `admin/vendors` endpoint returns me HTML page with "You need to enable JavaScript to run this app." text in response. Is this expected because BE just isn't ready or something is wrong? Other endpoints (PUT and DELETE) also doesn't work with HTML "The method is not allowed for the requested URL" response.

2. In the Issue description there are requirements about Save button, and at first I was thinking about something like adding new or edited vendors to a new Set and the put in to BE with Save button, but as I can see from BE expected  [behavior](https://github.com/Recidiviz/justice-counts/issues/1646#issuecomment-2619773682) we can only add/edit one vendor at a time. Am I understanding correctly? In this case Add/Update button will be sending vendors straight off to BE, and I'm not exactly understand the purpose of the Save button.

3. Or we can just send something like this to PUT endpoint?

```
[
      { id: null, name: "Vendor A", url: "https://vendor-a.com" },
      { id: null, name: "Vendor B", url: "https://vendor-b.com" },
      { id: 1, name: "Vendor C", url: "https://vendor-c.com" },
      { id: 4, name: "Vendor D", url: "https://vendor-d.com" },
      { id: null, name: "Vendor E", url: "https://vendor-e.com" },
]
```
In this case Add/Update button will be adding/updating vendors in local Set and then sending it to BE via Save button.

nichelle-hall:
Hi @nasaownsky

1. The Endpoints:
I just merged all the backend endpoints as described in your Github Issue description. They will be available on staging tomorrow morning, around midday your time. In the meantime, you can point the FE to:
https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/

2. About the Save Button:
I may have interpreted the designs incorrectly. I initially understood that the "Add" button would immediately create the vendor through the backend, but I now see how this makes the "Save" button redundant. I’ll let @mxosman confirm the intended functionality and design. If the plan is to batch the additions/updates and only send them on "Save," I can update the endpoint to accept a list of vendors instead.

Let me know once you hear back!

mxosman:
Hi @nichelle-hall & @nasaownsky!

For question 2 - I was originally thinking the "Save" would send a batch update just for the sake of efficiency incase they need to add/edit/remove multiple vendors in one session. How cumbersome would it be to update the API to handle that @nichelle-hall? Sorry for not making this clear earlier. I also don't feel super strongly one-way or another, but figured it might be preferable to making a bunch of calls if users need to do a lot of vendor management.

nasaownsky:
The BE works as expected, thank you @nichelle-hall! 

@mxosman Hmm, if a user would create/add a bunch of vendors and then something goes wrong the progress can be lost, maybe it will be best to use current approach after all?  Anyway, please check my current implementation! 

1. Changed "Save" button to "Close"
2. Added "Clear" button 
3. Added borders for vendors to be better visually distinct

mxosman:
Hi @nasaownsky - I'm good with the current implementation! And, great call with the borders and buttons - it looks and feels great.

One small nit - if a user clicks Edit and then removes the vendor, can we clear the input? Just for the scenario where they edit an input they removed. If they're adding and click remove, we should keep their input. LMK if that's a huge pain.

https://github.com/user-attachments/assets/4f17ba3f-a273-496b-b548-08350960b102

Will do a code review later on this afternoon.

publisher/src/components/AdminPanel/AgencyProvisioningOverview.tsx
```diff
@@ -285,8 +294,14 @@ export const AgencyProvisioningOverview = observer(() => {
           </Styled.InputLabelWrapper>
         </Styled.InputLabelContainer>
 
-        {/* Create Agency Button */}
         <Styled.ButtonWrapper>
+          {/* Manage Vendors Button */}
+          <Button
```
mxosman, line 299: Can we gate this button to staging and dev environments for now?

publisher/src/components/AdminPanel/VendorManagementModal.tsx
```diff
@@ -0,0 +1,195 @@
...
+    const hasChanges =
+      hasValues &&
+      !urlValidationError &&
+      (nameValue !== editNameValue || urlValue !== editlUrlValue); // Changes detected in "Update" mode
+
+    useEffect(() => {
+      if (!nameValue && !urlValue) {
+        handleClearVendorInfo();
+      }
+    }, [nameValue, urlValue]);
```
mxosman, line 113: Can you help me understand what's going on with this effect? I'm not 100% following. Why are we clearing vendor info if it's already pretty much cleared? Or is it to clear the other states (editNameValue, and editURLValue, and UrlValidationError)?

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -631,7 +695,8 @@ class AdminPanelStore {
       | User
       | UserWithAgenciesByID
       | AgencyWithTeamByID
-      | AgencyTeamMember
+      // eslint-disable-next-line prettier/prettier
```
mxosman, line 698: Oi - any idea what's been going on with our `prettier` lately?

publisher/src/components/AdminPanel/VendorManagementModal.tsx
```diff
@@ -0,0 +1,195 @@
...
+              ))}
+            </Styled.VendorsScrollableInnerWrapper>
+            <Styled.VendorsButtonsContainer>
+              <Button label="Close" onClick={closeModal} buttonColor="blue" />
+            </Styled.VendorsButtonsContainer>
+          </Styled.VendorsContent>
+        </Styled.VendorsWrapper>
+      </Modal>
+    );
+  });
```
mxosman, line 195: So clean!

mxosman (commented):
Code looks great to me! Made one suggestion to gate it and another in my previous comment - otherwise, this is practically good to go!

publisher/src/components/AdminPanel/VendorManagementModal.tsx
nasaownsky: Thank you!

publisher/src/stores/AdminPanelStore.ts
nasaownsky: I believe there are conflicts between our `@recidiviz/eslint-config` and default prettier config files. I see that `@recidiviz/eslint-config` is not presented in `recidiviz-dashboards` repo and I was not observing prettier problems there in my time.

publisher/src/components/AdminPanel/VendorManagementModal.tsx
nasaownsky: If you comment out this effect, then go to Manage Vendors, hit edit and then manually delete Name and URL the modal will remain in "edit" state with the button showing "Update" text. I feel like this is kinda not right, and when you deleting these values you should go back to "add" mode, so this effect handles that.

mxosman: OH - I see. That makes a lot of sense. Thank you for helping me understand! Great call on that detail!

publisher/src/stores/AdminPanelStore.ts
mxosman: I'm totally good removing it from our repo - esp. if `recidiviz-dashbaords` no longer uses it. Sounds like it needs some updating and folks felt it wasn't worth it. Feel free to do it in this PR or a follow up one!

mxosman (approved):
Amazing, as always! Thank you so much, @nasaownsky!

publisher/src/stores/AdminPanelStore.ts
nasaownsky: @mxosman Hmm, I'm played with it a bit and it seems like if we remove this from our repo -- it will not have `eslint` configured at all. What do you think?

mxosman: Oh strange - I would assume it would fallback to the default eslint package, no? We probably have to update a few more configurations to delink from the @recidiviz/eslint-config - we can worry about it in a separate PR. I'll look into it or write up a ticket. Thanks for entertaining it!

nasaownsky: Got it, thank you!

PR #1665, opened 2025-02-06
[Publisher][Admin Panel] Add Metrics Reporting Agency tab

### LINKED ISSUE #1645: [Publisher][Admin Panel] Metrics Reporting Agency Feature
CSG would like a new feature in the Admin Panel where they can input the reporting agency for all metrics belonging to agencies. The goal is to help CSG track the source of data for all agency metrics. 

This new feature will be a new tab within the Agency Provisioning modal that will populate a list of the agency's metrics (for each sector) and provide a dropdown where they can select the reporting agency for each individual metric. 

The dropdown options will include (in this specific order):
* Current Agency
* CSG Justice Center
* [Child Agencies - only if the current agency is a superagency with child agencies]
* [Vendors]

The list of vendors will be provided by admins via a separate flow in a new "Manage Vendors" modal, where they can add/edit/remove vendors (see separate ticket: #1646).

Reference: [Designs](https://www.figma.com/design/ycrbtuWFuTlAuBraaVsLQx/Recidiviz-x-CSG-(Publisher--%3E-Dashboard)?node-id=1-84&t=xWrzkjDmezcitnSe-1)

nasaownsky:
## Description of the change

Added Metrics Reporting Agency feature UI and logic

## Type of change

## Related issues

closes #1645

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
+++ b/publisher/src/components/AdminPanel/AgencyProvisioning.tsx
@@ -49,6 +49,7 @@ import {
   InteractiveSearchList,
   InteractiveSearchListAction,
   InteractiveSearchListActions,
+  MetricsReportingAgency,
   ProvisioningProps,
   SaveConfirmation,
   SaveConfirmationType,
@@ -87,6 +88,8 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
       csgAndRecidivizUsers,
       csgAndRecidivizDefaultRole,
       teamMemberListLoading,
+      reportingAgencyMetadata,
+      reportingAgenciesUpdates,
       updateAgencyName,
       updateAgencyDescription,
       updateAgencyURL,
@@ -101,6 +104,7 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
       saveAgencyProvisioningUpdates,
       saveAgencyName,
       copySuperagencyMetricSettingsToChildAgencies,
+      saveReportingAgencies,
     } = adminPanelStore;
     const scrollableContainerRef = useRef<HTMLDivElement>(null);
 
@@ -175,6 +179,17 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
           setCurrentSettingType(AgencyProvisioningSettings.TEAM_MEMBERS_ROLES),
         selected:
           currentSettingType === AgencyProvisioningSettings.TEAM_MEMBERS_ROLES,
+      },
+      {
+        key: "metrics-reporting-agency",
+        label: AgencyProvisioningSettings.METRICS_REPORTING_AGENCY,
+        onClick: () =>
+          setCurrentSettingType(
+            AgencyProvisioningSettings.METRICS_REPORTING_AGENCY
+          ),
+        selected:
+          currentSettingType ===
+          AgencyProvisioningSettings.METRICS_REPORTING_AGENCY,
       },
     ];
 
@@ -337,6 +352,14 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
           userStore.email,
           Array.from(selectedMetricsKeys),
           Array.from(selectedChildAgencyIDsToCopy).map((id) => String(id))
+        );
+      }
+
+      /** Save Reporting Agencies Updates */
+      if (reportingAgenciesUpdates && selectedIDToEdit) {
+        saveReportingAgencies(
+          String(selectedIDToEdit),
+          reportingAgenciesUpdates
         );
       }
 
@@ -549,6 +572,10 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
     const hasChildAgenciesCopyUpdates = selectedChildAgencyIDsToCopy.size > 0;
     const hasMetricsCopyUpdates = selectedMetricsKeys.size > 0;
     /**
+     * An update has been made when there are new Metrics Reporting Agencies selected
+     */
+    const hasReportingAgenciesUpdates = reportingAgenciesUpdates.length > 0;
+    /**
      * Saving is disabled if saving is in progress OR an existing agency has made no updates to either the name, state,
      * county, systems, dashboard enabled checkbox, superagency checkbox and child agencies, child agency's superagency
      * selection, and team member additions/deletions/role updates, or a newly created agency has no input for both name and state.
@@ -566,7 +593,8 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
       hasIsSuperagencyUpdate ||
       hasChildAgencyUpdates ||
       hasSuperagencyUpdate ||
-      hasTeamMemberOrRoleUpdates;
+      hasTeamMemberOrRoleUpdates ||
+      hasReportingAgenciesUpdates;
     const hasRequiredCreateAgencyFields =
       hasNameUpdate && hasStateUpdate && hasSystems;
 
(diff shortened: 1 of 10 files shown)


nasaownsky:
"Hi @mxosman and @nichelle-hall! Aside from some small UI bugs that I'll address tomorrow, I think it turned out quite solid!

The only issue I've noticed with the API behavior is that it seems like I cannot save the "None" option to the BE. It acts like it's saved (with a 200 code), but it's not having any effect on the actual data. I'm sending `reporting_agency_id: null` with the updated `reporting_agencies` array when "None" is selected, so maybe this is the issue. Could you please check this out, @nichelle-hall?

Anyway, please check my changes and let me know your thoughts on them. Thank you!"

mxosman:
> "Hi @mxosman and @nichelle-hall! Aside from some small UI bugs that I'll address tomorrow, I think it turned out quite solid!
> 
> The only issue I've noticed with the API behavior is that it seems like I cannot save the "None" option to the BE. It acts like it's saved (with a 200 code), but it's not having any effect on the actual data. I'm sending `reporting_agency_id: null` with the updated `reporting_agencies` array when "None" is selected, so maybe this is the issue. Could you please check this out, @nichelle-hall?
> 
> Anyway, please check my changes and let me know your thoughts on them. Thank you!"
> 
> Edit: I also noticed that if we assign vendor as reporting agency -> save changes -> then delete this vendor from "Manage Vendors" modal -- the BE will preserve `is_self_reported` and `reporting_agency_id` values when we go back to "Metrics Reporting Agency" tab.

Thank you so much, @nasaownsky! I'll take a closer look at this on Monday.

Great call on the deleting the vendor - I didn't give this too much thought, but I'm thinking we don't need to preserve these if a vendor gets deleted, and the UI can just show None. What do you think, @nichelle-hall?

mxosman:
Thank you for your patience, @nasaownsky - I'll get to this tomorrow as soon as I can! @nichelle-hall will follow up with you as well!

nichelle-hall:
@nasaownsky 

Sorry I didn't get time to look at this today! I was heads-down in another task. I'll respond first thing tomorrow!

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
```diff
@@ -176,6 +180,19 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
...
+        key: "metrics-reporting-agency",
+        label: AgencyProvisioningSettings.METRICS_REPORTING_AGENCY,
+        onClick: () =>
+          setCurrentSettingType(
+            AgencyProvisioningSettings.METRICS_REPORTING_AGENCY
+          ),
+        selected:
+          currentSettingType ===
+          AgencyProvisioningSettings.METRICS_REPORTING_AGENCY,
+        /** Hide metrics reporting agency tab when creating a new agency */
```
mxosman, line 193: Great call! Thanks for thinking of this.

publisher/src/components/AdminPanel/MetricsReportingAgency.tsx
```diff
@@ -0,0 +1,148 @@
...
+              <Styled.ReportingAgencyMetricsContainer>
+                {metrics.map((metric) => {
+                  const initialReportingAgency =
+                    reportingAgencyMetadata.reporting_agencies[system].find(
+                      (agency) => agency.metric_key === metric.key
+                    );
+                  const updatedReportingAgency = reportingAgenciesUpdates.find(
+                    (agency) => agency.metric_key === metric.key
+                  );
+                  const selecteReportingdAgency =
```
mxosman, line 69: typo: `selectedReportingAgency`

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -369,45 +379,29 @@ class AdminPanelStore {
     }
   }
 
-  async addOrEditVendor(name: string, url: string, id?: number) {
+  async fetchReportingAgency(agencyID: string) {
     try {
+      this.reportingAgencyMetadataLoading = true;
```
mxosman, line 384: I think we need to wrap this in an action as well since this is an async method.

mxosman (commented):
Amazing! Works great (aside from the None)!

A couple of things:
1. Can we include ungating the Vendor Management so we can release all of this once we merge this PR in?
2. @nichelle-hall - when you get a chance, would you mind playing around with this and double checking that the saved selections propagate to the dashboard API correctly?

nichelle-hall:
@mxosman @nasaownsky 

This message is regarding not being able to save `null`. I'm still thinking about the deleting vendors issue and if there is a quick solution on the BE.

I think we should remove the `None` option. 

On the backend, we don’t currently handle saving "None" as a valid value for metric settings changes—particularly when an agency wants to null out a previously configured metric. 

From my perspective, "None" shouldn’t be an available option. Here’s why:

- It’s technically not valid to transition from a non-null reporting agency to a null value. There are really only two states: either the agency is reporting the metric itself, or it’s being reported by a vendor or super agency.
- "None" or null would indicate the metric is "unconfigured," but that doesn’t fit how we handle other metric configurations. (IE un-configuring a metric. It's either not available, reported monthly, or reported annually.). Once configured, you can switch between valid options, but you don’t "unconfigure" it.
- I’m also struggling to see a use case where an agency would intentionally select "None," given that the reporting agency options we provide are the only valid ones.

What do you think about this approach? I’d be happy to discuss if you see something I might be missing.

nichelle-hall:
I went ahead and handled the reporting agency once a vendor is deleted cleanup on the backend. I know you came up with a fix on the frontend, but it’s important to me that the database stays consistent, so I wanted to make sure this was handled properly at the BE level.

If you want to test it out, you can point the BE URL of your local application to  https://nichelletest---publisher-web-b47yvyxs3q-uc.a.run.app/.

Let me know if you run into any issues or if anything needs adjusting! 🚀😊

nasaownsky:
>I went ahead and handled the reporting agency once a vendor is deleted cleanup on the backend. I know you came up with a fix on the frontend, but it’s important to me that the database stays consistent, so I wanted to make sure this was handled properly at the BE level.

@nichelle-hall Thank you for your fix! Unfortunately, when performing action chain with vendor removal (on your test URL) I described earlier, the server now responds with 500 error when I'm trying to load the agency, in which the removed vendor was selected as reporting agency.

nasaownsky:
@mxosman @nichelle-hall 

Thank you both so much! Fixes have been applied, except for the vendor and 'None' case issues -- waiting on the decisions for that part.

publisher/src/stores/AdminPanelStore.ts
```diff
@@ -394,10 +392,14 @@ class AdminPanelStore {
       }
 
       runInAction(() => {
+        this.reportingAgencyMetadataLoading = true;
```
mxosman, line 395: Wait, this doesn't look right - I guess, I meant another `runInAction` up above the fetch because we want to set the loading _before_ we make the request, right?

mxosman (commented):
Hi @nasaownsky - so based on the discussion @nichelle-hall had with CSG, they are comfortable removing the "None" option (thank you so much for chatting with them, Nichelle!).

I'll let @nichelle-hall follow up with you on the server errors you're running into. We're almost there!

publisher/src/stores/AdminPanelStore.ts
nasaownsky: You're right! I guess I was fooled into thinking it worked in this case because there was another loading property that showed mini-loader. 

It actually works with or without wrapping it into `runInAction` before the request -- but I wrapped it anyway. Thank you for noticing!

nasaownsky:
@nichelle-hall Thank you for your fixes, everything works flawlessly!
@mxosman I think it should be ready to go!

publisher/src/stores/AdminPanelStore.ts
mxosman: Ah - amazing! Thank you, @nasaownsky !

mxosman (approved):
Amazing! Thank you both so much, @nasaownsky and @nichelle-hall!! This is good to go! Yay!

PR #1672, opened 2025-02-14
[Publisher][Bug] Fix breakdowns data overriding

### LINKED ISSUE #1663: [Publisher] Incorrect Arrest Data Display for Kimberling City PD
**Description of the Bug**

An issue has been identified on Publisher's "Explore Data" page where arrest data for Kimberling City PD (MO) is incorrectly displayed for January 2021. Specifically, when the user selects the "Offense Type" breakdown, the data for January 2021 is missing. This discrepancy affects both the table and chart views, leading to inconsistencies in the displayed data.

**Expected Behavior**

The table view for January 2021 should display an overall total of 2 arrests.

When selecting the "Offense Type" breakdown under the chart view, the data for January 2021 should be correctly shown and reflect the breakdown by offense type.

**Actual Behavior**

Table View: The data table currently shows an overall total of 0 arrests for January 2021, instead of the expected value of 2.

Chart View: When the "Offense Type" breakdown is selected, no data for January 2021 is displayed.

Testing Dashboard: The correct data (2 arrests for January 2021) is reflected correctly on the testing dashboard, indicating the issue may be specific to how the data is displayed on the "Explore Data" page

nasaownsky:
## Description of the change

Fixed breakdowns data overriding

## Type of change

## Related issues

closes #1663 

### Development

### Code review

### DIFF
--- a/common/stores/BaseDatapointsStore.ts
+++ b/common/stores/BaseDatapointsStore.ts
@@ -88,7 +88,7 @@ abstract class DatapointsStore {
           frequency: dp.frequency,
           dataVizMissingData: 0,
         });
-      } else if (dp.value !== null) {
+      } else if (dp.value !== null || dp.value !== undefined) {
         if (
           !res[dp.metric_definition_key].disaggregations[
             dp.disaggregation_display_name
--- a/common/stores/BaseDatapointsStore.ts
+++ b/common/stores/BaseDatapointsStore.ts
@@ -88,9 +88,7 @@ abstract class DatapointsStore {
           frequency: dp.frequency,
           dataVizMissingData: 0,
         });
-        // the res object could be overwritten based on the datapoint order since the logic is grouping by start_date
-        // and this check essentially skips all null datapoints and writes correct data
-      } else if (dp.value || dp.value === 0) {
+      } else if (dp.value !== null) {
         if (
           !res[dp.metric_definition_key].disaggregations[
             dp.disaggregation_display_name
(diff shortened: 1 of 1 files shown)


nasaownsky:
This was brutal, but I find the solution.

It was again due to the data overriding. [Previously](https://github.com/Recidiviz/justice-counts/pull/1498#issuecomment-2345859265) we already fixed this issue, but then [after](https://github.com/Recidiviz/justice-counts/pull/1488#pullrequestreview-2406458281) we also included `0` entries to be shown on the Explore Data page.

In this case we were having annual Arrests record which was `0`, so it leaked into monthly Arrests record (they were sharing the same `start_date`) and overridden it.

mxosman:
Will review this next week as well - thank you for your brain and for figuring this out! These issues are always so brutal.

common/stores/BaseDatapointsStore.ts
```diff
@@ -88,9 +88,7 @@ abstract class DatapointsStore {
           frequency: dp.frequency,
           dataVizMissingData: 0,
         });
-        // the res object could be overwritten based on the datapoint order since the logic is grouping by start_date
-        // and this check essentially skips all null datapoints and writes correct data
-      } else if (dp.value || dp.value === 0) {
+      } else if (dp.value !== null) {
```
mxosman, line 91: We should probably catch `undefined` as well.

common/stores/BaseDatapointsStore.ts
```diff
@@ -101,41 +99,41 @@ abstract class DatapointsStore {
           ] = {};
         }
 
+        /**
+         The res object can be overwritten depending on the order of datapoints,
+         when grouping by a single key (e.g., start_date), since start_date
+         can be the same for different frequencies, so we must include frequency 
+         as part of the grouped key to prevent data from being overridden.
+        */
+        const groupedKey = `${dp.start_date}_${dp.frequency}`;
```
mxosman, line 108: Nice! Thanks for the comment - this makes sense.

common/stores/BaseDatapointsStore.ts
```diff
@@ -101,41 +99,41 @@ abstract class DatapointsStore {
...
+        const existingEntry =
           res[dp.metric_definition_key].disaggregations[
             dp.disaggregation_display_name
-          ][dp.start_date] &&
-          Object.keys(
-            res[dp.metric_definition_key].disaggregations[
-              dp.disaggregation_display_name
-            ][dp.start_date]
-          ).includes(dimensionName);
+          ][groupedKey];
```
mxosman, line 119: I don't see where you are grouping these already by key - how are you able to index into this entry by `groupKey` if `res[dp.metric_definition_key].disaggregations[dp.disaggregation_display_name]` isn't grouped that way?

common/stores/BaseDatapointsStore.ts
```diff
@@ -101,41 +99,41 @@ abstract class DatapointsStore {
...
         }
 
         res[dp.metric_definition_key].disaggregations[
           dp.disaggregation_display_name
-        ][dp.start_date] = {
-          ...res[dp.metric_definition_key].disaggregations[
-            dp.disaggregation_display_name
-          ][dp.start_date],
+        ][groupedKey] = {
+          ...existingEntry,
```
mxosman, line 136: I guess this is where it's happening, but I'm not 100% following - can you bring me up to speed?

mxosman:
I just played with it and it looks like it's working out great! Just left some questions for my understanding.

>It was again due to the data overriding. https://github.com/Recidiviz/justice-counts/pull/1498#issuecomment-2345859265 we already fixed this issue, but then https://github.com/Recidiviz/justice-counts/pull/1488#pullrequestreview-2406458281 we also included 0 entries to be shown on the Explore Data page.

Oh boy - we're finding every edge case with this one! Thanks again for your help with this!

common/stores/BaseDatapointsStore.ts
nasaownsky: Yes, grouping happens here, basically you can group by any string -- previously we was grouping by `dp.start_date`, but because of possible overlapping we just need to make grouping key more complex. You will see it if you log this inside our else block.

```
console.log(res[dp.metric_definition_key].disaggregations[dp.disaggregation_display_name]);

mxosman: That part makes sense, I guess I'm not following why it's happening at the very end and not before we query by this grouping? It feels like the first datapoint will not have a `res[dp.metric_definition_key].disaggregations[dp.disaggregation_display_name][groupedKey]`, but I think I'm misunderstanding something...

mxosman (approved):
Thank you again for working on the fixes and for debugging this one! Approving because it works, but still confused per my last comment - but it shouldn't be blocking, mostly me just misunderstanding!

common/stores/BaseDatapointsStore.ts
nasaownsky: Actually we need `existingEntry` to calculate `dimensionValue`, and `existingEntry` is only defined if the grouped object already exists, which means that the grouped object will be ready before `dimensionValue` calculated. Also since js objects store references, any updates to `existingEntry` are immediately reflected in `res`, preventing accidental overwrites while preserving accumulated values. Hope this clears a bit! @mxosman

nasaownsky:
Thank you @mxosman! I left a comment with additional explanation, hope it helps a bit!

common/stores/BaseDatapointsStore.ts
mxosman: Ah - OK, I'm following now. Wow, that made my head spin for a sec - but following through the entire file and your explanation helped me latch on to what was going on here! Thanks again for your clarifications! Excellent work sorting all this out and plugging in a great solution!

PR #1683, opened 2025-02-28
[Publisher][Admin Panel] Refactor Agency Provisioning

### LINKED ISSUE #1624: [Refactor] Admin Panel: Agency Provisioning
The Agency Provisioning component has grown significantly in size and complexity, making it increasingly difficult to manage and extend as we plan for upcoming features. To ensure the system remains flexible and maintainable, we propose breaking down the component into smaller, more modular pieces. This refactor will make it easier to add new functionality, reduce technical debt, and improve overall code quality as we continue to scale.

This PR focuses on the following key goals:

**Modularization**: Begin the process of splitting the Agency Provisioning component into smaller, more manageable modules. See starting point in screenshot below.

**Test Coverage**: Before breaking down the component, we should increase our coverage by writing tests to ensure the existing functionality remains intact and behaves as expected and continues to behave as expected when we refactor.

Image

This is just a starting point suggestion - please feel free to come up with your own abstraction. We can begin by breaking the tabs into separate components (purple highlight), and each orange highlighted section can be broken up into its own individual components - one for the general agency information that is simple and unlikely to change, one for the superagency/child agency selection, and one for the copy metric settings feature.

nasaownsky:
## Description of the change

Refactored Agency Provisioning to smaller components with the use of Contexts

## Type of change

## Related issues

closes #1624 

### Development

### Code review


nasaownsky:
Now when we finally done with design updates we can revisit this one! So I was thinking about approach like this, what do you think about it @mxosman?

mxosman:
This is a great start, @nasaownsky! Super helpful to see your approach. I want to see if we can break things down even further and based on our conversation a few months ago, consider using `contexts` to help us with all of this:

* Maybe let's create an `AgencyProvisioningContext` and `AgencyProvisioningProvider` to house a bunch of the intermediary state logic (`selectedSystems`, `nameValue`, `URLValue`, etc), and make a `useAgencyProvisioning` to use the context.
* Breakdown the `AgencyGeneralInformation`'s individual components:
  * `AgencyGeneralInformation` will now take in the `selectedAgency` and use the `AgencyProvisioningProvider`
  * `AgencyGeneralInformationContent` as the parent that'll be inside of the provider, and use the values in `useAgencyProvisioning`, and include things like the URL validation logic
  * `AgencyNameInput`
  * `AgencyStateInput`
  * `AgencyCountyInput`
  * `AgencySystemsInput`
  * `AgencyDescriptionInput`
  * `AgencyURLInput`
  * `DashboardEnabledCheckbox`

Let me know your thoughts - I think the above would help clean things up even more.

publisher/src/components/AdminPanel/AgencyProvisioningComponents/AgencyGeneralInformation.tsx
```diff
@@ -0,0 +1,341 @@
...
+                  SupervisionSubsystems.forEach((subsystem) =>
+                    currentSystems.delete(subsystem)
+                  );
+                }
+
+                return toggleAddRemoveSetItem(
+                  currentSystems,
+                  id as AgencySystems
+                );
+              });
```
mxosman, line 258: Would love if possible for large inline logic like this to be extracted outside of the component body or better yet, be made into a util if possible.

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+  setShowSelectionBox: React.Dispatch<
+    React.SetStateAction<SelectionInputBoxType | undefined>
+  >;
+  nameValue: string;
+  setNameValue: React.Dispatch<React.SetStateAction<string>>;
+  URLValue: string;
+  setURLValue: React.Dispatch<React.SetStateAction<string>>;
+  descriptionValue: string;
+  setDescriptionValue: React.Dispatch<React.SetStateAction<string>>;
+}
```
nasaownsky, line 38: This part will be expanding with further refactor, but I find it a little bit clunky. Maybe you aware of some better way we could group this?

mxosman:
Thank you for your patience - I promise to review as soon as I get a chance!

publisher/src/components/AdminPanel/AgencyProvisioningComponents/index.ts
```diff
@@ -0,0 +1,25 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+export { default as AgencyCountyInput } from "./AgencyCountyInput";
```
mxosman, line 18: Any reason to lean towards exporting `default` vs just exporting the component?

publisher/src/components/AdminPanel/AgencyProvisioningComponents/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+    },
+    {
+      label: "Close",
+      onClick: () => closeCallback(),
+    },
+  ];
+};
+
+export const updateSystemsSelections = (
+  id: string | number,
```
mxosman, line 60: Since the id here will always be systems, could we type it as `AgencySystems`?

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
mxosman: Do we need to pass around the SetState functions via context? I'm thinking those can live in their respective components, and just update the store directly. Or are we doing this because we're checking to see if there are any updates?

I do agree, this does feel clunky. LMK your thoughts on the questions above, and we can rethink this. If we need these for the purposes of checking updates, we should think of a different solution to check whether or not a user made updates - but I'm thinking each component can either internalize their state or update the store directly - and we should only pass items shared between these individual components via context. But, let me know if I'm not thinking this all the way through as you have.

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const [nameValue, setNameValue] = useState<string>(
+    selectedAgency?.name ?? ""
+  );
+  const [URLValue, setURLValue] = useState<string>(
+    selectedAgency?.agency_url ?? ""
+  );
+  const [descriptionValue, setDescriptionValue] = useState<string>(
+    selectedAgency?.agency_description ?? ""
+  );
```
mxosman, line 78: Same comment as above re: these being internalized in their own components.

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const providerValue = useMemo(
```
mxosman, line 80: Ah, interesting that we're memoizing this - any reason?

publisher/src/components/AdminPanel/AgencyProvisioningComponents/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+      onClick: (filteredList: SearchableListItem[] | undefined) => {
+        setState(selectAllSet);
+        if (filteredList) {
+          const filteredSet = new Set(
+            filteredList.map((obj) => obj.id)
+          ) as Set<T>;
+          setState(filteredSet);
+        }
+        if (selectAllCallback) selectAllCallback();
+      },
```
mxosman, line 46: Small refactor suggestion to avoid two calls to `setState` (worth testing):

```
onClick: (filteredList: SearchableListItem[] | undefined) => {
  const newSelection = filteredList
    ? new Set(filteredList.map((obj) => obj.id)) as Set<T>
    : selectAllSet;

  setState(newSelection);

  if (selectAllCallback) {
    selectAllCallback();
  }
},
```

publisher/src/components/AdminPanel/AgencyProvisioningComponents/AgencyURLInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { useStore } from "../../../stores";
+import * as Styled from "../AdminPanel.styles";
+import { useAgencyProvisioning } from "../AgencyProvisioningContext";
+
+const AgencyURLInput: React.FC = observer(() => {
+  const { adminPanelStore } = useStore();
+  const { updateAgencyURL } = adminPanelStore;
+
+  const { URLValue, setURLValue } = useAgencyProvisioning();
```
mxosman, line 30: Sorry, I might've mislead you here... I'm now wondering whether or not we need to pass things like this around, or if it can just live within this component?

publisher/src/components/AdminPanel/AgencyProvisioningComponents/AgencySystemsInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+  } = useAgencyProvisioning();
+
+  return (
+    <>
+      {showSelectionBox === SelectionInputBoxTypes.SYSTEMS && (
+        <InteractiveSearchList
+          list={searchableSystems}
+          boxActionType={InteractiveSearchListActions.ADD}
+          selections={selectedSystems}
+          buttons={getInteractiveSearchListSelectDeselectCloseButtons(
```
mxosman, line 55: Let's break this apart outside of the component and pass in the final result. Even breaking apart the arguments here would be more readable.

publisher/src/components/AdminPanel/AgencyProvisioningComponents/AgencyStateInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <InteractiveSearchList
+          list={AdminPanelStore.searchableStates}
+          boxActionType={InteractiveSearchListActions.ADD}
+          selections={
+            agencyProvisioningUpdates.state_code
+              ? new Set([agencyProvisioningUpdates.state_code])
+              : new Set()
+          }
+          buttons={interactiveSearchListCloseButton}
+          updateSelections={({ id }) => {
```
mxosman, line 60: nit: let's move this logic outside of the prop and put it in a callback

publisher/src/components/AdminPanel/AgencyProvisioningComponents/AgencyCountyInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+            agencyProvisioningUpdates.fips_county_code
+              ? new Set([agencyProvisioningUpdates.fips_county_code])
+              : new Set()
+          }
+          buttons={interactiveSearchListCloseButton}
+          updateSelections={({ id }) => {
+            updateCountyCode(
+              agencyProvisioningUpdates.fips_county_code === id
+                ? null
+                : (id as FipsCountyCodeKey)
```
mxosman, line 62: nit: move this outside and pass in the final result here

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
```diff
@@ -464,17 +433,20 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
      * New agency: an update has been made when the agency has a value for `agencyProvisioningUpdates.agency_description`
      */
     const hasDescriptionUpdate = selectedAgency
-      ? descriptionValue !== (selectedAgency.agency_description || "")
-      : Boolean(agencyProvisioningUpdates.agency_description);
+      ? agencyProvisioningUpdates.agency_description !== null &&
```
mxosman, line 436: I reviewed from the bottom up - sorry if this feels out of order. I'm wondering if we should rethink how we're handling detecting updates because this part is also one of the clunkiest parts... Might be worth handling in the store - what are your thoughts?

mxosman (commented):
@nasaownsky - my sincerest apologies for the long delay in reviewing this. I like how this is coming along - very much appreciate you working through this as it's super gnarly. Let me know if any of my comments make sense/I can clarify further. Sorry if it feels thrashy, but we're honing in on a good clean up I think.

publisher/src/components/AdminPanel/AgencyProvisioningComponents/index.ts
nasaownsky: Yeah, I think just components will do better.

publisher/src/components/AdminPanel/AgencyProvisioningComponents/utils.ts
nasaownsky: I think no, because `id` comes from `updateSelections` as `string | number`, and it's used in several other places, so I think this was done on purpose for the sake of versatility.

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
nasaownsky: We're using `useMemo` here to avoid unnecessary re-renders of components that will consume `providerValue`.

nasaownsky:
@mxosman Thank you for your review! So I answered/fixed some of your questions, but returning to the contexts: I hooked up all Agency Information tab (since I worked on it anyway) for us to have bigger picture. Yes, I totally agree that in places where we can internalize states this would be ideal, but in some cases we need them for cross-use between multiple components and/or Agency Provisioning (but they're not really suitable for store, because Agency Provisioning would be the only place we will use them, so that's kinda the reason we choosed the contexts way in the first place). I think with the excluded internalized states and some type improvements I made it actually works flawlessly for our needs here. What do you think?

I'll look into the rest of your review tomorrow, but you're welcome to add more comments on the new code! And also I have one small question below. Thank you so much!

publisher/src/components/AdminPanel/AgencyProvisioningComponents/ChildAgenciesList.tsx
```diff
@@ -0,0 +1,108 @@
...
+              new Set(agencyIDs.filter((id) => id !== selectedAgency?.id)),
+              () => setShowSelectionBox(undefined)
+            )}
+            updateSelections={({ id }) => {
+              setSelectedChildAgencyIDs((prev) =>
+                toggleAddRemoveSetItem(prev, +id)
+              );
+              // Question: when selecting new child agency from superagencies list should we also update selected child agencies to copy metrics into? I'd say we should.
+              setSelectedChildAgencyIDsToCopy((prev) =>
+                toggleAddRemoveSetItem(prev, +id)
```
nasaownsky, line 72: The question! For example, when the copy metrics section is enabled, when we selecting new child agencies for superagency they will not be included in the selected child agencies in copy metrics section. I think maybe we should include them?

mxosman: Oh this is a great question - I think we should too - but, let me check in with @nichelle-hall. Were we including them beforehand?

@nichelle-hall - re: @nasaownsky's question - do you know if the copy metric settings to child agencies script needs the child agencies to be connected to the agency beforehand? Or (in the above scenario), can we add child agencies and trigger the copy in the same request?

mxosman:
@mxosman Thank you for your review! So I answered/fixed some of your questions, but returning to the contexts: I hooked up all Agency Information tab (since I worked on it anyway) for us to have bigger picture. Yes, I totally agree that in places where we can internalize states this would be ideal, but in some cases we need them for cross-use between multiple components and/or Agency Provisioning (but they're not really suitable for store, because Agency Provisioning would be the only place we will use them, so that's kinda the reason we choosed the contexts way in the first place). I think with the excluded internalized states and some type improvements I made it actually works flawlessly for our needs here. What do you think?

For sure! I think for the state that's shared - we can pass them around the context, but what gave me pause were things that aren't shared being passed through the context (e.g. updating the agency URL, description, etc.) - I think it makes more sense to keep those components state internalized and only share things that are used by other components. That'll reduce some of the bloating of the `AgencyProvisioningContextProps` that we were seeing before. I see that you cleaned that up though - so no worries there. 

Thank you for stitching everything together - I'll keep giving this my full attention now that I have a little more room to breathe w/ other focuses. Hoping to help wrap this up soon! It's much cleaner than the towering monolith it was before! I appreciate you so much. 🙏

publisher/src/components/AdminPanel/AgencyProvisioningComponents/ChildAgenciesList.tsx
nichelle-hall: The script requires that the child agencies are already connected to the agency because it queries for them in the DB directly.

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
```diff
@@ -745,633 +636,66 @@ export const AgencyProvisioning: React.FC<ProvisioningProps> = observer(
...
-                                agenciesByID[
-                                  agencyProvisioningUpdates.super_agency_id
-                                ]?.[0].name}
-                            </Styled.Chip>
-                          )}
-                        </Styled.ChipContainer>
-                        <Styled.ChipContainerLabel>
-                          Superagency
-                        </Styled.ChipContainerLabel>
-                      </Styled.InputLabelWrapper>
```
mxosman, line 1373: Super satisfying to see all of the logic removed in this file!

publisher/src/components/AdminPanel/AgencyProvisioningComponents/ChildAgenciesList.tsx
nasaownsky: Got it, thank you @mxosman and @nichelle-hall!

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
nasaownsky: Yes, everything comes together!

publisher/src/components/AdminPanel/AgencyProvisioningComponents/utils.ts
nasaownsky: Thank you for suggestion, working as expected!

publisher/src/components/AdminPanel/AgencyProvisioning.tsx
nasaownsky: Hmm, I think maybe let's finish with breaking into small components and then think some more about internal `AgencyProvisioning` logic. Maybe it is could be optimized, but we need to have all small components to be settled and dandy.

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
```diff
@@ -39,6 +44,14 @@ interface AgencyProvisioningContextProps {
...
   selectedMetricsKeys: Set<string>;
   setSelectedMetricsKeys: SetState<Set<string>>;
+  addOrDeleteUserAction?: InteractiveSearchListAction;
+  setAddOrDeleteUserAction: SetState<InteractiveSearchListAction | undefined>;
+  selectedTeamMembersToAdd: Set<number>;
+  setSelectedTeamMembersToAdd: SetState<Set<number>>;
+  selectedTeamMembersToDelete: Set<number>;
+  setSelectedTeamMembersToDelete: SetState<Set<number>>;
+  teamMemberRoleUpdates: UserRoleUpdates | Record<number, never>;
+  setTeamMemberRoleUpdates: SetState<UserRoleUpdates | Record<number, never>>;
```
mxosman, line 54: Yeah, this is now a mammoth... One option would be to break up the modularize each context more, and have a special context + provider for each set of things (agency info, super/child agency selection, copy metrics, team members, etc.). A lower lift option that I'm also open to would be to extract all of this into the `types.ts` and creating a different type for each state group? No biggie - just a thought (feel free to ignore). At the very least we can move it to `types.ts`

E.g.

```
export interface SuperChildAgencyInfo { // or a better name/suffix
  isChildAgencySelected: boolean;
  setIsChildAgencySelected: SetState<boolean>;
  selectedChildAgencyIDs: Set<number>;
  setSelectedChildAgencyIDs: SetState<Set<number>>;
}

export interface CopyMetricSettingsInfo {
  selectedChildAgencyIDsToCopy: Set<number>;
  setSelectedChildAgencyIDsToCopy: SetState<Set<number>>;
  isCopySuperagencyMetricSettingsSelected: boolean;
  setIsCopySuperagencyMetricSettingsSelected: SetState<boolean>;
  selectedMetricsKeys: Set<string>;
  setSelectedMetricsKeys: SetState<Set<string>>;
}

export interface TeamMemberInfo {
  selectedTeamMembersToAdd: Set<number>;
  setSelectedTeamMembersToAdd: SetState<Set<number>>;
  selectedTeamMembersToDelete: Set<number>;
  setSelectedTeamMembersToDelete: SetState<Set<number>>;
  teamMemberRoleUpdates: UserRoleUpdates | Record<number, never>;
  setTeamMemberRoleUpdates: SetState<UserRoleUpdates | Record<number, never>>;
}
```

Then in the `AgencyProvisioningContextProps`:

```
export interface AgencyProvisioningContextProps {
  ...
  // Something like...
  superChildAgencyInfo: SuperChildAgencyInfo; // will have to destructure/de-nest
  copyMetricSettings: CopyMetricSettingsInfo; // will have to destructure/de-nest
  teamMembers: TeamMemberInfo; // will have to destructure/de-nest
}
```

mxosman (commented):
This is looking fantastic! I'm sure there are ways we can improve this even more via refactoring the stores and introducing new ones, but the goal was to dramatically improve the previous abstraction, which I think you accomplished beautifully given the challenges of the previous abstraction! I appreciate all your time, efforts and ideas so much. I love this version so much better than before!

I left some super minor suggestions - Nichelle and I will do some manual testing this week and then I think this should be good to go. LMK once you feel like this is ready for me to deploy to a playtesting branch and we'll proceed with a manual test.

helperbot-recidiviz:

publisher/src/components/AdminPanel/AgencyProvisioningContext.tsx
nasaownsky: Sure, elevated your second suggestion -- please check my changes.

nasaownsky:
> LMK once you feel like this is ready for me to deploy to a playtesting branch and we'll proceed with a manual test.

I think it should be ready for testing. I also thoughtfully tested this on every step, comparing existed behavior from `staging` with refactored changes. So yeah, please do test it once more and let me know if anything work not as expected.

helperbot-recidiviz:

publisher/src/components/AdminPanel/types.ts
```diff
@@ -274,6 +274,39 @@ export type InteractiveSearchListProps = {
   isActiveBox?: boolean;
 };
 
+/** Agency Provisioning Context Types */
+
+export type SetState<T> = React.Dispatch<React.SetStateAction<T>>;
+
+export interface SuperagencyChildAgencyContext {
+  selectedSystems: Set<AgencySystem>;
+  setSelectedSystems: SetState<Set<AgencySystem>>;
```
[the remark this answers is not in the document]
nasaownsky: They're used both in general info and in superageny/child agency context, but you're right, they probably should be in general.

nichelle-hall:
@nasaownsky Here’s what I observed during manual testing:

- The “Is Stepping Up Agency” checkbox is no longer visible.

- When creating a new agency from the Create User modal, the new agency is created successfully — but all previously existing agencies are deleted.

- If I create a new agency from the Create User modal (same issue as above), then save the user and reopen the modal, the agency that was just created is automatically added to the new user's agency list.

Let me know if you'd like me to test a specific flow or retest after a fix!

mxosman:
Great catches, Nichelle! I just finished up testing as well and have similar findings:

* +1 to Nichelle, I think we forgot to add in the Stepping Up checkbox that was worked on separately.
* Related to the first bullet point - we're also missing the Reporting Agency Metric feature & tab as well.

Those might be the only two that were worked on in parallel to this PR - would you mind doing a quick check to see if there are other Admin Panel related work that we might be missing as well?

* I too am noticing the same weird behavior when creating an agency via the secondary flow from the User Provisioning screen. It clears out all of the agencies, and when you refresh and open up the user again, now they have two instances of the newly created agency. Oi - it looks like this is happening in staging - I'll create a new ticket for it since it's high priority, but we should take this down until we fix it.

* There's a separate bug that's unrelated to this and might be BE during the copying of metric settings which I'll create a separate ticket for. 

Other than the above, I wasn't able to find anything else.

> If I create a new agency from the Create User modal (same issue as above), then save the user and reopen the modal, the agency that was just created is automatically added to the new user's agency list.

@nichelle-hall - I think this is the expected behavior. If you're creating a new agency from that secondary modal that pops up, the code assumes that you're wanting to add that agency to the current user. LMK if I'm misunderstanding you though!

nasaownsky:
@mxosman @nichelle-hall I'm seeing Stepping Up Agency checkbox & Metrics Reporting Agency tab on localhost. Maybe playtest builded with some different env?

mxosman:
> @mxosman @nichelle-hall I'm seeing Stepping Up Agency checkbox & Metrics Reporting Agency tab on localhost. Maybe playtest builded with some different env?

Oh strange... It looks like the cloud build for the deploy to playtesting actually failed silently. Thanks for calling this out. Hmm... I'll debug and follow up soon.

mxosman:
OK - no clue what happened... I just redeployed again and it built just fine. This deploy might've resolved a caching issue as it was blocking on the node version being too high - sorry about the false alarms, @nasaownsky!

@nichelle-hall - this means the version we tested was not this one... sorry for the back and forth, would you mind retesting again?

@nasaownsky - in the meantime, would you mind taking a look at #1728?

mxosman:
@nasaownsky - a couple of things:

1. One thing I'm noticing is when you go to other tabs and go back to the Agency Information, the inputs are wiped out. Would you mind taking a look?

It's more visible when trying to create a new agency. Try that, enter a name and description and URL, and go to the Team Members tab and then back.

2. When creating a new agency, and you fill out the required fields, the error state on the URL no longer blocks saving. It allows you to save with a bad url that then gets stripped out when you reopen the modal.

nasaownsky:
>One thing I'm noticing is when you go to other tabs and go back to the Agency Information, the inputs are wiped out. Would you mind taking a look?

@mxosman this is the downside of separating things into smaller components. Moving states to context resolves the problem though. So I  think the solution here is to move user edited text field's states to the context. What do you think?

mxosman:
> > One thing I'm noticing is when you go to other tabs and go back to the Agency Information, the inputs are wiped out. Would you mind taking a look?
> 
> @mxosman this is the downside of separating things into smaller components. Moving states to context resolves the problem though. So I think the solution here is to move user edited text field's states to the context. What do you think?

Yeah, that makes sense. I think anything that's shared/observed by other components should live in the context and anything that isn't should continue being encapsulated. In this case - maybe just the state and not the setter?

nasaownsky:
>In this case - maybe just the state and not the setter?

I don't think this is possible, because in that case (without setter in the context) the state value just won't be updated. So should I include agency name, url and description values/setters to context for them not to be wiped out when switching tabs?

mxosman:
> > In this case - maybe just the state and not the setter?
> 
> I don't think this is possible, because in that case (without setter in the context) the state value just won't be updated. So should I include agency name, url and description values/setters to context for them not to be wiped out when switching tabs?

Ah, right - yes, sorry - not thinking about this all the way through - as all the state is up on the provider and the components themselves need the setter from there - my bad! Yes, please.

mxosman (approved):
Thank you for all of the sweat and effort you put into cleaning this all up, @nasaownsky! I appreciate your help more than you know and very much appreciate your patience throughout the review cycle, and with all of the changes requested. This was a hell of an undertaking! It feels so much better than before, all thanks to you! 🙏 🙏 🙏

nasaownsky:
Thank you so much for all your help with figuring this out and final testing @mxosman and @nichelle-hall!

PR #1703, opened 2025-03-19
[Publisher][Admin Panel] Add more unit tests for Admin Panel

### LINKED ISSUE #1530: [Publisher][Admin Panel] Add unit test for adding/removing team members

nasaownsky:
## Description of the change

Added more unit tests for Admin Panel

## Type of change

## Related issues

closes #1530 

### Development

### Code review

### DIFF
--- a/publisher/src/components/AdminPanel/AdminPanel.test.tsx
+++ b/publisher/src/components/AdminPanel/AdminPanel.test.tsx
@@ -25,12 +25,16 @@ import { BrowserRouter } from "react-router-dom";
 import { rootStore, StoreProvider } from "../../stores";
 import {
   mockAgenciesResponse,
+  mockReportingAgencyMetadata,
   mockUsersResponse,
+  mockVendorsResponse,
 } from "../../stores/AdminPanelStore.test";
 import { AdminPanel } from "./AdminPanel";
 
 const { adminPanelStore } = rootStore;
+
 const mockAgencyID = "10";
+
 const usersByID = groupBy(
   mockUsersResponse.users.map((user) => ({
     ...user,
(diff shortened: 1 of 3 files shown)


mxosman:
> Hi @mxosman! It seems like "adding/removing team members" tests already in, but I took the opportunity and added some more tests for Admin Panel to increase coverage.

Hi @nasaownsky - thank you so much for adding more tests - they look great! I think the adding/removing team member test we punted on and only implemented a simpler test. For that one, I believe we wanted to cover:
* the bug that we ran into where creating multiple new users via the secondary flow would not override the previously created users
* separate bug, where deleting a user deleted all of the users within the agency

nasaownsky:
Ahhh, got it, will do!

nasaownsky:
@mxosman 

>the bug that we ran into where creating multiple new users via the secondary flow would not override the previously created users

Hmm, is it even possible to make such a test? I'm tried to recreate a flow for this, but then realised that we store newly created users in a local state inside `AgencyProvisioning` before saving, so I think there is no way to inject new users as it meant to be for this bug?

>separate bug, where deleting a user deleted all of the users within the agency

Doesn't test "Deleting a user deletes a card to the list of team members" handles that?

mxosman:
We could consider refactoring the logic so we don't have this mix of local and store state - I think that might be the thing setting us up for issues down the road. (All my bad when I originally implemented this)

mxosman (approved):
Ah, right right right... I forgot about the roundtrip the the BE when creating a new user - for some reason I thought we persisted that, but we wouldn't have a new user id if we did. Gotcha - that makes sense. Thanks for thinking through this with me. I'll meditate on this some more but this is good to go!

